PR #286, opened 2014-01-01
Output arguments

egeerardyn:
Added output arguments to `matlab2tikz`. This is mainly for development purposes.
Practically, when one output argument is used, the `m2t` object is the output. Together with #285, this allows to read the preamble from `matlab2tikz`.

Next to that, I added functionality to obtain a handle to any local function. Unfortunately, this has to use `eval`, but I think that's the only way. By adding this functionality, we can eventually develop unit tests for each local function. Personally, I mostly use the ACID test as some kind of acceptance test, but ideally unit tests should do the job for many parts of the code.


egeerardyn:
I'll close this since this hasn't been acted upon for over a year.

Issue #6, opened 2010-10-26
Plot order and axis background color


mredd:
matlab2tikz isn't currently able to draw unaligned axes environments on the right positions.

To reproduce, plot the current script in MATLAB and in LaTeX:

    wave = rand(1, 2048);
    h = figure();
    ax1 = axes( 'Parent', h );
    ax2 = axes( 'Parent', h, 'Position', [0.5, 0.45, 0.35, 0.4] );
    plot( ax1, wave );
    plot( ax2, 1200:1500, wave(1200:1500) ); % Generates a "zoom window"
    matlab2tikz( 'foo.tikz' );

PeterPablo:
Since this is about to be fixed, shouldn't we also create a testcase? If yes, we have to refrain from using `rand` but rather something like a function like `magic()` with deterministic output:

```
wave = magic(16);
wave = wave(:);
h = figure();
ax1 = axes( 'Parent', h );
ax2 = axes( 'Parent', h, 'Position', [0.5, 0.45, 0.35, 0.4] );
plot( ax1, wave );
plot( ax2, 1:64, wave(1:64) ); % Generates a "zoom window"
```

![image](https://cloud.githubusercontent.com/assets/4340267/5649940/6a23a434-969a-11e4-8e57-a36914390a38.png)

PeterPablo:
Current `m2t` output (notice the transparent background):
![image](https://cloud.githubusercontent.com/assets/4340267/5650059/30acc72a-969b-11e4-872a-45d2072a8a31.png)

egeerardyn:
@PeterPablo :+1:

PeterPablo:
Alright. I'll take care of the testcase. Any suggestions for the naming?

egeerardyn:
@PeterPablo What about "overlapping plots" or "zoomed plot" ?

PeterPablo:
testcase is ready

okomarov:
#509, #510 fixed this and #512 provided test case in ACID.

PR #509, opened 2015-01-06
Keep axes plotting order

okomarov:
Instead of `setdiff()` (used in `getRelevantAxes`) which reorders the axes handles, keep the previously queried order of the handles. The order of the handles is reversed later in `m2t` and with `setdiff()` it's basically nullified. 

Fixes the ordering problem in https://github.com/matlab2tikz/matlab2tikz/issues/6


nschloe:
Excellent!

PR #510, opened 2015-01-06
Axis background

### LINKED ISSUE #6: Plot order and axis background color

okomarov:
Together with #509 fixes #6 since enforces color bg of axis when there are multiple axes and the current one is visible. Basically, complex plot with axes overlays, see `ACID(14)`, will mantain transparency while still plotting the children (lines). 

Also:
* renamed `axisIsVisible` to `isAxisVisible`
* in `getRelevantAxes` made double conversion of axes handles explicit
* store `relevantAxesHandles` in `m2t`, which is needed for the multiple axes check in `drawBackgroundOfAxes`

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -996,15 +996,11 @@ end
 function m2t = drawBackgroundOfAxes(m2t, handle)
 % draw the background color of the current axes
     backgroundColor = get(handle, 'Color');
-    if ~isNone(backgroundColor)
+    if ~isNone(backgroundColor) && isVisible(handle)
         [m2t, col] = getColor(m2t, handle, backgroundColor, 'patch');
-        % Fill background if not just white or if multiple axes and current
-        % is visible
-        if ~strcmp(col, 'white') || (numel(m2t.relevantAxesHandles) > 1 && isVisible(handle))
-            m2t.axesContainers{end}.options = ...
-                opts_add(m2t.axesContainers{end}.options, ...
-                'axis background/.style', sprintf('{fill=%s}', col));
-        end
+        m2t.axesContainers{end}.options = ...
+            opts_add(m2t.axesContainers{end}.options, ...
+            'axis background/.style', sprintf('{fill=%s}', col));
     end
 end
 % ==============================================================================
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -435,11 +435,11 @@ function m2t = saveToFile(m2t, fid, fileWasOpen)
 
     % Alternative Positioning of axes.
     % Select relevant Axes and draw them.
-    [relevantAxesHandles, axesBoundingBox] = getRelevantAxes(m2t, axesHandles);
+    [m2t, axesBoundingBox] = getRelevantAxes(m2t, axesHandles);
 
     m2t.axesBoundingBox = axesBoundingBox;
     m2t.axesContainers = {};
-    for relevantAxesHandle = relevantAxesHandles
+    for relevantAxesHandle = m2t.relevantAxesHandles(:)'
         m2t = drawAxes(m2t, relevantAxesHandle);
     end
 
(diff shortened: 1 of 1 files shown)


okomarov:
This is #6 finally right:

capture

src/matlab2tikz.m
```diff
@@ -998,7 +998,9 @@ function addCustomCode(fid, before, code, after)
     backgroundColor = get(handle, 'Color');
     if ~isNone(backgroundColor)
         [m2t, col] = getColor(m2t, handle, backgroundColor, 'patch');
-        if ~strcmp(col, 'white')
+        % Fill background if not just white or if multiple axes and current
+        % is visible
+        if ~strcmp(col, 'white') || (numel(m2t.relevantAxesHandles) > 1 && isVisible(handle))
```
egeerardyn: While this is a good start, I wonder what happens with regular `subplot`s: wouldn't this add backgrounds there as well?
Another edge case, `plotyy` should work, since then one of the backgrounds should be `none`.

If `subplots` don't work as expected (i.e. only put the background is something non-default is set), we could check if the current axes overlaps with another relevant axes to exclude subplots from this kind of behavior.

However, I think very people won't notice the difference, since in 99% of the cases the paper background is also white.

okomarov: Yes, many subplots will have white bg, as does the Matlab figure. However, physically printing it on colored paper will not print white bg...

I tested the whole ACID and I can put before after .pdfs if necessary. 

The overlapping test is something I thought about but then it is really not worth the effort for now, since things work fine (I had to ensure axis is visible and without that check, ACID(14) was breaking).

egeerardyn: @okomarov I agree with you that this fix is more important than the edge case I saw (because it is only visible when the background is not white, which should only happen on posters or beamer slides, as far as I know and the fix is not possible in a one-liner).

nschloe: > Yes, many subplots will have white bg, as does the Matlab figure.

I support the change for this exact reason. This will change many plots on a non-whitebackground, but this is exactly what MATLAB does too.

egeerardyn: Then maybe we should just leave out the `~strcmp(col, 'white')`, then the behavior is universal and one can probably set `set(gca,'Color','none')` or something to prevent that?

okomarov: I will remove the bg check and leave the visibility check only then (on that line).

src/matlab2tikz.m
```diff
@@ -435,11 +435,11 @@ function matlab2tikz(varargin)
 
     % Alternative Positioning of axes.
     % Select relevant Axes and draw them.
-    [relevantAxesHandles, axesBoundingBox] = getRelevantAxes(m2t, axesHandles);
+    [m2t, axesBoundingBox] = getRelevantAxes(m2t, axesHandles);
```
PeterPablo: By looking at the code I don't see where you need the `relevantAxesHandles` in `drawBackgroundOfAxes()`.  Consequently it is not necessary to store the `relevantAxesHandles` in the `m2t` struct. You check the handle's visibility by directly calling `isVisible()` with the current `handle`.

PeterPablo: I see now that you needed it before your most recent commit cf4f961d39bd24ec3c9402fd502e0804c2e535d0.

okomarov:
@PeterPablo I could revert on that by I thought it might help in the future with handling multiple plots. For instance, you have the collection of relevant peer axes from which you could query the positions.

src/matlab2tikz.m
okomarov: @PeterPablo I could revert on that by I thought it might help in the future with handling multiple plots. For instance, you have the collection of relevant peer axes from which you could query the positions.

PeterPablo:
Your PR fixes the issue:
image

src/matlab2tikz.m
PeterPablo: That's a valid point! Please add a short comment that documents that proposed purpose to the code.

PR #511, opened 2015-01-06
Rehash in place and continue conversion with new version

### LINKED ISSUE #172: updater needs improvement
Just downloaded the latest release from here but when I run it, it says new version available! (0.3.1). I assume it's an async problem of github and file exchange, and I allowed updater to update (Although I assume github should be always the latest version, ain't I right?)

The updater updated files:
> *** Would you like matlab2tikz to self-upgrade? y/n [n]:y
> *** Downloading and unzipping to *[some_folder]*/Library Matlab/../...`parent folder?`
> *** done.

However, all the three files are in folder (which also a path in Matlab that I put almost every extensions):
> *[some_folder]*/Library Matlab/

The downloaded files are "random" for me. It includes the following unwanted files:
> ./AUTHORS
> ./ChangeLog
> ./helpers
> ./helpers/figure2dot.m
> ./helpers/pointReduction2d.m
> ./license.txt
> ./README.md
> ./THANKS
> ./version-0.3.1

Questions are:

1. Do I actually need the two files in `helpers` folder? I couldn't find any use of them in `matlab2tikz.m` and `matlab2tikzInputParser.m`.

2. I don't know how other people install the library, however I assume there might be many people like me are more happy to maintain minimum files and centralise them in somewhere. By the default updater, those files are not updated, instead, created lots of "random" files in the parent folder.

3. I'd prefer if updater doesn't stops me from converting plot into tikz file. Instead, I would like to suggest that the converting just gone through and give a warning at the end to notify users to update. In another word, `updater.m` called within `matlab2tikz.m` is for detecting new versions and gives notification, user can call `updater.m` alone to update the library.

4. I would also like to suggest updater.m detects the existence of files before being updated, and only update the existing files in proper folder. i.e., keep a very simple searching algorithm to find where these library files are, verifies if they are the right file (in order to prevent a false search and override other library files), and update only these file exists.

okomarov:
Inspired/fixes #172: rehashes the updated version in place and continues with the conversion using the newest version.

The rehasing is tested with the following suite of functions (save the 3 functions in some folders and make a copy somewhere else):
```matlab
function main(flag)
if flag
    delete('plotrectangle.m')
    movefile('plotrectangle_2.m','plotrectangle.m')
    rehash
    main(false)
    return
end
plotrectangle
end

function plotrectangle
    plot(1:10)
end

function plotrectangle
% Save it as plotrectangle_2.m
    rectangle
end
```

Then calling from cmd window:

    main(true)

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -333,9 +333,12 @@ if m2t.cmdOpts.Results.checkForUpdates && isempty(VCID)
     m2t.cmdOpts.Results.showInfo, ...
     m2t.env...
     );
-    % Terminate conversion if update was successful (the user is notified
-    % by the updater)
-    if isUpdateInstalled, return, end
+    % Reload updated version and execute m2t conversion with that
+    if isUpdateInstalled
+        rehash
+        matlab2tikz(varargin{:})
+        return
+    end
 end
 
 %% print some version info to the screen
--- a/src/private/m2tUpdater.m
+++ b/src/private/m2tUpdater.m
@@ -147,9 +147,9 @@ function upgradeSuccess = m2tUpdater(name, fileExchangeUrl, version, verbose, en
               end
               
               upgradeSuccess = true; %~isempty(unzippedFiles);
-              userInfo(verbose, 'UPDATED: the current conversion will be terminated. Please, re-run it.');
+              userInfo(verbose, 'UPDATED.');
           catch
-              userInfo(verbose, ['FAILED: continuing with the' name ' conversion.']);
+              userInfo(verbose, ['FAILED: continuing with the old ' name]);
           end
       end
       userInfo(verbose, '');


egeerardyn:
There is a flaw in the way of testing this: `plotrectangle` is an external file that hasn't been loaded at the time the upgrade is done.

I've done my own test with three files `main.m`, `main_old.m` and `main_new.m`.

```matlab
%main_new.m
function main(flag)

fprintf('Starting NEW version ');    
if flag
    fprintf('upgrading ...')
    copyfile('main_old.m','main.m'); % restore old version
    rehash
    main(false)
    return
end

disp('New Version')

end
```
and
```matlab
% main_old.m
function main(flag)
    
fprintf('Starting OLD version ');

if flag
    fprintf('upgrading ...');
    copyfile('main.m','main_old.m');
    copyfile('main_new.m','main.m');
    rehash
    main(false)
    return
end

disp('Old Version')

end
```

The `main.m` file is a copy of either of them. What should happen, is running `main(true)` "upgrades" (in this case this actually switches between both implementations).
I've tested it and multiple calls of `main(true)` produces:

```
>> main(true)
Starting new version upgrading ...Starting new version New Version
>> main(true)
Starting OLD version upgrading ...Starting OLD version Old Version
>> main(true)
Starting NEW version upgrading ...Starting NEW version New Version
>> main(true)
Starting OLD version upgrading ...Starting OLD version Old Version
>> main(true)
Starting NEW version upgrading ...Starting NEW version New Version
>> main(true)
Starting OLD version upgrading ...Starting OLD version Old Version
```

For our purpose, this is unwanted behavior (`Starting OLD version upgrading ...Starting NEW version New Version` would be wanted behavior), because the new implementation is NOT loaded until the running version has finished.

So this means we would actually need a wrapper script that does the upgrading: that seems similar to the test you did and which passes (I guess?).

egeerardyn:
I agree, it's highly unlikely we can get this working properly. Setting a timer is a bad idea, because that can introduce race conditions.

The only thing that would work, is to wrap m2t into another function that takes care of the updating part as I suggested in #172, but that only moves the problem if the wrapper is ever to be updated. Probably, it's just better to upgrade after the conversion has taken place, but let's keep the discussion in #172.

Issue #414, opened 2014-11-04
Use look-up table for "random" data in test functions


PeterPablo:
Since it is cumbersome to come up with random data in both MATLAB and Octave, it is proposed to drop the use of `rand()`, ~~`magic()`~~ and `svd()` inside `ACID.m`. Instead, pseudo random data from a look-up table should be used.
See commit https://github.com/PeterPablo/matlab2tikz/commit/734bed7332ecf04a38503e660740b9f161ecd1c2 of pull request https://github.com/nschloe/matlab2tikz/pull/413 for discussion.

PeterPablo:
The function `magic()` seems to be fine. Initially I must have assumed that it is not available in Octave, which is [not true](http://octave.sourceforge.net/octave/function/magic.html).
`magic(n)` returns deterministic output.

PeterPablo:
`SVD()` seems to come with numerical precission issues on different platforms. See https://github.com/matlab2tikz/matlab2tikz/pull/590#issuecomment-91370672.
Concerning our efforts to reduce hash mismatches see #552.

PeterPablo:
For a contourPenny replacement, see https://github.com/matlab2tikz/matlab2tikz/issues/615#issuecomment-92420501.

PR #512, opened 2015-01-08
add testcase overlappingPlots = ACID(98)

PeterPablo:
* be warned: generation of PDF from MATLAB is incorrect (green background missing)
* m2t output will be fixed by PR #510
* tested on R2014a and R2014b
* testcase for issue #6 

image

image

### DIFF
--- a/test/suites/ACID.m
+++ b/test/suites/ACID.m
@@ -131,7 +131,8 @@ function [status] = ACID(k)
                            @texInterpreter      , ...
                            @stackedBarsWithOther, ...
                            @colorbarLabelTitle  , ...
-                           @textAlignment
+                           @textAlignment       , ...
+                           @overlappingPlots
                          };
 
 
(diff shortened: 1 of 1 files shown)


test/suites/ACID.m
```diff
@@ -2379,6 +2380,40 @@
...
 end
 % =========================================================================
+function [stat] = overlappingPlots()
+    stat.description = 'Overlapping plots with zoomed data and varying background.';
+    stat.issues = 6;
+
+    % create pseudo random data and convert it from matrix to vector
+    l = 16^2;
+    l_zoom = 64;
+    wave = magic(sqrt(l));
```
okomarov: You suggested in https://github.com/matlab2tikz/matlab2tikz/issues/414 to avoid `magic()`. 
Why don't we just use `sin()`?

PeterPablo: Good catch! I commented in #414 that `magic()` should be fine. It produces deterministic output and is available in both MATLAB and Octave.

PeterPablo: `sin()` would certainly work as well!

PeterPablo:
New output:
image

PeterPablo:
Also works on octave 3.8.2 (here the reference pdf correctly contains the green background):
image

test/suites/ACID.m
```diff
@@ -2379,6 +2380,38 @@
...
+    % plot data
+    h = figure(); clf;
+    ax1 = axes('Parent', h);
+    plot(ax1, wave);
+
+    % overlapping plots with zoomed data
+    ax3 = axes('Parent', h, 'Position', [0.2, 0.6, 0.3, 0.4]);
+    ax4 = axes('Parent', h, 'Position', [0.7, 0.2, 0.2, 0.4]);
+    ax2 = axes('Parent', h, 'Position', [0.25, 0.3, 0.3, 0.4]);
+
```
okomarov: @PeterPablo My bad, I overlooked the fact we should not create figures inside the testing sub-fuctions since `testM2t` already does that. Now, at the end of the full ACID, the last test is created and the figure pops into visibility. 

The test case should be:
```matlab
% plot data
ax1 = axes()
plot(ax1, wave);

% overlapping plots with zoomed data
ax3 = axes('Position', [0.2, 0.6, 0.3, 0.4]);
ax4 = axes('Position', [0.7, 0.2, 0.2, 0.4]);
ax2 = axes('Position', [0.25, 0.3, 0.3, 0.4]);
```

PeterPablo:
@okomarov, thank you! I copied your suggestion and pushed a new PR. I confirm it is working now as intended in R2014b and octave 3.8.2 with `figureVisible` set to `true` and `false` (default).

PR #518, opened 2015-01-10
Fix getPositionOfText (improves annotation positioning)

okomarov:
Improves `ACID(91) - annotationTextUnits` and lets Matlab carry out the conversion from e.g. `characters` to `centimeters`. The bottom right box is finally close enough, compare against `ACID(90)`.

Refactored the function (also in preparation for 3D check on axis).

capture


okomarov:
It also improves on `ACID(69) - latexmath2`. The output is however still bad in the acid.pdf due to `scale only axis` and the cropping of the `standalone` package.

capture

PR #519, opened 2015-01-11
Fix axesLocation missing box bug

### LINKED ISSUE #259: XAxisLocation on top, does not print bottom axis
Hello,
I already asked the same question on the mathworks, and you requested me to ask it hier, as it sounds a s a bug matlab2tikz, although I still find it an excellent job that you did. Here is a simple matlab code

```
t=0:0.01:3;
x1=sin(2*t);
x2=cos(3*t);

figure;plot(t,x1,'b',t,x2,'r');set(gca,'XAxisLocation','top'); 
xlabel('t');ylabel('x')
text(1.28,0.5494,'w=2')
text(2.44,0.509,'w=3','BackgroundColor',[1 1 .6])

matlab2tikz('Test.tikz')
```

Unfortunatelly, I couldn't upload any images because I gopt an unknown error.

Thanks in advance.

okomarov:
Fixes #259 and `ACID(50) - axesLocation` bug of missing box when flipping x/y axis to top/right.

### DIFF
--- a/test/suites/ACID.m
+++ b/test/suites/ACID.m
@@ -1048,6 +1048,7 @@ end
 % =========================================================================
 function [stat] = axesLocation()
   stat.description = 'Swapped axis locations.';
+  stat.issues = 259;
 
   plot(cos(1:10));
   set(gca,'XAxisLocation','top');
@@ -2390,7 +2391,7 @@ function [stat] = overlappingPlots()
     wave = sin(linspace(1,10*2*pi,l));
 
     % plot data
-    ax1 = axes()
+    ax1 = axes();
     plot(ax1, wave);
 
     % overlapping plots with zoomed data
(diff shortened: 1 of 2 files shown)


test/suites/ACID.m
```diff
@@ -2390,7 +2391,7 @@
     wave = sin(linspace(1,10*2*pi,l));
 
     % plot data
-    ax1 = axes()
+    ax1 = axes();
```
PeterPablo: :+1: ;-)

PeterPablo:
I just ran the whole testsuite under octave and noticed that for ACID(95) the box is missing. Is this related? @okomarov, do you mind checking, while your memory is still fresh on that matter?
image

okomarov:
@PeterPablo `ACID(95)` alone produces a figure with boxes off, so I guess it's `cleanfigure` (?)
capture

PeterPablo:
I just had a look at the code for `ACID(95)`. As it turns out the reason for the vanished box is the `hold on`!
Compare 
`clf; subplot(2,1,1); plot(magic(3))`
with 
`clf;subplot(2,1,1); hold on; plot(magic(3))`.
This seems to be by design (also in MATLAB). So it is fine that way.

Issue #515, opened 2015-01-09
ACID test 68 latexInterpreter and 69 latexmath2 fail


PeterPablo:
These tests fail in octave 3.8.2 because some double `$$` are present in the tex output.

PeterPablo:
fixed by PR #520 and PR #521.

PR #520, opened 2015-01-12
Do not escape dollar in replacement string (e.g., "\$$2\$")

PeterPablo:
This is not properly handled by octave 3.8.2 and produces the literal
output "$$2$", when the testsuite was executed.

The change does not affect the bahavior of MATLAB.

ACID(68) octave
image

ACID(68) R2014b
image


PeterPablo:
Fixes `ACID(68)` (issue #515).

PR #523, opened 2015-01-13
3D axis detected from view

okomarov:
Detect 3D axis through the elevation of `view()` rather than the plotting heuristic of the children. The latter is also subject to spuriously force a flattened 3D axis when the z coordinates are always all 0. With the new approach, the data is trimmed if the axis is 2D.

This approach is more robust and reduces the amount of .tex code produced.


src/matlab2tikz.m
```diff
@@ -1778,14 +1780,15 @@ function assertRegularAxes(handle)
...
-        m2t.currentAxesContain3dData = true;
+    is3D = m2t.axesContainers{end}.is3D;
+    if is3D
         columnNames = {'x', 'y', 'z'};
         Vertices    = applyHgTransform(m2t, Vertices);
         plotCmd     = 'addplot3';
     else
         columnNames = {'x', 'y'};
         plotCmd     = 'addplot';
+        Vertices    = Vertices(:,1:2);
```
PeterPablo: one line up, so that it matches the preceeding if-branch

src/matlab2tikz.m
```diff
@@ -1778,14 +1780,15 @@ function assertRegularAxes(handle)
     Vertices = get(handle,'Vertices');
     
     % 3D vs 2D
-    if size(Vertices,2) == 3
-        m2t.currentAxesContain3dData = true;
+    is3D = m2t.axesContainers{end}.is3D;
```
PeterPablo: necessity of temporary variable is questionable

src/matlab2tikz.m
```diff
@@ -2460,34 +2480,29 @@ case size(cData,2) % specific y-data is given
...
+    data = [data, color];
+        
     % Add mesh/rows=<num rows> for specifying the row data instead of empty
     % lines in the data list below. This makes it possible to reduce the
     % data writing to one single sprintf() call.
     opts = opts_add(opts,'mesh/rows',sprintf('%d', numrows));
 
-    opts = opts_print(m2t, opts, ',\n');
-    str = [str, sprintf(['\n\\addplot3[%%\n%s,\n', opts ,']'], s.plotType)];
+    str = [str, sprintf('\n\\%s[%%\n%s,\n%s]', plotCmd, s.plotType, opts_print(m2t, opts, ','))];
```
PeterPablo: may be an additional comment: `% finally print depending on the present plot command`

src/matlab2tikz.m
```diff
@@ -5624,29 +5634,23 @@ function printAll(m2t, env, fid)
...
-    if strcmpi(m2t.env,'MATLAB') && ...
-        ~isVersionBelow(m2t.env, m2t.envVersion, [8,4])
-        isHG2 = true;
-    end
+    tf = strcmpi(m2t.env,'MATLAB') && ...
+        ~isVersionBelow(m2t.env, m2t.envVersion, [8,4]);
 end
 % ==============================================================================
-function isBelow = isVersionBelow(env, versionA, versionB)
+function tf = isVersionBelow(env, versionA, versionB)
```
PeterPablo: may be at a comment somewhere that `tf` means `truefalse` (boolean) for those helper functions.

src/matlab2tikz.m
```diff
@@ -1354,23 +1355,29 @@ function assertRegularAxes(handle)
...
             end
         end
     else
-        bool = true;
+        tf = true;
     end
 end
 % ==============================================================================
+function tf = isAxis3D(axisHandle)
+% Check if elevation is not orthogonal to xy plane
```
PeterPablo: nice idea!

src/matlab2tikz.m
```diff
@@ -2414,12 +2416,24 @@ case size(cData,2) % specific y-data is given
     % to avoid it for the most default situations, e.g., when dx and dy are
     % rank-1-matrices.
     % Enforce 'z buffer=sort' if shader is flat
-    if isempty(strfind(opts_get(opts, 'shader'),'interp')) ||...
-        any(~isnan(dx(1,:)) & dx(1,:) ~= dx(2,:)) ||...
-        any(~isnan(dy(:,1)) & dy(:,1) ~= dy(:,2))
+    % Avoid 'z buffer=sort' for hist3D plots
+    isShaderFlat = isempty(strfind(opts_get(opts, 'shader'),'interp'));
+    isHist3D     = strcmpi(get(handle,'tag'),'hist3');
+    if m2t.axesContainers{end}.is3D && isShaderFlat && ~isHist3D
```
okomarov: Fixes a regression in 3D bar introduced when enforcing teh zbuffer sort with flat shader.

src/matlab2tikz.m
okomarov: Just improving readability a bit. Does not hurt terribly I suppose :smile:

src/matlab2tikz.m

src/matlab2tikz.m

src/matlab2tikz.m
okomarov: I am following Matlab's convention from the documentation (almost all is* functions have that). I am up for using `bool` if it is clearer for us. I just reckon we should stick with one standard rather than having several different `out` names depending on what the function does. At the end it is a true or false.

egeerardyn: @okomarov My personal convention is to use `bool` (or a name that actually means something like `isValid`).

I know at TMW, most of the time `tf` is used as boolean return value, but I prefer to not use it, because it shadows the `tf`  function/object (i.e. a transfer function parametrized as polynomials from their own Control Toolbox, which I use a lot for my work) and it may cause confusion in code where both are used at the same time (not likely to happen in `m2t` though).
The other reason why I dislike `tf`, is because it doesn't really mean anything except `true-false` in an obfuscated form. I also find the term `tf` for `truefalse` a bit immature: you wouldn't call an enum after all possible values it can take (e.g. "SmallMediumLarge"), but rather something meaningful, such as "shirtSize".

On the other hand, `bool` (short for boolean) is something that is present in some other programming languages, so my assumption is that such a name is more familiar to a wider audience. But it actually conveys the same info as `tf`: this value is supposed to be boolean value. But to be honest:  `bool` is just my own lazy alternative to `tf` for when I don't want to think about a proper name.

MATLAB handles boolean variables in quite a few dirty ways:
*  actual `logical` (i.e. `true` vs `false`) variables (i.e. booleans in most other languages),
* `'on'` vs `'off'` "booleans" (actually strings) that are still used in the HG framework and some other places,
* implicit conversions, e.g. the output of `exist()` that we talked about a while ago.
So I would argue that TMW isn't always the best example to follow.

But of course, this is a matter of taste. Personally, I prefer `isValid` since that actually means something and that is also the convention in many other languages.
If I'm not mistaken, that's also what most MATLAB style guides recommend.

But if there's a consensus of everybody else to use something specific, also fine by me.

okomarov: I agree with most of your points but I prefer to concentrate the verbosity on the question being asked by the `is*` function, so that the naming is clear on what the function is doing. At the end we get 0 or 1 in all cases (with those functions), so I do not see much gain in coming up with a different output name for every function. This means that I am indifferent between `bool` and `tf` but see the `isValid` alternative as a good name for a function rather than for an output name. 

Anyways, that commit should be irrelevant to the PR, so we can always revert it.

src/matlab2tikz.m
PeterPablo: certainly no problem!
just was wondering and saw that you did not do that below

src/matlab2tikz.m
PeterPablo: My engineering guts tell me to use `bool` -- mostly because I never read the `tf` "convention" in MATLAB code before and `bool` is intuitive for me. But nevermind, since this is nitpicking.

PeterPablo:
I'd say let's merge this! Are there opposing opinions?

src/matlab2tikz.m
```diff
@@ -2430,16 +2446,23 @@ case size(cData,2) % specific y-data is given
...
+        
+        columnNames = [columnNames, 'c'];
+        formatType = 'table[row sep=crcr, colormap name=surfmap, point meta=\thisrow{c}]';
+        
+%         r = CData(:, :, 1);
+%         g = CData(:, :, 2);
+%         b = CData(:, :, 3);
+%         colorFormat = join(m2t, repmat({m2t.ff},[3 1]),',');
+%         color = arrayfun(@(r,g,b)(sprintf(colorFormat,r,g,b)), ...
+%             r(:),g(:),b(:),'UniformOutput',false);
```
egeerardyn: Why is all this code in comments? Either we should document that or remove the code for clarity.

okomarov: I will remove it.

PeterPablo:
While checking the ACID testsuite for regressions I noticed that ACID(62) fails (at least on R2014b) and this can thus not yet be merged. I did not test for further failing tests.
`/test62-converted.tex:2315: Package PGF Math Error: Unknown function `z__column_not_found' (in 'z__column_not_found.'). [};]`

okomarov:
@PeterPablo I cannot test that, do not have the toolbox. Can you paste the .tex?

PeterPablo:
Here you go:
https://www.wuala.com/PeterPablo/Upload/m2t/test62-converted.tex/?key=zNl74H2fU0k3
https://www.wuala.com/PeterPablo/Upload/m2t/test62-reference.pdf/?key=zNl74H2fU0k3

Not too fancy. I hope it is fine to post the few crucial lines inside `displayspectrogram`:
```
surf(args{:},'EdgeColor','none'); %#ok<NASGU>
axis xy; axis tight;
colormap('default');
view(0,90); % AZ = 0, EL = 90 is directly overhead and the default 2-D view.
```

okomarov:
To reproduce the issue with 2D surface:
```matlab
[X,Y,Z] = peaks(25);
surf(X,Y,Z);
view(0,90)
```

Maybe we can also substitute the spectrogram with this MWE so that it does not depend on Toolboxes?

PeterPablo:
Please go ahead and substitute the ACID test! We defined it a goal to reduce dependencies somewhere. I'll test later. Thank you!

PeterPablo:
:+1: I confirm that `ACID(62)` is fixed now.

PeterPablo:
There is more going on... Quite some tests fail, e.g., ACID(66). Something is wrong in `getPositionOfText()`...

okomarov:
@spectro emulation shall belong to a separate PR. I already had a fix for `getPositionOfText()` which I wanted to submit with the next PR for HG2 annotations.

PeterPablo:
Testsuite looks good on [R2014b] (https://www.wuala.com/PeterPablo/Upload/m2t/acid_%20cc08035.pdf/?key=zNl74H2fU0k3) expect `ACID(87)` which also fails on current `HEAD`. 

Since discussion concerning `spectrogram` is still in progress, a separate PR is best.

PeterPablo:
Sorry for the merge. Hopefully I reverted it properly.
The testsuite fails in octave. Eq. with `ACID(70)` with this PR, but works on `develop`.

src/matlab2tikz.m
```diff
@@ -2617,36 +2618,36 @@ case size(cData,2) % specific y-data is given
...
-    is3D = false;
+    is3D = m2t.axesContainers{end}.is3D;
     
-    switch npos
-        % Check if 3D
-        case 3
-            if pos(3) ~= 0
-                is3D = true;
+    try 
+        cl = class(handle(h));
```
PeterPablo: This fails in octave.

okomarov: Sorry, forgot that `handle()` is not supported. I am looking into it. For the PR do not worry for now.

okomarov:
I think the simplest course of action right now is to reset the develop to 27a7672dd5dbe7cfa730550d1ab7599ff32376cd and apply an additional PR with the fix for Octave. 

Rebasing the whole 3D PR and solving all conflicts is just more error prone than the reset approach. Also, if in the future we happen to merge an unfinished PR, is best to issue fixes than to revert.

What do you think @egeerardyn?

PeterPablo:
That is fine with me. I'm a little afraid to mess up that reset. Would you please do this or @egeerardyn?

egeerardyn:
@okomarov I think the revert strategy is better because it documents what happened.
Resetting the `develop` branch is not advisable for public branches, since this alters history and may produce divergent branches (it's the same kind of mess I put you through in one of your first PRs, and that shouldn't happen again).

The problem is that resetting changes what is meant by the `develop` branch: for some it will be ahead by the faulty commits and there is no good way to signal that problem to them. A revert on the other hand,  is a signal "X introduced something bad, it's better not to have it (right now)".

IMHO, the best course of action is: to open a new PR on to of the reverted commit and explicitly reference this one.

okomarov:
@egeerardyn The revert strategy obviously tracks what is going on but here it is actually misleading. A new PR will have to start off the reverted merge with all the cherry picked commits from `is3D` PR. Also, I will have to re-solve all merge conflicts. Then if someone looks at the history, it will actually make no sense. First merge, then revert, then merge all over again. 

Note that I cannot just rebase the old `is3D` onto the reverted commit because it is already part of the history!

Thus, in this case, a reset makes actually more sense.

okomarov:
@egeerardyn Actually, I think I will just revert the reverted merge and apply the fix and issue a new PR. So, we keep the history and avoid the potential problems with conflict solving.

PeterPablo:
I am truly sorry for the mess I caused. I quickly applied the revert so that the development branch stays in a properly working condition. A reverted revert plus octave-fix sounds like a reasonable approach to me. The cluttered history reflects what happens. 
I am too unexperienced with git to judge, whether you can't simply duplicate your current PR branch and rebase it on the current HEAD of `develop`.

egeerardyn:
@okomarov Seems a ~~great~~ brilliant idea! 

@PeterPablo Don't worry. Git can be hard at some times.

You can certainly rebase an old branch on top of `develop` (as you suggest). The downside is that you need to remerge all conflicts as @okomarov argues. For such a workflow (which IMHO is better than resetting a public branch), you have to make sure that `git rerere` is enabled to avoid having to resolve the same conflicts over and over again.

PR #526, opened 2015-01-18
Histogram support

### LINKED ISSUE #525: [R2014b] histogram
When trying to save (master + develop ; both show the same error) a plot generated by the new function histogram [1] the following error output is generated:

Error using matlab2tikz>handleAllChildren (line 692)
I don't know how to handle this object: histogram

[1]  http://de.mathworks.com/help/matlab/ref/histogram.html

okomarov:
Fixes #525 by adding support for `histogram()`.

capture

On the left side the transparency is lost possibly due to `cleanfigure()`

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -3042,22 +3042,26 @@ function [m2t, str] = drawHistogram(m2t, h)
         opts = opts_add(opts, 'draw', xEdgeColor);
     end
     
-    % Bar type
-    opts = opts_add(opts, 'ybar interval');
-    
     % Data
     binEdges = get(h, 'BinEdges');
     binValue = get(h, 'Values');
     data     = [binEdges(:), [binValue(:); binValue(end)]];
-        
+    
+    % Bar type (depends on orientation)
+    isVertical = strcmp(get(h,'Orientation'),'vertical');
+    if isVertical
+        opts = opts_add(opts, 'ybar interval');
+    else
+        opts = opts_add(opts, 'xbar interval');
+        data = fliplr(data);
+    end
+    
     % Make table
     [m2t, table] = makeTable(m2t, {'x','y'},data);
     
-    % Add 'area legend' to the options as otherwise the legend indicators
-    % will just highlight the edges.
-    m2t.axesContainers{end}.options = ...
-        opts_add(m2t.axesContainers{end}.options, 'area legend');
-
+    % Add 'area legend' (x/ybar interval legend do not seem to work)
+    opts = opts_add(opts, 'area legend');
+    
     % Print out
     drawOpts = opts_print(m2t, opts, ',');
     str      = sprintf('\\addplot[%s] plot table[row sep=crcr] {%s};\n', drawOpts, table);
@@ -3181,8 +3185,7 @@ function [m2t, str] = drawBarseries(m2t, h)
 
     % Add 'area legend' to the options as otherwise the legend indicators
     % will just highlight the edges.
-    m2t.axesContainers{end}.options = ...
-        opts_add(m2t.axesContainers{end}.options, 'area legend');
+    drawOptions = opts_add(drawOptions, 'area legend');
 
     % plot the thing
     if isHoriz
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -676,6 +676,9 @@ function [m2t, pgfEnvironments] = handleAllChildren(m2t, handle)
 
             case 'rectangle'
                 [m2t, str] = drawRectangle(m2t, child);
+                
+            case 'histogram'
+                [m2t, str] = drawHistogram(m2t, child);
 
             case {'uitoolbar', 'uimenu', 'uicontextmenu', 'uitoggletool',...
                     'uitogglesplittool', 'uipushtool', 'hgjavacomponent'}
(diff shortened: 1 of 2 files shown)


test/suites/ACID.m
```diff
@@ -2434,6 +2435,24 @@
...
+  stat.issues      = 525;
+
+  x     = randn(1000,1);
+  edges = [-10 -2:0.25:2 10];
+  histogram(x,edges);
+  hold on
+  y     = 3 + randn(1000,1);
+  histogram(y);
+end
+% =========================================================================
```
okomarov: Working to remove rand calls

egeerardyn:
@okomarov The left side is an EPS produce by MATLAB itself. Since we export to EPS, it is quite likely that the alpha channel is not exported by that (I think that depends on what renderer is being used).

If the EPS looks vectorized, it should be the `painters` renderer that has no transparency, otherwise it might be the OpenGL renderer that should support transparency.

test/suites/ACID.m
```diff
@@ -2434,6 +2435,26 @@
     set(ax4, 'Color', 'none');
 end
 % =========================================================================
+function [stat] = histogramPlot()
+
+  if ~exist('histogram')
```
egeerardyn: I'm not happy with `~exist('histogram')` because this is way too broad. Just creating a file called "histogram" would pass that test, but would make the code after that break.
IMHO, the best way is `isempty(which('histogram'))` because this directly checks for the existence of such a function.

src/matlab2tikz.m
```diff
@@ -3005,6 +3008,61 @@ case size(cData,2) % specific y-data is given
...
+    binValue = get(h, 'Values');
+    data     = [binEdges(:), [binValue(:); binValue(end)]];
+        
+    % Make table
+    [m2t, table] = makeTable(m2t, {'x','y'},data);
+    
+    % Add 'area legend' to the options as otherwise the legend indicators
+    % will just highlight the edges.
+    m2t.axesContainers{end}.options = ...
+        opts_add(m2t.axesContainers{end}.options, 'area legend');
```
egeerardyn: Can't these be added to the actual entries instead of the axis? Because adding it to the axis will certainly cause other plots to become "area legends" as well.

src/matlab2tikz.m
```diff
@@ -3005,6 +3008,61 @@ case size(cData,2) % specific y-data is given
...
+    edgeColor = get(h, 'EdgeColor');
+    if isNone(edgeColor)
+        opts = opts_add(opts, 'draw', 'none');
+    else
+        [m2t, xEdgeColor] = getColor(m2t, h, edgeColor, 'patch');
+        opts = opts_add(opts, 'draw', xEdgeColor);
+    end
+    
+    % Bar type
+    opts = opts_add(opts, 'ybar interval');
```
egeerardyn: You might want to check `get(h, 'Orientation')` to see whether `ybar` or `xbar` should be used.

src/matlab2tikz.m
PeterPablo: We should probably add an ACID test with several subplots of different type to check for such issues. Opinions?

test/suites/ACID.m
okomarov: I was borrowing the code and I would personally go with `isHG2()` (which is implicitly Matlab + version check) since histogram has just been introduced in R2014b.

PeterPablo: My vote goes to `isHG2()` and please add a short comment that `histogram()` is only available in R2014b.

src/matlab2tikz.m
okomarov: @egeerardyn Sure. Was borrowing the code from `drawBarseries()` which actually suffers from this problem. I now get why mixed bars and line have the wrong legend (for the line). Basically, we should extend this fix to bars as well.

@PeterPablo You can see this in the ACID with bars and line. I personally would not dig too deep unless you wanna fix all those legend problems :P

src/matlab2tikz.m

test/suites/ACID.m
egeerardyn: @okomarov The problem with `isHG2()` is that it might not be present in older versions of MATLAB and Octave. So to be able to skip the test, one needs to check whether that function exists (otherwise an error is thrown), i.e. a catch-22.

An alternative approach, could be to have a `try-catch` inside the test and explicitly catch the error code for "function not found" (or we could even introduce that into the test handler).

src/matlab2tikz.m
egeerardyn: @okomarov Hmm, if that is indeed the state in `drawBarSeries()` I might have overlooked that issue or `area legend` cannot be added to a single plot since I touched that part of the code quite recently. Or, that `area legend` thingy changes some axis-specific things behind the scenes. 
Anyhow: the broken legends are something that I am responsible for (the upside was that at least the plots show both normal lines and bars together, which was explicitly disabled before).

okomarov: @egeerardyn I observed different behaviour in the legend depending on the version 2014a vs 2014b. So, it could be trickier than that.

test/suites/ACID.m
okomarov: @egeerardyn Not sure I follow. 

`isHG2 = strcmpi(m2t.env,'MATLAB') && ~isVersionBelow(m2t.env, m2t.envVersion, [8,4]);`

And [R2014b release notes on histogram](http://uk.mathworks.com/help/matlab/release-notes.html?#buf5bgh)

PeterPablo: @egeerardyn , we introduced `isHG2()` in `m2t` quite recently. It checks for the new graphical system which implies a) MATLAB and b) at least R2014b. Since we ship it wit m2t it is existent and sufficient for the intented check.

egeerardyn: @PeterPablo @okomarov I was confusing `isHG2` with a related function of R2014b that explicitly checks whether an handle graphics object is a HG2/MCOS objects or HG1 one. Sorry for the confusion. I agree with using `isHG2()`.

PeterPablo: Don't worry! :-)

PeterPablo:
image

PeterPablo:
Changing the orientation of the second histogram also works.
image

okomarov:
@PeterPablo I like your idea for the ACID, we should include that test!

egeerardyn:
@PeterPablo @okomarov :+1:

PeterPablo:
@okomarov, done -- see PR #529

PR #485, opened 2014-12-10
Unify test harness

egeerardyn:
This is basic support for what I introduced in #475. 
This is not completely finished, but I think this should considerably improve the CI output.

PR #527, opened 2015-01-19
[do not merge]Pr/485

PeterPablo:
test for Travis CI


egeerardyn:
See #485. I contacted Travis and apparently the problem is that the PR doesn't merge cleanly. GitHub apparently only want to hand them a "merge commit", which cannot be produced in this case and so no testing happens.
I will have a go at producing code that merges cleanly into the current develop, I'm having a few ideas that should work without rebasing this PR to make sure that this history is retained, but I still have to try a few things for that.

PeterPablo:
Ok! Just wanted to gives this a try, since I have no idea how responsice Travis is. (it seems, that tey are fast ;-) )

Issue #378, opened 2014-10-27
R2014b: text boxes aren't always handled


egeerardyn:
As reported in #344, the abs text in test case `80` is missing when running in MATLAB R2014b.

![screen shot 2014-10-27 at 08 47 53](https://cloud.githubusercontent.com/assets/177360/4787605/b67880aa-5dad-11e4-82a4-4833d8619df1.png)

okomarov:
The problem is actually in `cleanfigure` which switches the text visibility to `off`. It carries a test on the extent property to check for overlap (with the axes?), which does not pass because the extent is all NaNs.

PeterPablo:
Thank you for checking! So we need to come up with a way to reliably determine the extent of 3d axis objects it seems.

okomarov:
From what I understand we do not want to be plotting the title, xyz labels and the tickmarks as separate command in pgplots. The extent is a proxy to filter those text objects out. 

We could check if the handle is not a title or xyz label. However, we would need to discriminate boxplot labels (?). Also, I do not think those are currently handled properly anyways.

PeterPablo:
You are right that we only want to add separate TikZ commands if the appearance is caused by (manual) user interaction. Otherwise the defaults of pgfplots/TikZ should be applied.

I agree that the whole labeling mechanism needs to be overhauled. As you know there are some open issues linked to labels one way or another (colobar tick labels, HG2, textboxes, ...).

PeterPablo:
@okomarov, due to the precise pointer of you I was able to prepare a fix. Thank you!

PeterPablo:
This issue is fixed, though at some point we might reconsider the removal of textboxes outside an axis, which is currently done on (manual) execution of `cleanfigure()`.

PR #528, opened 2015-01-20
fix issue #378 of removed textboxes

PeterPablo:
* implement handling of 3D case
* improve 2D handling
* extend testcases
* fixes issue #378 

image
image

Tested on R2014a/b and octave 3.8.2.

### DIFF
--- a/src/cleanfigure.m
+++ b/src/cleanfigure.m
@@ -116,6 +116,10 @@ function indent = recursiveCleanup(meta, h, minimumPointsDistance, indent)
           coarsenLine(meta, h, minimumPointsDistance);
 
       elseif strcmp(type, 'text')
+          % Ensure units of type 'data' (default) and restore the setting later
+          units_original = get(h, 'Units');
+          set(h, 'Units', 'data');
+
           % Check if text is inside bounds by checking if the position is inside
           % the x, y and z limits. This works for both 2D and 3D plots.
           x_lim = get(meta.gca, 'XLim');
@@ -126,17 +130,25 @@ function indent = recursiveCleanup(meta, h, minimumPointsDistance, indent)
           pos = get(h, 'Position');
           bPosInsideLim = ( pos' >= axLim(:,1) ) & ( pos' <= axLim(:,2) );
 
-          % In 2D plots the extent of the textbox is available and also
+          % In 2D plots the 'extent' of the textbox is available and also
           % considered to keep the textbox, if it is partially inside the axis
           % limits.
           extent = get(h, 'Extent');
-          if all(~isnan(extent))
-             x_lim(1) = x_lim(1) - extent(3); % x-limit is extended by width
-             y_lim(1) = y_lim(1) - extent(4); % y-limit is extended by height
-             axLim = [x_lim; y_lim; z_lim];
 
-             bPosInsideLimExt = ( pos' >= axLim(:,1) ) & ( pos' <= axLim(:,2) );
-             bPosInsideLim = bPosInsideLim | bPosInsideLimExt;
+          % Restore original units (after reading all dimensions)
+          set(h, 'Units', units_original);
+
+          % This check makes sure the extent is only considered if it contains
+          % valid values. The 3D case returns a vector of NaNs.
+          if all(~isnan(extent))
+            % Extend the actual axis limits by the extent of the textbox so that
+            % the textbox is not discarded, if it overlaps the axis.
+            x_lim(1) = x_lim(1) - extent(3);    % x-limit is extended by width
+            y_lim(1) = y_lim(1) - extent(4);    % y-limit is extended by height
+            axLim = [x_lim; y_lim; z_lim];
+
+            bPosInsideLimExt = ( pos' >= axLim(:,1) ) & ( pos' <= axLim(:,2) );
+            bPosInsideLim = bPosInsideLim | bPosInsideLimExt;
           end
 
           if ~all(bPosInsideLim)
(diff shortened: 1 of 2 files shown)


src/cleanfigure.m
```diff
@@ -114,16 +114,32 @@ function cleanfigure(varargin)
...
           % Don't be too precise.
           coarsenLine(meta, h, minimumPointsDistance);
+
       elseif strcmp(type, 'text')
-          % Check if text is inside bounds by checking if the Extent rectangle
-          % and the axes box overlap.
-          xlim = get(meta.gca, 'XLim');
-          ylim = get(meta.gca, 'YLim');
+          % Check if text is inside bounds by checking if the position is inside
+          % the x, y and z limits. This works for both 2D and 3D plots.
```
egeerardyn: :+1: on the comment

src/cleanfigure.m
```diff
@@ -114,16 +114,32 @@ function cleanfigure(varargin)
...
+
+          % In 2D plots the extent of the textbox is available and also
+          % considered to keep the textbox, if it is partially inside the axis
+          % limits.
           extent = get(h, 'Extent');
-          extent(3:4) = extent(1:2) + extent(3:4);
-          overlap = xlim(1) < extent(3) && xlim(2) > extent(1) ...
-                 && ylim(1) < extent(4) && ylim(2) > extent(2);
-          if ~overlap
+          if all(~isnan(extent))
```
egeerardyn: It took me quite a while to understand what is going on here. So perhaps, we should put a comment to document the idea behind this.
As I see it: the X/Y boundaries are extended to the top left to make sure that e.g. if you have a text box that starts left/above of the axes, and overlaps it partially, it gets included.

I wonder whether that's always what we want (e.g. imagine a text label placed in the middle of a 2x2 subplot).
Wouldn't it make more sense to check if the text box lies within (or intersects) the bounding box of all axes?
Anyhow: I think the current approach is an improvement, so this should be merged once it is documented a bit more.

The other thing that popped up in my mind was: is it guaranteed that `extent` is given in plot units (i.e. as XLim/YLim/...)? I can imagine that is the default situation, but I couldn't find that property in the help at first glance.

PeterPablo: Please open a disussion issue for how to proceed with textboxes outside a single axis. I also wondered why those should be removed, but this is the way the code handled the situation before I laid my hands on it. Shortly I will add another commit that adds a comment.

The [help](http://de.mathworks.com/help/matlab/ref/text-properties.html?searchHighlight=extent) states that by default `extent` is defined in `data units` but it can be overwritten with the property `Units`.

okomarov: As I was mentioning, I think we filter out boxes outside the axes in order to avoid catching the xyz labels, titles and tick labels created ad hoc by specific plots as texts. If we were to convert those as texts, then other axes specific routines would re-draw a second time resulting in duplication. 

Now, things changed quite a bit with HG2 and it may indeed not consitute a problem anymore, but also in R2014a I am not sure this is a big issue and we should try to come up with an exact example on why we want ot filter out text boxes (maybe we weren't able to handle annotations lying on the figure, as another reason). Otherwise, I would suggest to simply kill that check and try to draw everything.

PeterPablo:
@egeerardyn, I addressed your comment and extended testcase `ACID(97)` to cover the case of different units.

src/cleanfigure.m
PeterPablo: May be @nschloe can shed some light on the history of the textbox removal?

okomarov:
@nschloe Any insights/historical reasons on the need for textbox removal in `cleanfigure` (as discussed in https://github.com/matlab2tikz/matlab2tikz/pull/528#discussion_r23218845)?

okomarov:
I did some testing and it appears that the `'text'` cleaning is mostly due to Octave. 

I put a breakpoint in cleanfigure on `set(h, 'Visible', 'off');` within the `'text'` case and run the ACID. In R2014b, it only triggers for `plotyy` on texts that are already invisble. However, this filter mostly worsens the rendering, since e.g. `@latexmath2` loses two text nodes because outside of the invisble axis (and we also have the case that this PR fixes). 

I kindly ask someone to run a full ACID on Octave with the `set(h, 'Visible', 'off');` line commented out. Then, if you could zip the .pdf together with all the converted.tex so that we can compare and decide wether it is worth keeping this filter at all (texts with empty 'String' are not rendered anyways even if visible, if I am not mistaken).

Gonna test also on R2008a, especially boxplots with specific ticklabels, and update.

okomarov:
I checked R2008a and effectively the problem is that x/y labels and title are texts which, if not removed with cleanfigure, will appear as pgfplots labels but also as text nodes. There is no easy way to detect those labels in Matlab.

For `@latexmath2` we should consider avoid applying `cleanfigure()` in the first place. @egeerardyn Do you think it's worth adding `stats.nocleanfigure = true` to the ACID?

PeterPablo:
Sorry, I could not provide octave's output in a timely manner.
As an alternative approach, we could also enlarge the axis of `@latexmath2` and add a line to the documentation that TextBoxes outside visible axis (plural) are removed, if `cleanfigure()` is executed. Keeping `cleanfigure()` enabled for the whole testsuite has the advantage of having the ability to spot strange issues, that might go undetected with only a handful of `cleanfigure`-tests -- at least until Travis is up and running.

okomarov:
No worries @PeterPablo, the problem affects HG1 Matlab too, thus HG2 is the exception. 

The `nocleanfigure` will be for a restricted set of tests like `@latexmath2` or those tests that do have a call to cleanfigure inside the specific ACID test.

Issue #392, opened 2014-10-30
cleanfigure / detecting partly visible lines


cpc-pk:
Applying cleanfigure() made disappear some lines completely from my plots. I created teh following minimal example:
plot([2 2 3 3 4 4]-.5,[2 6 2 6 2 6]); axis([1 4 1 4]);
![orig](https://cloud.githubusercontent.com/assets/503455/4843966/a2023c38-6037-11e4-942b-5fbc77746c85.png)
I briefly looked into the code and thought cleanfigure() would try to find out whether parts of the connecting lines are visible or not. However, apparently it crops all but the first two points.
![cleaned](https://cloud.githubusercontent.com/assets/503455/4843975/c9661c86-6037-11e4-8a01-b7e05d451800.png)

PeterPablo:
I was successful in recreating the described issue. Key is to further reduce the visible axis so points lay outside the axis both below and above:
```
plot([2 2 3 3 4 4]-.5,[2 6 2 6 2 6]);
axis([1 4 2.5 4]);
cleanfigure();
```
The call of `cleanfigure()` currently removes all points, because additional `NaN` are added. I am working on a fix. This is closely related to #400.

PeterPablo:
This is fixed by https://github.com/matlab2tikz/matlab2tikz/pull/531.
![image](https://cloud.githubusercontent.com/assets/4340267/6379753/df725f12-bd34-11e4-9741-f140397e2de3.png)

Issue #400, opened 2014-11-03
[ACID(22)] zoom fails


PeterPablo:
ACID(22) zoom fails, since horizontal and vertical line are not plotted. Screenshot shows output from MATLAB 2014b. 2014a gives identical results.
![image](https://cloud.githubusercontent.com/assets/4340267/4883746/292db126-6365-11e4-87d5-aa8bda0296b8.png)

PeterPablo:
This also affects `ACID(47) zplanePlot1` is fixed by https://github.com/matlab2tikz/matlab2tikz/issues/531.

![image](https://cloud.githubusercontent.com/assets/4340267/6379690/75252482-bd34-11e4-9b9b-5e8d3a8ec39b.png)

egeerardyn:
This should be fixed by #531 / #547.

PR #531, opened 2015-01-21
Fix issues #392 and #400 caused by cleanfigure

PeterPablo:
Existing ACID(21) `pruneOutsideBox()` is fixed and ACID(7) `linesWithOutliers` keeps working:
image
image

Fixes issues #392 and #400.

**edit:** Thanks go to @akloeckner for providing the missing real fix (PR #547). I incorporated it here, since I added quite some comments and restructured the code a little bit.

### DIFF
--- a/src/cleanfigure.m
+++ b/src/cleanfigure.m
@@ -383,6 +383,7 @@ function movePointsCloser(meta, handle)
     return;
   end
 
+  numberOfPoints = length(xData);
   data = [xData(:), yData(:)];
 
   xlim = get(meta.gca, 'XLim');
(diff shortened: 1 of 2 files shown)


nschloe:
Is there a reason why the revert is part of this?

PeterPablo:
No there is not :-( My git magic is struggling with  me. Only the three commits 494ff4a, f1a0df6 and ad00575 should go into this PR.

PeterPablo:
Ok, I think I got it sorted out with a dangerous `git push --force`... The pull request now contains only the three relevant commits.

okomarov:
I think the additional ACID is not strictly necessary since we can edit ACID(21) to improve the explanation, or am I missing something?

Also, the description in the new ACID is not very informative and maybe this is why I do not get in what it is different from ACID(21).

PeterPablo:
The  new test is indeed not strictly necessary but helped me track down the issue. A new aspect is the red plot that has data points inside the axis limits. The crucial comments in the new test are `% some points are outside visible axis` and `% all points are outside visible axis`.
This is not covered by ACID(21). Certainly ACID(21) could be adapted, but I didn't want to go through the hassle.

okomarov:
ACID(21) has the blue line with points inside the axis. 

By description I meant `stats.description`.

PeterPablo:
Please wait with merging. I just discovered that ACID(7) fails. This is probably the reason for the check I removed in d42ac8447f5bb04ee82fc2e45eddc85921e5c9df.

PeterPablo:
@okomarov You are right. Totally forgot about that blue sine :-) I guess we can drop the new test then.

PeterPablo:
I had another look and the issue  is tougher than it seemed at first. The new test does cover a new aspect, namely a line of two outlying points that intersect the rectangle defined by the axis.
As the last commit indicates code for the calculation of the boolean needs to be added to actually fix the problem. Otherwise either ACID(7) or the newly added ACID(100) fails.
Pointer for implementation idea: http://blogs.mathworks.com/loren/2011/08/29/intersecting-lines/

PeterPablo:
Now I am certain that the newly added test can be removed again since both aspects of moving points closer to the axis (`movePointsCloser()`) are covered by ACID(7) and ACID(22). I spent more time than intended on figuring out the base of this and am unsure if I can finish this up timely. That's why I pushed that preliminary commit to write down the findings and write down a sketch of a viable approach.

okomarov:
As you said, the problem is more complicated than it seems. I will try to think about the most common cases we need to tackle. We also need to consider that people do not necessarily use `cleanfigure()` and thus try to give priority in the algorithm to some realistic scenario, leaving corner cases to potential workarounds in the workflow, rather than trying to solve everything.

PeterPablo:
As far as I can see I reached the bottom of the issue and documented it in the code with a boolean flag that makes one of the two named ACID tests fail. I described there how to derive the state of that boolean. It isn't overly complicated, but still one has to write the code...

PeterPablo:
I rebased this PR, removed the newly added testcase, applied the fix from @akloeckner and updated the initial post to reflect this. Could please somebody else check my changes?

PR #532, opened 2015-01-21
is3D with fix fopr Octave handle()

okomarov:
Refer to #523 + fix for Octave avoiding `handle()` in `getPositionOfText`

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -2628,13 +2628,14 @@ function [m2t,posString] = getPositionOfText(m2t, h)
     
     is3D = m2t.axesContainers{end}.is3D;
     
-    try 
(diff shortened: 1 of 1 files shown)


okomarov:
I re-run the whole ACID on [R2014b](https://www.dropbox.com/s/y3mi1i8z623nib8/R2014b_after.pdf?dl=0) and also Octave now has no problems with ACID(70):

capture

and just to be sure ACID(91) on R2014a also comes out ok:

capture

Since, the problem before the previous merge was with the `handle()` in `getPositionOfText()`, these tests should suffice.

PeterPablo:
:+1: I simply was afraid of hitting that button again ;-) We are making progress! :-)

Issue #226, opened 2013-09-16
Cleanfigure doesn't properly remove points outside axis


jonasp1:
`cleanfigure()` on a stairs plot removes hidden data in the matlab figure, but not in the resulting tikz file.
```
figure();
stairs(rand(10000,1));       % generate 10000 numbers
xlim([0 20]);                % only show 20 of them
cleanfigure();               % works fine in the matlab figure window
matlab2tikz('stairs.tikz');  % still contains all 10000 points
```

PeterPablo:
This is not related to issue #400.

PeterPablo:
This issue is specific to the `stairs()` function. `plot()` was not affected.

PeterPablo:
fixed in `develop` branch

PR #533, opened 2015-01-21
fix extend pruneOutsideBox for stairs() plot

PeterPablo:
* fixes issue #226
* adds ACID test `@pruneOutsideBox`
* checked on R2014b and octave

image

### DIFF
--- a/src/cleanfigure.m
+++ b/src/cleanfigure.m
@@ -143,8 +143,7 @@ function pruneOutsideBox(meta, handle)
   yData = get(handle, 'YData');
 
   % Obtain zData, if available
-  handleFields = get(handle);
-  if isfield(handleFields, 'ZData')
+  if isprop(handle, 'ZData')
     zData = get(handle, 'ZData');
   else
     zData = [];
--- a/src/cleanfigure.m
+++ b/src/cleanfigure.m
@@ -114,6 +114,8 @@ function indent = recursiveCleanup(meta, h, minimumPointsDistance, indent)
           movePointsCloser(meta, h);
           % Don't be too precise.
           coarsenLine(meta, h, minimumPointsDistance);
+      elseif strcmpi(type, 'stair')
+          pruneOutsideBox(meta, h);
       elseif strcmp(type, 'text')
           % Check if text is inside bounds by checking if the Extent rectangle
           % and the axes box overlap.
@@ -139,7 +141,14 @@ function pruneOutsideBox(meta, handle)
 
   xData = get(handle, 'XData');
   yData = get(handle, 'YData');
-  zData = get(handle, 'ZData');
+
+  % Obtain zData, if available
+  handleFields = get(handle);
+  if isfield(handleFields, 'ZData')
+    zData = get(handle, 'ZData');
+  else
+    zData = [];
+  end
 
   if isempty(zData)
     data = [xData(:), yData(:)];
--- a/test/suites/ACID.m
+++ b/test/suites/ACID.m
@@ -133,7 +133,8 @@ function [status] = ACID(k)
                            @colorbarLabelTitle  , ...
                            @textAlignment       , ...
                            @overlappingPlots    ,...
-                           @histogramPlot
+                           @histogramPlot       , ...
+                           @pruneOutsideBox
                          };
 
 
@@ -2456,6 +2457,31 @@ function [stat] = histogramPlot()
   hold on
   h = histogram(y);
   set(h, 'orientation', 'horizontal');
+end
+% =========================================================================
+function [stat] = pruneOutsideBox()
+    % test function 'pruneOutsideBox' that is part of 'cleanfigure'
+    stat.description = ['Test function \texttt{pruneOutsideBox()} ', ...
+                        'of \texttt{cleanfigure()}.'];
+    stat.issues      = 226;
+
+    % generate 16 deterministic values
+    data = magic(4);
+    data = data(:);
+
+    % create plots
+    subplot(2,1,1);
+    plot(data,'x-');
+    xlim([10 12]);      % only show 3 values
+    title('regular plot');
+
+    subplot(2,1,2);
+    stairs(data,'x-');
+    xlim([10 12]);      % only show 2 steps
+    title('stairs plot');
+
+    % force application of 'cleanfigure()'; don't rely on testsuite
+    cleanfigure();
 end
 % =========================================================================
 function env = getEnvironment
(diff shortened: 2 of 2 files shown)


PeterPablo:
The issue is fixed, since now the output only contains five data points in each table:
```
\addplot [color=mycolor1,solid,mark=x,mark options={solid},forget plot]
  table[row sep=crcr]{%
9.8	8.6\\
10	10\\
11	6\\
12	15\\
12.2	14.6\\
};
\end{axis}

...
\addplot[const plot,color=mycolor1,solid,mark=x,mark options={solid}] plot table[row sep=crcr] {%
9	3\\
10	10\\
11	6\\
12	15\\
13	13\\
};
```

okomarov:
The description of the new test case is not very informative. If I understand correctly this test case differs from ACID(7) because we are in HG2 and thus testing `matlab.graphics.chart.primitive.Stair` instead of a simple line.

PeterPablo:
ACID(7) moves far away points close to the axis limits, whereas the new tests prunes (removes) unneeded data points outside the axis limits. The test creates a vector with 16 elements. Without `cleanfigure()` all those would be saved to the tex file. With `cleanfigure()` and more specifically the call to `pruneOutsideBox()` the excess data points are removed, resulting in a tex file with five points per subplot (down from 16).

src/cleanfigure.m
```diff
@@ -139,7 +141,14 @@ function pruneOutsideBox(meta, handle)
...
   yData = get(handle, 'YData');
-  zData = get(handle, 'ZData');
+
+  % Obtain zData, if available
+  handleFields = get(handle);
+  if isfield(handleFields, 'ZData')
+    zData = get(handle, 'ZData');
+  else
+    zData = [];
+  end
```
okomarov: I personally prefer a `try/catch` in these cases:

```matlab
  try
      zData = get(handle, 'ZData');
  catch
      zData = [];
  end
```
However, since I think we are in disagreement, I would suggest to use directly `isprop(handle,'ZData')` without qu
erying all properties `handleFields = get(handle);`.

egeerardyn: :+1: on `isprop`

test/suites/ACID.m
```diff
@@ -2458,6 +2459,31 @@
...
+    % create plots
+    subplot(2,1,1);
+    plot(data,'x-');
+    xlim([10 12]);      % only show 3 values
+    title('regular plot');
+
+    subplot(2,1,2);
+    stairs(data,'x-');
+    xlim([10 12]);      % only show 2 steps
+    title('stairs plot');
```
okomarov: I would avoid adding an additional test to the acid and instead edit the ACID(7) by:
* making it a subplot where the second plot is the `stairs...` (which in R2014b has its own type, otherwise is a `'line'`)
* add `stat.issues = 226;`

We already have too many tests in the acid, and many overlap of those overlap. This one tests the stairs type which was introduced in R2014b but otherwise is really not too different from ACID(7) to require a separate figure.

test/suites/ACID.m
```diff
@@ -2458,6 +2459,31 @@
...
+
+    % force application of 'cleanfigure()'; don't rely on testsuite
+    cleanfigure();
```
okomarov: The `cleanfigure` here breaks the modularity of the testing suite and must be avoided.

egeerardyn: @okomarov I don't fully agree. The test serves explicitly to test `cleanfigure`, so I think it is better to have it right there. That way we have the freedom to remove `cleanfigure()` from the test runner if we want.

PeterPablo: I take the position of @egeerardyn.

test/suites/ACID.m
PeterPablo: @okomarov, I agree but would additionally add the call to `cleanfigure()` (see your note below).

test/suites/ACID.m
okomarov: @egeerardyn @PeterPablo I interpreted the test in a wrong way. Now I get what we want to test. Still, I think the following testing case shows more clearly the states of the figure, i.e. before and after cleanfigure:

capture

You can clearly **see** the effects of cleanfigure. The figure is before any fixes, thus the blue line also disappears and stairs are not captured in R2014b. I reckon this one 


```matlab
% Setup
subplot(311)
title setup
hold on
plot(1:10,10:-1:1,'-r*',1:15,repmat(9,1,15),'-g*',[5.5,5.5],[1,9],'-b*')
stairs(1:10,'-m*')
plot([2,8.5,8.5,2,2],[2,2,7.5,7.5,2],'--k')
legend('cross with points','no cross','cross no points','stairs','zoom area')

% Last comes before simple zoomin due to cleanfigure
subplot(313)
title 'zoom in, cleanfigure, zoom out'
hold on
plot(1:10,10:-1:1,'-r*',1:10,repmat(9,1,10),'-g*',[5.5,5.5],[1,9],'-b*')
stairs(1:10,'-m*')
xlim([2, 8.5]), ylim([2,7.5])
cleanfigure()
plot([2,8.5,8.5,2,2],[2,2,7.5,7.5,2],'--k')
xlim([0, 15]), ylim([0,10])

% Simple zoom in
subplot(312)
title 'zoom in'
hold on
plot(1:10,10:-1:1,'-r*',1:10,repmat(9,1,10),'-g*',[5.5,5.5],[1,9],'-b*')
stairs(1:10,'-m*')
xlim([2, 8.5]), ylim([2,7.5])
```

okomarov:
I meant `ACID(@zoom)` not the 7.

okomarov:
@egeerardyn @PeterPablo Any opinions/consensus on the proposed changes, so that we can proceed with final review, re-base with conflict solutions and merge?

* use `isprop` instead of `get(handle)`
* subs `ACID(@zoom)` with my proposed snippet and add `stat.issues = 226;`

PeterPablo:
Sorry, I can't focus on m2t right now. I agree with your proposed approach! Any reason you closed this PR? I try to update it soon.

okomarov:
Really sorry if I closed it, was not my intention. I really dislike the close and comment button especially because of its different colour as opposed to the grey comment button. It's the second time I do it, before with my PR.

Sent from my iPhone

On 5 Feb 2015, at 10:32, "Peter Pablo" <notifications@github.com<mailto:notifications@github.com>> wrote:


Sorry, I can't focus on m2t right now. I agree with your proposed approach! Any reason you closed this PR? I try to update it soon.

PeterPablo:
I applied the changes. I couldn't test compilation, since I am having some issues with my setup, but at least thinks look fine in MATLAB.

PR #535, opened 2015-01-21
Remove handle() from getNumBarAndId

okomarov:
It is a precautionary measure since Octave does not support `handle()` but at the moment it treats a bar plot as a collection of patches.

In HG2 graphical handles are objects but we can convert them to HG1 handles by using `double`.

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -3144,15 +3144,20 @@ function [numBarSeries, barSeriesId] = getNumBarAndId(m2t,h)
     bargroup     = get(h, prop);
     numBarSeries = numel(bargroup);
     
-    % HG2 handling:
-    % - BarPeers are sorted in reversed order wrt HG1
+    
     if isHG2(m2t)
+        % In HG2, BarPeers are sorted in reverse order wrt HG1
         bargroup = bargroup(end:-1:1);
+    
+    elseif strcmpi(m2t.env,'MATLAB')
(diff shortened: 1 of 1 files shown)


src/matlab2tikz.m
```diff
@@ -3143,10 +3143,19 @@ case size(cData,2) % specific y-data is given
...
+    
+    % HG2 handling:
+    % - BarPeers are sorted in reversed order wrt HG1
     if isHG2(m2t)
         bargroup = bargroup(end:-1:1);
     end
-    [~, barSeriesId] = ismember(handle(h), bargroup);
+    % - Ensure graphic classes are double handles 
+    %   NOTE: Octave does not support handle()
+    h        = double(h);
```
egeerardyn: I thought I saw somewhere that converting the new HG2 objects to doubles would be deprecated in the future, however, I cannot find a source for that claim.

But it makes me wonder: maybe we should just make the `handle()` call dependent on Octave/ML instead of using a cast to doubles, e.g. `if isMATLAB, h = handle(h), end`.
Also, I have to say I'm unsure why we would need to call the `handle` function. So probably that needs to be documented (or even removed, and then the problem would be solved as well).

okomarov:
Casting graphic object to double might become deprecated. However, it is not an official communication. 

I think your suggested approach is better than a panacea call to double. Basically depending on whether we are in HG2, HG1 or Octave we have some minor fixes, which I comment on.

PR #536, opened 2015-01-21
Improve backward compatibility

okomarov:
The changes in the code are minimal. Just  not using tilde (~) to suppress output.

I ran the whole ACID with R2008a and it goes all through.


egeerardyn:
I like the idea, because indeed the `~` output argument is not supported before R2009b if I remember correctly.
However, I suggest you rename `~` to e.g. `dummy`.
I trust that overwriting the same variable like you do works, but IMHO this is not obvious from the first use.
If you overwrite the same variable because of `mlint` messages, please just put an ignore flag `%#ok` at the end of the line.

okomarov:
I am gonna issue a clean PR.

PR #554, opened 2015-03-02
Fixes for 3d-axes and transparent images

LJeub:
Bugfixes:

- Fixes calculation of the tight bounding box for 3d-axes with manual PlotBoxAspectRatio or DataAspectRatio (current calculations assume 2d-axes)

- Preserves manual aspect ratio by setting the plot box ratio property in the tikz file

- Fixes bugs with transparent images (alpha data was not flipped along with color data and integer alpha data was not treated correctly)


okomarov:
Do you have some snippets to reproduce the bugs?

src/matlab2tikz.m
```diff
@@ -2378,6 +2395,9 @@ case size(cData,2) % specific y-data is given
             end
         case 3    % Text within a 3d plot
             pos = applyHgTransform(m2t, pos);
+            
+            %3d axes (need to set this for axes that contain only text
+            m2t.currentAxesContain3dData=true;
```
okomarov: We are not using that approach for 3D axis. We have now a direct check with the method `isAxis3D()`. Basically, these 3 lines are unnecessary.

LJeub: Here is a test case that should reproduce the 3d-axis related bugs:

figure;
axes('view',[30,15],'DataAspectRatio',[0.5,1,1],'position',[0.1,0.1,0.8,0.8],'box','on’);
text(1,1,1,'test');
axes('position',[0.1,0.1538,0.8,0.6924],'box','on','color','none’);

(the line mentioned below does indeed not seem to be necessary anymore, I wrote this before that change was introduced)



For the transparency bug:

figure;
line([0.5,2.5],[0.5,2.5]);
hold on
imagesc([0,1;1,0],'AlphaData',uint8([100,200;100,200]));

The attached files show what the output should look like in matlab.

egeerardyn:
@LJeub: please issue future PRs against our *develop* branch from a topic branch of your own as described on [our wiki](http://../wiki/Recommended-git-workflow). That makes it easier for us to incorporate your changes.

@matlab2tikz/developers : *Please do note that this PR requests a pull into our master branch. If we are to incorporate this PR, it has to be merged into* **develop** *by hand (or maybe even rebased on top of it).*

LJeub:
Ok, I could close this pull request and resubmit it as two separate requests in that way if that helps?

PeterPablo:
@LJeub Yes, this would be ideal, since then also the two issues are separated. Please add the test code and screenshot before and after your fix, so that other developers can easily follow you. This will certainly speed up code review.

PR #555, opened 2015-03-02
Use `onCleanup` to close all files

### LINKED ISSUE #508: Use onCleanup where needed
Use `onCleanup` (MATLAB's `try finally` construct) where this is appropriate.
E.g. the `try catch` in 8ea662e94c72c9c26ac4f27bc919dd4ad6f73596 and some places discussed in the testing framework as discussed in #485.

The same functionality is supported in Octave, by the way :smile:

egeerardyn:
Some `try-catch` constructions -- that rethrew the error AFTER `fclose(fid)` --
were also removed, since these were no longer needed.

This will close #508 when merged.

### DIFF
--- a/src/figure2dot.m
+++ b/src/figure2dot.m
@@ -35,6 +35,7 @@ function figure2dot(filename)
   set(0, 'ShowHiddenHandles', 'on');
 
   filehandle = fopen(filename, 'w');
+  fclose_filehandle = onCleanup(@() fclose(filehandle));
 
   % start printing
   fprintf(filehandle, 'digraph simple_hierarchy {\n\n');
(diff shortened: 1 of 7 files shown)


PeterPablo:
Code looks good and testsuite ran well in octave (and Travis agrees ;-) ). :+1: 

One minor remark: Since code with `onCleanup` is not that common, it might be a good idea to add `% see #508` in all lines with `onCleanup`.

PeterPablo:
`grep` found one left over case of `fclose` in `./test/private/VersionControlIdentifier.m`. Would you please adapt there as well or add a short comment, since it is not really needed?

egeerardyn:
I've addressed most of your comments.

The only thing I don't like is adding such a comment as that is a standard (albeit somewhat uncommon) MATLAB construct. Looking up #508 from MATLAB takes longer than just typing `help onCleanup`, for people who don't understand it from the context.
Instead, I've renamed the placeholder variable to `finally_fclose_fid` to communicate the intent more clearly and in the same jargon used in almost every other programming language.

Issue #556, opened 2015-03-03
Output of testHeadless in octave


PeterPablo:
Should we disable [paging](https://www.gnu.org/software/octave/doc/interpreter/Paging-Screen-Output.html) by `more off` in a preparation step of `testHeadless()`?

Otherwise execution waits for user input. As a local workaround I added `more off` to my `.octaverc`.

egeerardyn:
I've run into the same hinderance a lot during the preparation of #485 but was too lazy to set that setting.
For `runMatlab2tikztests`, I would agree on disabling the pager. For `testHeadless` I wouldn't do it.

PeterPablo:
If you as a *pro* run into this, others certainly will also struggle. We need to at least document it?! Let's change it then at least in `runMatlab2tikztests`.

egeerardyn:
I know MATLAB quite well, but I don't know all the ins-and-outs of Octave at all :smile_cat: 
But I agree with you that the least we can do, is to document it in some way (e.g. check that pager setting and raising a warning with the right syntax to overcome this issue).

PR #559, opened 2015-03-03
Ensure that paging is disabled when running travis

PeterPablo:
see issue #556


test/runMatlab2TikzTests.m
```diff
@@ -13,11 +13,19 @@
 suite = @ACID;
 allTests = 1:numel(suite(0));
 
+%% Prepare environment
+if strcmp(getEnvironment(), 'Octave')
```
egeerardyn: Please use `strcmpi` instead of `strcmp`.

egeerardyn:
:+1:

Issue #540, opened 2015-02-13
`axis` options like `axis equal` and `axis image` are not handled


PeterPablo:
`ACID(56)` should generate tex code which contains the statement `axis equal` inside the axis options.
Issue #167 is related, since this is where the testcase originates from. The problem is more visible, when the axis limits are changed, e.g. with `xlim([-1 2])'.

Should be:
![image](https://cloud.githubusercontent.com/assets/4340267/6188846/ad3f1c2c-b395-11e4-8191-7fa045e5c7e3.png)

Current situation:
![image](https://cloud.githubusercontent.com/assets/4340267/6188851/bd70be2a-b395-11e4-8166-ed31d9aa0d7e.png)

PeterPablo:
PR https://github.com/matlab2tikz/matlab2tikz/pull/560 by @LJeub will fix this in a different manner for 3D plots.
Still, I think we need this for 2D plots (e.g., `imagesc(magic(3)); axis image`). Any opinions?

LJeub:
This should be easy to implement by setting the `unit vector ratio`, however this does not behave as expected in pgfplots currently (see the discussion in #560 and http://sourceforge.net/p/pgfplots/bugs/80/). PR #560 only brings 3D plots up to parity with 2D plots, i.e. it works when `width` and `height` are set correctly but gets stretched otherwise.

PeterPablo:
See also the test cases `mandrillImage` and `contourPenny`, which invoke `axis equal` and `axis image`.

PR #560, opened 2015-03-03
Fix 3d axes placement

LJeub:
This fixes a bug in drawing 3d axes with manual DataAspectRatio. The old code handling this case assumes that the axes are 2d. 

The following example should illustrate the problem nicely:

```matlab
figure('Position',[1,1,560,420]);
axes('view',[60,50],'DataAspectRatio',[0.5,1,1],'position',[0.1,0.1,0.8,0.8],'box','on');
annotation('rectangle',[0.2621,0.1,0.4759,0.8],'color','r')
```

Here is what the output should look like in matlab:
test_matlab

Here is what the tikz output looks like before the fix:
test_old

and after the fix:
test_fixed


okomarov:
I noticed that you are inferring the dimensionality of the axis from the data. Please, use the `m2t.axesContainers{end}.is3D` value (which is retrieved with `isAxis3D` with a direct test on the view). 

Also, I prefer the verbosity of `xLim`, `yLim` (and `'zLim'`) from the previous version, it makes the code easier to read.

src/matlab2tikz.m
```diff
@@ -4817,22 +4830,50 @@ case size(cData,2) % specific y-data is given
...
+                % Matlab to fit the axes position box)
+                projection = view(axesHandle);
+                vertex1 = projection * [0;0;0;1];
+                vertex2 = projection * [1;0;0;1];
+                vertex3 = projection * [0;1;0;1];
+                vertex4 = projection * [0;0;1;1];
+                vertex5 = projection * [1;1;0;1];
+                vertex6 = projection * [1;0;1;1];
+                vertex7 = projection * [0;1;1;1];
+                vertex8 = projection * [1;1;1;1];
```
egeerardyn: Can't we name these vertices more clearly (than just a number) or do away with those indices using some matrix calculations, e.g. as below (haven't double checked everything, though):

```matlab
projection = view(axesHandle);
vertices = projection * [0 1 0 0 1 1 0 1
                         0 0 1 0 1 0 1 1
                         0 0 0 1 0 1 1 1 
                         1 1 1 1 1 1 1 1];
verticesXY = vertices([1 2], :);

% the size of the projected plot box is limited by the long diagonals
A = [ 0  0  0 -1 +1  0  0  0
      0  0 -1  0  0 +1  0  0
      0 -1  0  0  0  0 +1  0
     -1  0  0  0  0  0  0 +1];
diagonals = verticesXY * A';
dimensions = max(abs(diagonals), [], 2);
...
```

I've also taken the liberty to re-order the diagonals (`diag1` moved behind the others, to get this nice structure).

LJeub: I double checked this and it is indeed equivalent. Also note that this calculation works equally well for 2d views so we could simplify the code even further as 2d and 3d views are equivalent in Matlab. 

Note that this calculation does not work correctly for perspective projections (the plot box is no longer necessarily constrained by the long diagonals for those). Pgfplots does not support perspective projections as far as I know, so maybe we should display a warning in that case?

egeerardyn: @LJeub:

This is from the `pgfplots` manual:
> The view is (currently) always an orthogonal projection, no perspective is possible, yet. You can, however, specify projection unit vectors for x, y, and z explicitly to get a skewed three–dimensional axis.

So it seems to be possible, but I don't think it's that common to do. In any case: it's out of scope for this PR. (But I think that with these explicit matrix calculations, it's probably easier to handle the generic case).

PeterPablo:
@egeerardyn 
I did not dive into this PR, but wanted to note that when it comes to defining the `unit vector ratio` in xyz directions, this is related to issue #540 (`axis equal`). Recently I created a simple line plot, where I wanted to retain the slope *m* (`y=m x+t`) in the graphical output. I had to manually set `unit vector ratio={1 1}` to convince pgfplots to do so, otherwise the line would be displayed skewed. This was mainly due to `xlim` and `ylim` being different by the factor of about three.

LJeub:
@PeterPablo 

This PR should fix issue #540 by setting the `plot box ratio` (see L831 and L4794). Setting the `unit vector ratio` to anything other than `{1 1}` is usually not a good idea as this actually changes the definition of the units in TikZ which has unintended consequences for the figure size. Setting the `plot box ratio` achieves the same effect without the scaling issues.

PeterPablo:
Interestingly I had to explicitly set the ratio to `{1 1}`. Please note that the mentioned plot dealt with an included `graphic`. But nevermind, since this is unrelated to your PR. I might give setting `plot box ratio` a try.
Thank you @LJeub for your contribution in any case!

LJeub:
Setting the ratio to `{1 1}` forces Tikz to only scale the figure proportionally, which is probably what you were trying to achieve. I would be curious to know if the changes in this PR fix your plot. They definitely fix the example mentioned in #540

PeterPablo:
**Test 1:**  regular 2D plots work (`axis tight` is probably fine like this)
```
x = 1:20;
y = 2*x-10;

subplot(4,1,1);
plot(x,y);
axis image;

subplot(4,1,2);
plot(x,y);
axis equal;

subplot(4,1,3);
plot(x,y);
axis square;

subplot(4,1,4);
plot(x,y);
axis tight;
```
image


**Test 2:** 2D image works: `axis tight` again behaves differently
```
x = 1:10;
y = 2*x;
pic = zeros(max(y),max(x));
for ii=1:length(x)
    pic(y(ii),x(ii)) = 1;
end

subplot(4,1,1);
imagesc(pic)
axis image;

subplot(4,1,2);
imagesc(pic)
axis equal;

subplot(4,1,3);
imagesc(pic)
axis square;

subplot(4,1,4);
imagesc(pic)
axis tight;
```
image


**Test 3:** 3D plot works; Issue #540 is fixed `ACID(56)` -- why is restricting `zlim` not clipping the sphere?
```
ACID(56);
xlim([-1 2]);
zlim([-1 0.8]);
```
image

LJeub:
Sorry, I managed to confuse myself when I updated the code, the axis tight situation should now be fixed.

PeterPablo:
The generated output of Test 1 and 2 look identical. Behavior of `axis tight` as such is unchanged. How does the output look like for you?

LJeub:
Still works for me, did you change anything else? Here is what the tests look like for me

Test 1:

test_1

Test 2:

test_2

PeterPablo:
You are right. Even though I made sure I have your latest commit, MATLAB seems to not have taken notice of the change. I redid Test 1 and 2 and the output is equivalent to yours!
:+1:  from my POV (running some tests manually, I still did not review the code (properly)).

@egeerardyn, @okomarov: It would be great if one you could make the review/merge.

okomarov:
We should add/edit existing test cases in the ACID. If we edit `ACID(56)`, then the `stat.description` should reflect that we are testing for aspect ratio too.

PeterPablo:
@LJeub, I had another look at why my figure does not reproduce properly and it turns out that calling `m2t` with `(..., 'nosize', true)` leads to `plot box ratio` not being present in the output.

**Test 4:**
``` 
x = 1:10;
y = 2*x;
pic = zeros(max(y),max(x));
for ii=1:length(x)
    pic(y(ii),x(ii)) = 1;
end

imagesc(pic)
axis image;
```

`matlab2tikz('figurehandle',gcf,'filename',filepath,'standalone',true);` vs.
`matlab2tikz('figurehandle',gcf,'filename',filepath,'standalone',false);`

Adding `plot box ratio=1 2,` to the tex file admittedly is not sufficient, though `unit vector ratio=1 1,` would be. What do you suggest?


The manual says about `plot box ratio`:
```
Allows to customize the aspect ratio between the three different axes in a three dimensional plot.
Note that this key is different from the related unit vector ratio: the plot box is only useful for
three dimensional axes, and it will usually distort the unit vector ratios. If you want equal unit ratios, consider using unit vector ratio.
```

LJeub:
I did some more testing using variants of test 3 and your example with different combinations of `plot box ratio`, `unit vector ratio`, and different options for `scale mode`. It seems that `scale mode = uniformly`, which is automatically set by setting `unit vector ratio`, is severely broken in pgfplots (tested with version 1.11 and 1.12).

 With a single scaling factor, it should always be possible to satisfy at least one of `width` or `height`but for 3D plots no matter which value I chose for `scale uniformly strategy` this did not work and scaling the `unit vector ratio` by a constant would sometimes result in different figure sizes (e.g. `2 2 1` vs. `4 4 2`). By using `plot box ratio`, we can work around this problem, because it works with `scale mode = stretch to fill`. The only problem with that is that changing the `width/height` ratio in the tikz file will change the appearance of the plot. (I'm going to change the code such that `plot box ratio` is set even when `noSize = true`, but this means that the output will still look stretched, depending on what values for `width` and `height` are provided later.)

For 2D plots using `unit vector ratio` sometimes works but one can still run into problems with the figure size by scaling the `unit vector ratio` depending on the choice for `scale uniformly strategy`. 

If the bug in pgfplots is fixed, the ideal solution would be to use `unit vector ratio` for both 2D and 3D plots (note that in Matlab the  `DataAspectRatio` is stored differently, such that `unit vector ratio = 1./DataAspectRatio`). For now the code works correctly provided we do not change the 'width/height' ratio of the axes, though this will always be a problem with the `noSize` option.

LJeub:
In conclusion, I don't think it makes sense to deal with the 2D case until the bug with pgfplots is fixed.

okomarov:
@cfeuersaenger, can you please check the bug LJeub is mentioning two comments above?

egeerardyn:
@LJeub Can you submit the possible bug you've found in pgfplots to their [bug tracker](http://sourceforge.net/p/pgfplots/bugs/)? I guess that's easier for @cfeuersaenger to keep track of.

LJeub:
I've submitted a bug report to pgfplots here: https://sourceforge.net/p/pgfplots/bugs/80/

cfeuersaenger:
I will look into it. However, I might be unable to work on this item 
until next weekend.

Kind regards

Christian

Am 16.03.2015 21:23, schrieb Egon Geerardyn:

okomarov:
I am gonna resolve conflicts, merge into local branch, edit ACID(56) to have unequal axis limits (`xlim([-1 2]); zlim([-1 0.8]);`), updated hashes and either merge into develop or re-submit a review PR.

okomarov:
@LJeub I merged and re-solved conflicts in https://github.com/matlab2tikz/matlab2tikz/tree/fix3Daxes. 

Some simple 2D cases produce an additional instruction for the `plot box ratio`. For example, ACID(2) has the line:

`plot box ratio=6.283185 1,`

which appears to be redundant, i.e. no graphical differences if removed
Would you be able to look into it and see if it is legit?

LJeub:
Given the way I coded it, this is to be expected for axes that have either the `DataAspectRatioMode` or `PlotBoxAspectRatioMode` set to `manual` in Matlab. Setting the `plot box ratio` should not have any effect for 2D-plots according to the pgfplots manual, so I didn't add an additional check to remove it for 2D plots. Really what we want to do here is set the `unit vector ratio` here instead, which affects both 2D and 3D plots, but that doesn't work as expected in the current version of pgfplots (see the bug report mentioned above and also the discussion on PR #540 )

okomarov:
Then we should not set the plot box ratio for 2d.

PeterPablo:
For 2D axis I prefer *speaking* options `axis equal` and `axis image equal` over some hard to interpret floating number combinations in `plot box ratio` (see #540).
It would be great, if this could be achieved by this PR. If not, we can improve the m2t output for 2D axis, when implementing #540.

okomarov:
I agree with @PeterPablo and in this PR I would only stick to 3D. This means, we should not set the plot box ratio for 2d. 
**Anyone** can just introduce a check with `is3D()` and commit to https://github.com/matlab2tikz/matlab2tikz/tree/fix3Daxes, merge develop and update hashes. 

I won't have much time, and won;t be active until next Tuesday.

PeterPablo:
@LJeub, do you want to polish this PR during the next couple of days or should somebody else take over?

LJeub:
I should have some time to look at it this weekend

PR #561, opened 2015-03-03
Fix transparency

LJeub:
Fixes two bugs related to the handling of transparent images: 

1. alpha data needs to be flipped along with color data if axis direction is reversed

2. integer alpha data should not be clipped to [0,1]

Here is a small example to illustrate the problems:

```matlab
figure;
line([0.5,2.5],[0.5,2.5],'linewidth',2,'color','k');
line([0.5,2.5],[2.5,0.5],'linewidth',2,'color','k');
hold on
imagesc([0,1;2,3],'AlphaData',uint8([64,128;192,256]))
```

This is what the output should look like in Matlab:
test_matlab

This is what the tikz ouput looks like before the fix:
test_old
(rendering may vary depending on pdf reader)

and after:
test_fixed

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -2095,7 +2095,7 @@ function [m2t, str] = imageAsTikZ(m2t, handle, xData, yData, cData)
             hX = (xData(end)-xData(1)) / (length(xData)-1);
         otherwise
             error('drawImage:arrayLengthMismatch', ...
-                'Array lengths not matching (%d = size(cdata,1) ~= length(xData) = %d).', m, length(xData));
+                'Array lengths not matching (%d = size(cdata,1) ~= length(xData) = %d).', size(cData,1), length(xData));
     end
     X = xData(1):hX:xData(end);
 
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -2023,8 +2023,10 @@ function [m2t, str] = imageAsPNG(m2t, handle, xData, yData, cData)
     [colorData, alphaData] = flipImageIfAxesReversed(m2t, colorData, alphaData);
     
     % Write an indexed or a truecolor image
+    hasAlpha = true;
     if isfloat(alphaData) && all(alphaData(:) == 1)
         alphaOpts = {};
+        hasAlpha = false;
     else
         alphaOpts = {'Alpha', alphaData};
     end
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -4072,7 +4072,7 @@ function [m2t, xcolor] = getColor(m2t, handle, color, mode)
                             xcolor{i, j} = xc;
                         end
                     end
-                else
+                elseif ndims(color) <= 2
                     [m2t, colorindex] = cdata2colorindex(m2t, color, handle);
                     for i = 1:m
                         for j = 1:n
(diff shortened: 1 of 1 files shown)


src/matlab2tikz.m
```diff
@@ -2156,7 +2162,10 @@ case size(cData,2) % specific y-data is given
             error('matlab2tikz:UnknownAlphaMapping', ...
                   'Unknown alpha mapping "%s"', alphaMapping);
     end
-    alpha = min(1,max(alpha,0)); % clip at range [0, 1]
+    
+    if isfloat(alpha) %important, alpha data can have integer type which should not be scaled
```
egeerardyn: Shouldn't the alpha data always be mapped to a float range [0,1] ? With integers, I suppose `0` is invisible and `255` is opaque, but together with L2026 (` hasAlpha = ~all(alphaData(:) == 1);`) this has a different meaning for integers and floats.

I guess you know more about the actual values, so I hope you can enlighten me on what happens in such a case.

LJeub: Good point, the `hasAlpha` check on L2026 indeed only works correctly for floating point `alphaData`. I've modified L2026 to only check for opaqueness when `alphaData` is float. Always printing the `alphaData` when it is integer is save and should only be inefficient in very rare cases.

src/matlab2tikz.m
```diff
@@ -4050,14 +4060,25 @@ case size(cData,2) % specific y-data is given
...
-                        xcolor{i, j} = xc;
+                
+                if ndims(color) == 3
+                    for i = 1:m
+                        for j = 1:n
+                            [m2t, xc] = rgb2colorliteral(m2t, color(i,j, :));
+                            xcolor{i, j} = xc;
+                        end
+                    end
+                else
```
PeterPablo: Could you make this else branch more specific (`else if`) and trow an error in `else`, if dimensions of `color` differ from the expected/handled dimensions?

egeerardyn:
I will merge this into develop tonight, but I first want the nice test case you have provided to our ACID suite.

src/matlab2tikz.m
okomarov: I always wondered why we include the else case for the color with the error. It does not hurt but I think that it is conceptually unnecessary sicne the color data is queried from the object (as opposed to some user/developer input).

okomarov:
I additionally tested on a loaded image with scaled alpha data mapping and it works nicely:

capture


```matlab
load mandrill
figure;
line([0,500],[0,500],'linewidth',3,'color','r');
line([0,500],[500,0],'linewidth',3,'color','r');
hold on
image(X,'AlphaDataMapping','scaled','AlphaData',kron([64,128;192,256],ones(size(X)/2)))
colormap(map)
```

PeterPablo:
Instead of introducing additional tests, I think it would be a good idea to add the lines posted by @okomarov to our `mandrill` test case. Hmm, or while at it remove the dependency of `mandrill` to something that is also available in `octave`...

okomarov:
I did not mean to suggest an additional test. The one proposed by LJeub is perfectly fine. I just wanted a quick check on a real image with some funky option. It's my way to say yes to the merging.

PeterPablo:
My comment was rather directed to @egeerardyn, who indicated that he wants to add the test before merging.

egeerardyn:
@PeterPablo : Valid point about not introducing new tests. I was actually thinking to replace the current transparency test in there with a more structured alpha channel. I really like the cross on top to easily verify the image.

PeterPablo:
:+1:

Issue #391, opened 2014-10-30
Reduce complexity of the code


egeerardyn:
I've recently run `mlint -cyc` (or `checkcode -cyc`) on the current (i.e. 4202e7360d7e42aa24b7668e26c3dad3a7dfc1d5)  `matlab2tikz.m` file to determine the [cyclomatic complexity](http://en.wikipedia.org/wiki/Cyclomatic_complexity). This is a rough estimate of how complicated a function is and it basically measures how many paths there are through a function (e.g. `if`s, `switch/case`es, `return`s typically increase this measure) there are in the function, which again is a rough estimate of how hard the code is to understand, test and maintain.

Ideally, each function should have a complexity of less than about `10`, unless there are good reasons why it should be more complex. Below is a table that shows the cyclomatic complexity of every function we have:

Line | Function                     | Complexity
----:|:-----------------------------|:---------:
723  | `drawAxes`                   | 51         
3740 | `getLegendOpts`              | 37         
2233 | `drawText`                   | 28         
576  | `handleAllChildren`          | 27    
        | *others removed for brevity* |    

As with any metric, the results have to be interpreted with due care and not every value `>10` is by definition bad, but it might be an indication too much is being done in a single function. 
The following list of functions should be re-evaluated to see whether we can make them simpler (e.g. by introducing new functions) or whether the complexity is inherent to what the function is doing and is acceptable:

- [x] `drawAxes`         
- [x] `drawBarseries`         
- [x] `drawHggroup`         
- [x] `drawImage`      
- [x] `drawLine`         
- [x] `drawPatch`        
- [x] `drawQuiverGroup` 
- [x] `drawScatterPlot`         
- [x] `drawSurface`         
- [x] `drawText`                  
- [x] `drawTextbox`   
- [x] `findLegendInformation`
- [x] `getAxisOptions`         
- [x] `getColorbarOptions`         
- [x] `getLegendOpts`         
- [x] `getMarkerOptions`     
- [x] `getPositionOfText`    
- [x] `handleAllChildren`      
- [x] `legendPosition`   
- [x] `makeTable`        
- [x] `matlab2tikz`      
- [x] `matlabTicks2pgfplotsTicks`         
- [x] `parseTexSubstring`
- [x] `patchcolor2xcolor`         
- [x] `prettyPrint`      
- [x] `saveToFile`         
- [x] `surfaceOpts`         
- [x] `surfaceOptsOfSurf`
[...]

PR #567, opened 2015-03-07
Simplification

egeerardyn:
This partly addresses #391. So no new features here, mostly refactorings, although I did some tiny changes here and there (see commit messages in those cases).

Statistics:

          Function          |Complexity 
----------------------------|-----------
        `drawPatch`         |    22     
`matlabTicks2pgfplotsTicks` |    19     
     `drawScatterPlot`      |    18     
    `getColorbarOptions`    |    15     
   `shaderOptsSurfPatch`    |    14     
     `getMarkerOptions`     |    13     
       `drawContour`        |    13     
      `drawBarseries`       |    12     
  `findLegendInformation`   |    11     
         **Total**          |    137  

complexity

Each commit has been tested with R2014b. The whole changeset (i.e. everything until f31dd72) has been tested with R2014a as well. For the Octave part, I rely on Travis.

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -2712,11 +2712,7 @@ function [m2t,posString] = getPositionOfText(m2t, h)
 % makes the tikz position string of a text object
     pos   = get(h, 'Position');
     units = get(h, 'Units');
-    xlim  = getOrDefault(m2t.currentHandles.gca, 'XLim',[-Inf +Inf]);
-    ylim  = getOrDefault(m2t.currentHandles.gca, 'YLim',[-Inf +Inf]);
-    zlim  = getOrDefault(m2t.currentHandles.gca, 'ZLim',[-Inf +Inf]);
-    
-    is3D = m2t.axesContainers{end}.is3D;
+    is3D  = m2t.axesContainers{end}.is3D;
     
     % Deduce if text or textbox
     type = get(h,'type');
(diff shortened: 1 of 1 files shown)


okomarov:
Except from my comments to the single commits I agree on everything else. I add a couple of points:
* I am aware of `drawPatch` being complex and I can look into semplifying it. I recall thinking of a couple of methods before.
* Can we get rid of `drawStemSeries` and `drawStairSeries` which respectively simply call `drawStemOrStairSeries_(m2t, h, 'ycomb');` and `drawStemOrStairSeries_(m2t, h, 'const plot');`, and use those calls in the switch statement of `drawHggroup`?

egeerardyn:
@okomarov Thanks for reviewing my changes. I agree with most of your comments and will adapt accordingly. For the other things, I hope we can find a consensus. If not, I'm willing to revert those changes.

I'm a bit hesitant to remove `drawStemSeries` and `drawStairSeries` since those actually tell what is being done and encapsulate part of the `pgfplots` details (i.e. that the plot type is a `const plot` or `ycomb`). Having those inside `drawHgGroup` mixes the levels of abstraction, that is: HgGroup has nothing to do with those `pgfplots` concepts and I think we should keep that separation. Your observation that these functions do very little themselves is very correct, but allows for that separation of concerns.

PeterPablo:
By looking at the code I could not identify (potential) issues. As such I am fine with the changes and will have another brief look, as soon as you remove the wip label.

egeerardyn:
I consider this done, except maybe for a clear consensus on 63a3e66 and a3f8447?

okomarov:
For me it's good to go.

PR #570, opened 2015-03-11
Refactor environment

### LINKED ISSUE #563: Refactor `getEnvironment`
The `getEnvironment` function serves to determine whether MATLAB or Octave is run.
Right now, we cache the result (mostly as a variable `env` and `m2t.env`) that we pass around all over the place.

It's probably better to memoize its results using `persistent` variables and just call `getEnvironment` where we need it. I propose to memoize, since the code to determine Octave vs. MATLAB is very slow. The [same approach](https://www.gnu.org/software/octave/doc/interpreter/How-to-distinguish-between-Octave-and-Matlab_003f.html#How-to-distinguish-between-Octave-and-Matlab_003f) is suggested by Octave itself, although they use a different way to detect Octave (ours is better, since it works generically between MATLAB and Octave and gives version information).

egeerardyn:
This fixes #563. Basically, I got fed up of passing `m2t` around to access the global variable `env`. From within `matlab2tikz`, I've replaced all instances that I could find.
On the way, I found a little bug that was not covered by any test case (since it only occurs in error situations). I fixed it anyway :smile_cat:

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -464,7 +464,7 @@ function m2t = saveToFile(m2t, fid, fileWasOpen)
     set(0, 'ShowHiddenHandles', 'off');
     % - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
     % actually print the stuff
-    minimalPgfplotsVersion = formatPgfplotsVersion(m2t, m2t.pgfplotsVersion);
+    minimalPgfplotsVersion = formatPgfplotsVersion(m2t.pgfplotsVersion);
 
     m2t.content.comment = sprintf('This file was created by %s.\n', m2t.name);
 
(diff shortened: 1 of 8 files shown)


src/matlab2tikz.m
```diff
@@ -5808,28 +5805,39 @@ function printAll(m2t, env, fid)
...
 end
 % ==============================================================================
-function bool = isHG2(m2t)
+function bool = isHG2()
 % Checks if graphics system is HG2 (true) or HG1 (false).
 % HG1 : MATLAB up to R2014a and currently all OCTAVE versions
 % HG2 : MATLAB starting from R2014b (version 8.4)
-    bool = strcmpi(m2t.env,'MATLAB') && ...
-           ~isVersionBelow(m2t.env, m2t.envVersion, [8,4]);
+    [env, envVersion] = getEnvironment();
```
PeterPablo: I meant this line in the commit note. You might want to remove the parenthesis here as well -- please do according to your liking.

PeterPablo:
I added some minor comments. :+1: from my point of view.

src/matlab2tikz.m
egeerardyn: :+1: I will keep the parentheses though (I like it much better this way, their absence at other places is just what I think is idiomatic Matlab code, but I'm really not fond of that style).

egeerardyn:
Small bug fix (also in an error branch, when the environment is not Octave or MATLAB) and typo fix inside.

Issue #281, opened 2013-12-13
Reducing the number of datapoints in a plot. 


WG-:
It is known that the compile time of tikz suffers a lot when there are a lot of datapoints. Now I have this matlab figure, 6 subplots, which you can download here: https://dl.dropboxusercontent.com/u/20782274/block_xs_1e-3.fig

Now if I just export this the file is 37 Megabytes which is really to much to compile with lualatex, it really takes a long time. 

Now I tried cleanfigure(), cleanfigure('minimumPointsDistance',0.001) and cleanfigure('handle',h,'minimumPointsDistance',0.001). Herein h is the handle of one of the subplots e.g. h = subplot(2,3,1).

However none of these methods actually work and reduce the number of datapoints :( 

What I want to establish with this issue that well I want my current issue resolved and to point this out that the minimumPointsDistance does not actually work as expected.

widdma:
Hi all, I too have had a go at adding line simplification to the cleanfigure function. My approach uses the Visvalingam--Whyatt algorithm, which iteratively removes points based the area of the triangle it creates with its neighbours. It's therefore fine for non-monotonic data. 

It also selects a tolerance based a target output size and resolution, which defaults to 15cm x 9cm @ 600ppcm. 

Runtime is O(n log(n)) in the number of points.

It resides in the line-simp branch of [widdma/matlab2tikz](https://github.com/widdma/matlab2tikz).
Tests, comments and suggestions are welcome.

egeerardyn:
@widdma Looks fine. If you file a pull request, it's easier for everybody to inspect your changes and comment on them.

PeterPablo:
I was about to write the same. Please file the PR against `develop` branch. It looks like you applied your changes to that anyhow :smile:

nschloe:
@widdma Nice idea! This is so much more appropriate than the crude `minimumPointsDistance` we've had.
@PeterPablo Thanks for sharing the links.

egeerardyn:
This should be fixed by means of #571.

PR #571, opened 2015-03-12
Add Visvalingam simplification to cleanfigure

widdma:
This patch replaces the `minimumPointsDistance` feature with a `targetResolution` option. The cleanfigure script then performs path simplification with Visvalingam's algorithm using a tolerance based on `targetResolution`. 

Essentially, cleanfigure now removes points which do not add any detail at the specified target resolution.

### DIFF
--- a/src/cleanfigure.m
+++ b/src/cleanfigure.m
@@ -6,8 +6,8 @@ function cleanfigure(varargin)
 %   CLEANFIGURE('handle',HANDLE,...) explicitly specifies the
 %   handle of the figure that is to be stored. (default: gcf)
 %
-%   CLEANFIGURE('minimumPointsDistance',DOUBLE,...) explicitly specified the
-%   minimum distance between two points. (default: 1.0e-10)
+%   CLEANFIGURE('targetResolution','WxH@Res',...) explicitly specify the
+%   target resolution of the path simplification. (default: '15x9@600')
 %
 %   Example
 %      x = -pi:pi/1000:pi;
@@ -50,15 +50,15 @@ function cleanfigure(varargin)
 
   % Set up command line options.
   m2t.cmdOpts = m2tInputParser;
-  m2t.cmdOpts = m2t.cmdOpts.addParamValue(m2t.cmdOpts, 'minimumPointsDistance', 1.0e-10, @isnumeric);
   m2t.cmdOpts = m2t.cmdOpts.addParamValue(m2t.cmdOpts, 'handle', gcf, @ishandle);
+  m2t.cmdOpts = m2t.cmdOpts.addParamValue(m2t.cmdOpts, 'targetResolution', '15x9@600',@isstr);
 
   % Finally parse all the elements.
   m2t.cmdOpts = m2t.cmdOpts.parse(m2t.cmdOpts, varargin{:});
 
   % Recurse down the tree of plot objects and clean up the leaves.
   for h = m2t.cmdOpts.Results.handle(:)'
-    recursiveCleanup(meta, h, m2t.cmdOpts.Results.minimumPointsDistance, 0);
+    recursiveCleanup(meta, h, m2t.cmdOpts.Results.targetResolution, 0);
   end
 
   % Reset to initial state.
@@ -67,7 +67,7 @@ function cleanfigure(varargin)
   return;
 end
 % =========================================================================
-function indent = recursiveCleanup(meta, h, minimumPointsDistance, indent)
+function indent = recursiveCleanup(meta, h, targetResolution, indent)
 
   type = get(h, 'Type');
 
@@ -93,7 +93,7 @@ function indent = recursiveCleanup(meta, h, minimumPointsDistance, indent)
   if ~isempty(children)
       for child = children(:)'
           indent = indent + 4;
-          indent = recursiveCleanup(meta, child, minimumPointsDistance, indent);
+          indent = recursiveCleanup(meta, child, targetResolution, indent);
           indent = indent - 4;
       end
   else
@@ -114,7 +114,7 @@ function indent = recursiveCleanup(meta, h, minimumPointsDistance, indent)
           movePointsCloser(meta, h);
           % Don't be too precise.
           % coarsenLine(meta, h, minimumPointsDistance);
-          simplifyLine(meta, h);
+          simplifyLine(meta, h, targetResolution);
       elseif strcmpi(type, 'stair')
           pruneOutsideBox(meta, h);
       elseif strcmp(type, 'text')
@@ -368,7 +368,7 @@ function coarsenLine(meta, handle, minimumPointsDistance)
   return;
 end
 % =========================================================================
-function simplifyLine(meta, handle, varargin)
+function simplifyLine(meta, handle, targetResolution)
   % Reduce the number of data points in the line handle.
   % this can help with plots that contain a large amount of data points not
   % all of which need to be plotted.
@@ -379,7 +379,7 @@ function simplifyLine(meta, handle, varargin)
   zData = get(handle, 'ZData');
 
   if ~isempty(zData)
-    % Don't do funny stuff when zData is present.
+    % Don't simplify 3d plots
     return;
   end
   if isempty(xData) || isempty(yData)
@@ -398,28 +398,33 @@ function simplifyLine(meta, handle, varargin)
   if isYlog
 	  vyData = log10(yData);
   end
-
-  area = featureArea(vxData,vyData);
-
-  if nargin < 3
-	  % Automatically guess a tol based on the area of the figure and the expected
-	  % area of the output
-	  a = axis(meta.gca);
-	  if ~isXlog
-		  xrange = (a(2)-a(1));
-	  else
-		  xrange = (log10(a(2))-log10(a(1)));
-	  end
-	  if ~isYlog
-		  yrange = (a(4)-a(3));
-	  else
-		  yrange = (log10(a(4))-log10(a(3)));
-	  end
-	  
-	  tol = xrange*yrange/(15*9*600);
-  else
-	  tol = varargin(1);
+  
+  % Automatically guess a tol based on the area of the figure and the expected
+  % area of the output
+  if strcmp(targetResolution,'off')
+      return
   end
+  [parms,nparms] = sscanf(targetResolution,'%fx%f@%f');
+  
+  if nparms < 3
+      parms(3) = 600;
+  end
+  
+  a = axis(meta.gca);
+  if ~isXlog
+      xrange = (a(2)-a(1));
+  else
+      xrange = (log10(a(2))-log10(a(1)));
+  end
+  if ~isYlog
+      yrange = (a(4)-a(3));
+  else
+      yrange = (log10(a(4))-log10(a(3)));
+  end
+
+  tol = xrange*yrange/(4*prod(parms));
+  
+  area = featureArea(vxData,vyData);
 
   mask = area>tol;
   mask = mask | any(isnan(xData)')' | any(isnan(yData)')';
@@ -432,13 +437,19 @@ function simplifyLine(meta, handle, varargin)
 end
 % =========================================================================
 function a = featureArea(x,y)
+  % Performs Visvalingam's algorithm:
   % Given the path defined by x,y, the function returns list of the
   % maximum area associated with each point in the list. Path then can be
   % simplified via x = x(a>tol)
+  %
+  % Runtime is O(n log(n))
+  %
+  % Used by 'simplifyLine'. 
 
   n = numel(x);
   
-  %Index of next and previous elements in the linked list which defines the path
+  % Index of next and previous elements in the linked list which defines
+  % the path
   llst = [(0:n-1)',[(2:n)';0]];
   
   % 'heap' stores the points in an array heap, referenced by their index
@@ -450,13 +461,14 @@ function a = featureArea(x,y)
   pos = (1:len)';
 
 
-  area = @(i,j,k) abs((y(j) - y(k)).*x(i) + (y(k)-y(i)).*x(j) + (y(i)-y(j)).*x(k))/2;
+  area = @(i,j,k) abs((y(j) - y(k)).*x(i) + (y(k)-y(i)).*x(j) + ...
+                                                    (y(i)-y(j)).*x(k))/2;
 
-  %precompute the areas
+  % Endpoints are assigned infinte area so they can't be removed
   a = [Inf;area((1:n-2)', (2:n-1)', (3:n)')';Inf];
-  %return
 
-  %store the maximum area removed so far
+  % Keep track of the maximum area removed so far to ensure 
+  % a given element will only be excluded after earlier elements
   maxArea = 0;
 
   %Heapify the 'heap' array based on the area, using Floyd's alg
@@ -469,26 +481,26 @@ function a = featureArea(x,y)
 
   %Now remove the element with the minimum area off the heap
   while len > 1
-
-    % if a(heap(1)) > maxArea
-    %   maxArea = a(heap(1));
-    % else
-    %   % ensure current element will only be excluded when earlier elements
-    %   % are also excluded
-    %   a(heap(1)) = maxArea;
-    % end
+     if a(heap(1)) > maxArea
+       maxArea = a(heap(1));
+     else
+       % ensure current element will only be excluded when earlier elements
+       % are also excluded
+       a(heap(1)) = maxArea;
+     end
 
     % remove smallest element from heap
     e = pop(1);
 
-    % remove element from linked list
+   
     left = llst(e,1);
     right = llst(e,2);
+    % remove element from linked list
     llst(left,2) = llst(e,2);
     llst(right,1) = llst(e,1);
 
 
-    %Update area of neighbouring points to the removed, if they're not the ends
+    %Update area of neighbouring points if they're not the ends points
     if llst(left,1) > 0
       a(left) = area(llst(left,1),left,llst(left,2));
       pop(pos(left-1));
@@ -506,7 +518,7 @@ function a = featureArea(x,y)
     a(heap(1)) = maxArea;
   end
 
-  % Move element at "root" down the heap, assuming the heap property
+  % Movse element at "root" down the heap, assuming the heap property
   % is satisfied for the rest of the tree
   function down(root)
     while 2*root  <= len %while the root has a child
--- a/src/cleanfigure.m
+++ b/src/cleanfigure.m
@@ -6,8 +6,14 @@ function cleanfigure(varargin)
 %   CLEANFIGURE('handle',HANDLE,...) explicitly specifies the
 %   handle of the figure that is to be stored. (default: gcf)
 %
-%   CLEANFIGURE('targetResolution','WxH@Res',...) explicitly specify the
-%   target resolution of the path simplification. (default: '15x9@600')
+%   CLEANFIGURE('targetResolution',[W,H,Res],...)  
+%   Reduce the number of data points in the line handle by removing points which
+%   add features with area smaller than 1/4 of a pixel at the target resolution.
+%   W is the target width of the figure, H is the target height, and Res is the
+%   target resolution.
+%      Use targetResolution = Inf, or targetResolution(3) = Inf to disable line
+%   simplification.
+%  (default [15 9 600])
 %
 %   Example
 %      x = -pi:pi/1000:pi;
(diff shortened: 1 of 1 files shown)


PeterPablo:
See issue #281.

@widdma Thank you for your contribution! Do you have a minimal example that nicely shows the performance of the new algorithm? We might want to either adapt an existing ACID test or introduce a new one.

src/cleanfigure.m
```diff
@@ -50,15 +50,15 @@ function cleanfigure(varargin)
 
   % Set up command line options.
   m2t.cmdOpts = m2tInputParser;
-  m2t.cmdOpts = m2t.cmdOpts.addParamValue(m2t.cmdOpts, 'minimumPointsDistance', 1.0e-10, @isnumeric);
   m2t.cmdOpts = m2t.cmdOpts.addParamValue(m2t.cmdOpts, 'handle', gcf, @ishandle);
+  m2t.cmdOpts = m2t.cmdOpts.addParamValue(m2t.cmdOpts, 'targetResolution', '15x9@600',@isstr);
```
egeerardyn: It is advised to use `@ischar` for checking a string. 

Anyhow, I think we're better off with a more structured format that we can validate more easily. I'd propose something like `[15 9 600]` or `{15, 9, 600}`. Those we can check, and we don't need to parse them at the time we use them.

We also have to provide some backward compatibility: the `minimumPointsDistance` parameter should remain valid for at least 1 release (even if it doesn't do anything). So I would propose to add:

```
m2t.cmdOpts = m2t.cmdOpts.addParamValue(m2t.cmdOpts, 'minimumPointsDistance', 1.0e-10, @isnumeric);

m2t.cmdOpts = m2t.cmdOpts.deprecateParam(m2t.cmdOpts, 'minimumPointsDistance', 'targetResolution')
```

PeterPablo:
Please furthermore adjust the ACID test, that makes use of `minimumPointsDistance`: `plain_cos()`.

src/cleanfigure.m
```diff
@@ -6,8 +6,8 @@ function cleanfigure(varargin)
 %   CLEANFIGURE('handle',HANDLE,...) explicitly specifies the
 %   handle of the figure that is to be stored. (default: gcf)
 %
-%   CLEANFIGURE('minimumPointsDistance',DOUBLE,...) explicitly specified the
-%   minimum distance between two points. (default: 1.0e-10)
+%   CLEANFIGURE('targetResolution','WxH@Res',...) explicitly specify the
+%   target resolution of the path simplification. (default: '15x9@600')
```
egeerardyn: I think we need a bit more explanation on this parameter to make sense of the three numbers that are given. I think it also should be mentioned that the dimensions of the 2 first ones are "length" (e.g. centimeter or inch) and the last is "points per [same unit]".

src/cleanfigure.m
```diff
@@ -367,6 +368,259 @@ function coarsenLine(meta, handle, minimumPointsDistance)
...
+    % Don't simplify 3d plots
+    return;
+  end
+  if isempty(xData) || isempty(yData)
+      return;
+  end
+  if numel(xData) <= 2
+      return;
+  end
+  if strcmpi(targetResolution,'off')
```
egeerardyn: Wouldn't it make more sense to have the algorithm switched off if the resolution (`param(3)`) is infinity or NaN?

src/cleanfigure.m
```diff
@@ -367,6 +368,259 @@ function coarsenLine(meta, handle, minimumPointsDistance)
...
+        root = swap;
+      end
+    end
+  end
+
+  function up(child)
+    while child > 1
+      parent = bitshift(n,-1);
+      if a(heap(child)) < a(heap(parent)) ...
+              || (a(heap(child)) == a(heap(parent)) && rand(1)>0.5)
```
egeerardyn: What's that `rand` doing here?
Is it really essential that the algorithm is randomized? If not: please hard-code a value, since such random behavior makes testing much harder.

src/cleanfigure.m
```diff
@@ -367,6 +368,259 @@ function coarsenLine(meta, handle, minimumPointsDistance)
...
+  % Move element at "root" down the heap, assuming the heap property
+  % is satisfied for the rest of the tree
+  function down(root)
+    while 2*root  <= len %while the root has a child
+      lchild = 2*root; 
+      rchild = lchild+1;
+      % Find the minimum of the root and its children
+      swap = root;
+      if a(heap(lchild)) < a(heap(swap)) ...
+          || (a(heap(lchild)) == a(heap(swap)) && rand(1)>0.5)
```
egeerardyn: Same remark as https://github.com/matlab2tikz/matlab2tikz/pull/571/files#r26300183: `rand` is bad for testing.

src/cleanfigure.m
```diff
@@ -367,6 +368,259 @@ function coarsenLine(meta, handle, minimumPointsDistance)
...
+        swap = lchild;
+      end
+      if rchild <= len && (a(heap(rchild)) < a(heap(swap)) ...
+          || (a(heap(rchild)) == a(heap(swap)) && rand(1)>0.5))
```
egeerardyn: Same remark as https://github.com/matlab2tikz/matlab2tikz/pull/571/files#r26300183: `rand` is bad for testing.

src/cleanfigure.m
```diff
@@ -367,6 +368,259 @@ function coarsenLine(meta, handle, minimumPointsDistance)
...
+      if swap == root
+        break
+        % otherwise, swap the root and its minimum child
+      else
+        %swap in the pos array
+        pos(heap(root)-1) = swap;
+        pos(heap(swap)-1) = root;
+
+        %swap in the heap array
+        t = heap(root);
```
egeerardyn: You can swap 2 variables `a` and `b` more easily using `[a, b] = deal(b, a);`.
Inside an array, it's even easier to do the following:
```matlab
heap([root swap]) = heap([swap root])
```
for `root` and `swap` of identical dimensions.

egeerardyn:
@widdma This contribution is very welcome and I really like most of it, so most of my comments are minor things on style.

However, two significant points that we should look at are:

* [ ] the format of the input parameter (and some backwards compatibility for the old parameter)
* [ ] the use of `rand` in the algorithm

src/cleanfigure.m
```diff
@@ -367,6 +368,259 @@ function coarsenLine(meta, handle, minimumPointsDistance)
...
+
+  % Extract the data from the current line handle.
+  xData = get(handle, 'XData');
+  yData = get(handle, 'YData');
+  zData = get(handle, 'ZData');
+
+  if ~isempty(zData)
+    % Don't simplify 3d plots
+    return;
+  end
```
okomarov: The algorithm does not work in 3d?

src/cleanfigure.m
```diff
@@ -367,6 +368,259 @@ function coarsenLine(meta, handle, minimumPointsDistance)
...
+  
+  if nparms < 3
+      parms(3) = 600;
+  end
+
+  % Get info about log scaling.
+  isXlog = strcmp(get(meta.gca, 'XScale'), 'log');
+  vxData = xData;
+  if isXlog
+	  vxData = log10(xData);
```
okomarov: One indentation only here

src/cleanfigure.m
PeterPablo: Quick searching yields: http://www.sciencedirect.com/science/article/pii/S0097849398000375
It seems it is possible.

okomarov: We could probably work with a projection of z, however let's put a todo note there.

src/cleanfigure.m
```diff
@@ -367,6 +368,259 @@ function coarsenLine(meta, handle, minimumPointsDistance)
...
+      yrange = (a(4)-a(3));
+  else
+      yrange = (log10(a(4))-log10(a(3)));
+  end
+  tol = xrange*yrange/(4*prod(parms));
+  
+
+  %Split up lines which are seperated by NaNs
+  nans = isnan(vxData) | isnan(vyData);
+  idx = sort([1,find(nans)-1,find(nans)+1,numel(vxData)]);
```
okomarov: Would this fail on consecutive nans, e.g. [1,2,3,nan,nan,10,20]?

src/cleanfigure.m
```diff
@@ -367,6 +368,259 @@ function coarsenLine(meta, handle, minimumPointsDistance)
...
+      end
+  end
+  xData = horzcat(linesx{:});
+  yData = horzcat(linesy{:});
+
+  % Set the new (masked) data.
+  set(handle, 'XData', xData);
+  set(handle, 'YData', yData);
+
+  return;
```
okomarov: Unnecessary(?)

src/cleanfigure.m
egeerardyn: @okomarov I think that indeed this code assumes there are no consecutive NaNs.

src/cleanfigure.m
okomarov: We can try to translate the 600 parameter into the minimum point distance and viceversa (when the user insists on using `'minimumPointsDistance'`), hence keeping backward compatibility in the syntax. 

Also, we could allow for double syntax `targetResolution` to make it easier to play with the algo:
* scalar, that sets only the points per area
* all 3 parameters in a vector, for more advanced tweaking.

For the rest I agree with @egeerardyn.

src/cleanfigure.m
egeerardyn: Regarding 3D: let's indeed put a TODO. 
Probably one of the reasons why this hasn't been addressed (in our old code and here), is because in 3D there are more intricate things that can be done improperly than in 2D:
- we need to reduce the number of points (e.g. with such an algorithm)
- the matter of perspective / view angle comes into play to determine whether a feature is distinguishable or not
- the data formats used by the different objects is more complicated (e.g. X/Y vectors or possibly meshgridded matrices and Z a matrix)
- the simplified data matrix needs to comply to the same constraints: I expect that we can only remove a complete row or column of the Z matrix in general, except maybe for line plots (which is a direct generalization of this code, but with 3D points).

okomarov:
Some general remarks about the algo. 
It is not Matlabish style, and it is fine as long as it works well :smile:, in terms of simplication results on some test cases and performance. 

Maybe @widdma if you used some sources, could you reference them (if not already done) in the help part of the algo?

PeterPablo:
Searching GitHub I found several implementations in C and Python. [This one](https://github.com/code42day/vis-why/blob/master/index.js) comes with easy readable variable and function names. I posted a link to the original research paper in the above referenced issue.

egeerardyn:
@PeterPablo I also found a Python implementation by @widdma [here](https://github.com/widdma/tikz-scripts/blob/master/tikzsimplify).

@okomarov I agree the style is not really MATLAB-style, but I wholeheartedly agree that having (reasonably fast) and working code is already a great first step. If it really bothers us, we can always come back to the algorithm and rewrite it in MATLAB style (although that is not always straightforward starting from an algorithm description or C-ish code).

widdma:
Thank you all for the encouragement and helpful criticism. I'll address general criticisms here and particulars in the line notes. I'm somewhat new to Github, so please if this is not the best way to do things.

@okomarov and @egeerardyn, I agree that it's not particularly MATLAB-style code which was a deliberate choice to keep the code fast; however, it's still somewhat slower than I'd like. I've tried to document my use of implicit data structures, but if you consider this too difficult to read/maintain, I can rewrite it in a more MATLAB-style.

On the use of `rand`: it's not necessary and can safely removed. This was to prevent the simplification from preferentially removing points at the start of the path. This came up in cases with periodic functions and even grid spacing.

src/cleanfigure.m
widdma: While the algorithm does indeed extend to 3D quite nicely, along with the issues @egeerardyn pointed out, selecting a good tolerance level is not quite as easy as the 2D case.

It's certainly a TODO.

src/cleanfigure.m
widdma: Thanks, by switching between the Matlab editor and vim, I seem to have mixed tabs and spaces (the horror!). I'll fix my indentation up. Is there a prefered shift width for this project?

egeerardyn:
@widdma At one point or another, I think we should rewrite it to more MATLAB-ese, but for me the code style wouldn't stop me from merging this PR. If you address the other points we've raised, that'd be :+1: 

If you have any more questions (related to this or to git/github): just shoot :smile: . But it's indeed good practice to reply to specifics in the commits and the overall stuff in the PR itself. The specifics will fold once they are addressed, so this keeps the high-level overview lean.

src/cleanfigure.m
egeerardyn: 4 spaces per indent is preferred, but to be honest we still have quite some chunks with 2 spaces per indentation level. In any case: 8 is too much :wink:

widdma:
Should this be enabled by default? It is quite a large change to apply, but then again, they are calling `cleanfigure`.

src/cleanfigure.m
widdma: Because we (currently) only use the total number of points/pixels in the target image, ie. product of these 3 parameters, perhaps we should just use that as the parameter and rename it to something like `targetNumberOfPoints`? We can then have a note on calculating this parameter in the description .

Alternatively, we could expose the bare `tol` parameter and allow the user to do their own `figArea/targetNumberOfPoints` calculation. This gives the user more freedom but means a good default is not possible.

Maybe we could expose both:
  -  A `minimumFeatureArea` option for fine grained control to set manually the `tol` parameter of Visvalingam algorithm,
  -  A `targetResolution` option which selects a good `tol` for `minimumFeatureArea` based on the figure properties.

If we want to keep support for `minimumPointsDistance` with this function, having `tol=sqrt(2*minimumPointsDistance)` seems like a good choice, but of course won't be identical.

src/cleanfigure.m
widdma: Great, I find 4 spaces much easier to read in matlab code, but most of the surrounding stuff was 2 spaces, so I went with that. 

Converted my sections to 4 spaces indentation.

src/cleanfigure.m
widdma: Thanks, now removed.

src/cleanfigure.m
widdma: @okomarov Thanks, I've added your snippet. Unless anyone has any objections, I'm replacing consecutive of NaNs with just one NaN.

src/cleanfigure.m
widdma: rand removed.

egeerardyn:
@widdma If the results and performance are comparable (or better) than before, I'd vote to enable it by default.

PeterPablo:
Me too. I see a crucial step in determining a sensible default value, in case the user does not specify one.

okomarov:
Unfortuantely, performance is not brilliant.
I tested it on `ACID(47)` (the slow test) with the .fig provided by Egon at [mat files](https://dl.dropboxusercontent.com/u/452005/m2t/acid47/index.html).

It reduces points from ~5e4 to 148, which is nice, but it takes 14seconds on my i7-2600. I reckon we should try to improve `pop()`, `push()` and `down()`.

untitled

src/cleanfigure.m
```diff
@@ -298,73 +300,253 @@ function pruneOutsideBox(meta, handle)
...
+    len = numel(heap);
+
+    % 'pos' stores the current position of each point in the heap array.
+    % Needed to lookup position of points when their neighbours are updated.
+    %
+    % pos(i) = 0 denotes that the elements are not in the heap.
+    pos = [0;(1:len)';0];
+
+    area = @(i,j,k) abs((y(j) - y(k)).*x(i) + (y(k)-y(i)).*x(j) + ...
+        (y(i)-y(j)).*x(k))/2;
```
okomarov: Add comment: area by schoelace formula.

okomarov:
I tried to look into the algo (also with some help from http://en.wikipedia.org/wiki/Binary_heap) and it is indeed intuitive that it should run faster than calling a `min()` every time an area is updated. However, the fact that `min` is builtin could outpace the interpreted implementation in Matlab. 

One last attempt is to initialize the `pos`, `heap` as `uint*` and area as single. Should be faster, but might not be enough. 

@matlab2tikz/developers How about .mex (ing) `featureArea`?

widdma:
@okomarov yeah, it is certainly slower than I would like. While algorithm is inherently O(nlog(n)), the matlab interpreter seems to be giving it quite a large constant of proportionality.

It seems like a lot of time is lost in function call overhead. If I explicitly expand all the function calls, the runtime on my machine for 1e5 points, as measured by `tic;cleanfigure;toc`, drops from 13.4s to 6.5s. (I'm only have access to matlab 2012b right now, which has a lot of overhead when running profile. Runtimes are closer to 70s with profiling on)

Of course, this leads to some terribly unwieldy code. I've put it in the branch `line-simp-test`

Using mex is an option, but it introduces distribution problems.

egeerardyn:
@okomarov I would refrain from MEXing this. As @widdma indicates, this creates a lot of distribution problems: if we provide binaries, we need to provide it for a lot of OSes, MATLAB versions, ... If we don't we make our tool less usable for a lot of people. I know that in my department, most people don't have a compiler installed by default (let alone have it configured). I doubt anybody would install a compiler specifically to export some plots.

MEX is nice-to-have, but I'd prefer to have everything in MATLAB code. Having small parts also (!) available in MEX is something we might consider. But for me that's a last resort.

@widdma I've also observed that call overhead can be quite high in MATLAB.

src/cleanfigure.m
egeerardyn: @widdma : Fine by me. As long as the parameters we expose are somewhat orthogonal to each other and understandable to our users, it's fine by me.

okomarov:
I agree, and have not thought about the distro nightmare :|

okomarov:
In terms of performance, I reckon we can leave the code as is and check that `testHeadless()` completes in the ~same time as without this PR. 

Basically, for now I am considering the impact on **development** and I expect that the cleaning time will trade away compile time in our favor, see ACID(47). (This won't be true on Win, since raw ACID(47) takes 4s to compile and 12s to clean, but I am the minority here and do not have the Tollbox anyways :smile:)

For the **user**, I consider the issue from two perspectives:
* the current algo is an improvement on the previous clenfigure which e.g. has no effect on ACID(47). Thus, reduces memory problems in latex, size of figures and compile time. A few seconds of wait time are a huge improvement on minutes on trying to figure out how to avoid memory issues in Latex. 
* I tend to think that plots with lots of points are somewhat an edge case (I might be wrong here) and that the average user never uses cleanfigure. 

I also think that this algo requires more resources but operates in a smarter way. Hence, to improve the overall cleanfigure performance we could first call a method that makes sure that no more than a data point falls into a pixel (up to some zoom multiplier) and then release Visvalingam for the precision cut. 

For me, to complete the PR we need to address the input case raised by Egon (and add a few additional notes on the algo with some links, which we can do)

widdma:
In order to address the speed concerns, I've changed `simplifyLine` to discretize  the data to a quarter of a pixel width before performing full path simplification. While this isn't going to be performing true Visvalingam simplification any more, it's certainly a good approximation with the chosen `tol`. 

Simplification on the fig file provided by @okomarov from ACID(47) now performs in under 1s on my current setup (Matlab 2014a, i5-2500K). Simplifying 5e6 points via 
```matlab
x  = linspace(0,2*pi,5e6);
y = sin(x);
plot(x,y);
cleanfigure;
```
takes about 10s. But the bottleneck now appears to be `segmentVisible` rather than the simplification code. It may be worth optimizing this in the future. With this level of performance, I would be much more comfortable enabling this feature by default.

I've also addressed the issue @egeerardyn raised with the `targetResolution` parameter. This is now an array in the format `[W,H,Res]`. With the recent changes to `simplifyLine`, all 3 parameters are now explicitly needed.

okomarov:
Great job, the sine plot with 5e6 cleans in less than a second on my i7-2600!
I am currently reviewing a few minor points and have some questions.

The `[W, H, res]` specification is somewhat elusive to me. 
I guess if the user is (physically) printing the figure, she must have a pretty good idea of the size and the dpi and assuming that the monitor has same resolution you can think of res in terms of PPI, i.e. `[W, H, PPI]` (assuming the unit is inches)?
Then, I think in this context `W` and `H` are somewhat superfluous, in the sense they can be queried from the gca. 
In that context, the zoom multiplier I mentioned in another comment makes sure that full monitor visualization is still at the PPI resolution (and if the user knows what he wants and calls [W,H,res], then you do not need the zoom multiplier). 

I am raising this point because I was thinking to *eventually* allow double syntax on `targetResolution` (in another PR). The base case would be `'targetResolution', PPI` where ppi is scalar and the current 3 element input.

src/cleanfigure.m
```diff
@@ -50,15 +56,18 @@ function cleanfigure(varargin)
 
   % Set up command line options.
   m2t.cmdOpts = m2tInputParser;
-  m2t.cmdOpts = m2t.cmdOpts.addParamValue(m2t.cmdOpts, 'minimumPointsDistance', 1.0e-10, @isnumeric);
   m2t.cmdOpts = m2t.cmdOpts.addParamValue(m2t.cmdOpts, 'handle', gcf, @ishandle);
+  m2t.cmdOpts = m2t.cmdOpts.addParamValue(m2t.cmdOpts, 'targetResolution', [15 9 600], @(a)numel(a) == 3);
```
egeerardyn: Please just write a proper function to validate this.
```matlab
function bool = isValidTargetResolution(val)
    bool = isnumeric(val) && numel(v) == 3;
```
I haven't look at the rest how `NaN` are handled, but if those cause failures, they should be caught at this location as well.

PeterPablo:
Yesterday and today I used `m2t` with this PR and I am very happy with the results and did not run into issues.

okomarov:
I reviewed the PR and added a bugfix and some help/comment edits. Issued a new PR #585 which we can build on if needed (it's on upstream branch too). Closing the current one.

PR #585, opened 2015-03-21
Line simp - review

okomarov:
Review of PR https://github.com/matlab2tikz/matlab2tikz/pull/571 by @widdma.

* Fixes end-point removal during pixelation, check on
```matlab
x = linspace(0,2*pi,5e6);
x(randi(5e6, [1e4,1])) = NaN;
y = sin(x);
h = plot(x,y);
cleanfigure;
```
* Mostly help edits
* Pre-allocation of `linesx`, `linesy`
* `featureArea` code **untouched**

Once everyone is ok, we can merge. 

I'd like an opinion on the **scalar input syntax** to 'resolutionTarget' advanced in https://github.com/matlab2tikz/matlab2tikz/pull/571#issuecomment-83539045

### DIFF
--- a/src/cleanfigure.m
+++ b/src/cleanfigure.m
@@ -6,8 +6,8 @@ function cleanfigure(varargin)
 %   CLEANFIGURE('handle',HANDLE,...) explicitly specifies the
 %   handle of the figure that is to be stored. (default: gcf)
 %
-%   CLEANFIGURE('minimumPointsDistance',DOUBLE,...) explicitly specified the
-%   minimum distance between two points. (default: 1.0e-10)
+%   CLEANFIGURE('targetResolution','WxH@Res',...) explicitly specify the
+%   target resolution of the path simplification. (default: '15x9@600')
 %
 %   Example
 %      x = -pi:pi/1000:pi;
@@ -50,15 +50,15 @@ function cleanfigure(varargin)
 
   % Set up command line options.
   m2t.cmdOpts = m2tInputParser;
-  m2t.cmdOpts = m2t.cmdOpts.addParamValue(m2t.cmdOpts, 'minimumPointsDistance', 1.0e-10, @isnumeric);
   m2t.cmdOpts = m2t.cmdOpts.addParamValue(m2t.cmdOpts, 'handle', gcf, @ishandle);
+  m2t.cmdOpts = m2t.cmdOpts.addParamValue(m2t.cmdOpts, 'targetResolution', '15x9@600',@isstr);
 
   % Finally parse all the elements.
   m2t.cmdOpts = m2t.cmdOpts.parse(m2t.cmdOpts, varargin{:});
 
   % Recurse down the tree of plot objects and clean up the leaves.
   for h = m2t.cmdOpts.Results.handle(:)'
-    recursiveCleanup(meta, h, m2t.cmdOpts.Results.minimumPointsDistance, 0);
+    recursiveCleanup(meta, h, m2t.cmdOpts.Results.targetResolution, 0);
   end
 
   % Reset to initial state.
@@ -67,7 +67,7 @@ function cleanfigure(varargin)
   return;
 end
 % =========================================================================
-function indent = recursiveCleanup(meta, h, minimumPointsDistance, indent)
+function indent = recursiveCleanup(meta, h, targetResolution, indent)
 
   type = get(h, 'Type');
 
@@ -93,7 +93,7 @@ function indent = recursiveCleanup(meta, h, minimumPointsDistance, indent)
   if ~isempty(children)
       for child = children(:)'
           indent = indent + 4;
-          indent = recursiveCleanup(meta, child, minimumPointsDistance, indent);
+          indent = recursiveCleanup(meta, child, targetResolution, indent);
           indent = indent - 4;
       end
   else
@@ -114,7 +114,7 @@ function indent = recursiveCleanup(meta, h, minimumPointsDistance, indent)
           movePointsCloser(meta, h);
           % Don't be too precise.
           % coarsenLine(meta, h, minimumPointsDistance);
-          simplifyLine(meta, h);
+          simplifyLine(meta, h, targetResolution);
       elseif strcmpi(type, 'stair')
           pruneOutsideBox(meta, h);
       elseif strcmp(type, 'text')
@@ -368,7 +368,7 @@ function coarsenLine(meta, handle, minimumPointsDistance)
   return;
 end
 % =========================================================================
-function simplifyLine(meta, handle, varargin)
+function simplifyLine(meta, handle, targetResolution)
   % Reduce the number of data points in the line handle.
   % this can help with plots that contain a large amount of data points not
   % all of which need to be plotted.
@@ -379,7 +379,7 @@ function simplifyLine(meta, handle, varargin)
   zData = get(handle, 'ZData');
 
   if ~isempty(zData)
-    % Don't do funny stuff when zData is present.
+    % Don't simplify 3d plots
     return;
   end
   if isempty(xData) || isempty(yData)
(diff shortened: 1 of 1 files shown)


widdma:
Factoring out the "pixelate" function has broken simplification of log scaled axes.

widdma:
The `ZOOM_MULTIPLIER` variable used differently in `pixelate` and the calculation of `tol`:
In `pixelate`, pixels are broken into `ZOOM_MULTIPLIER^2` sub-regions; whereas, in the calculation of `tol`, pixels are broken into `ZOOM_MULTIPLIER` regions.

This was intentional in my code, as we don't want to oversimplify in the 'pixelate' stage, just remove points that are most likely to be removed by the path simplification. Perhaps they should be considered as distinct parameters.

okomarov:
You are right I was confused, thanks for double-checking. The  regression in the log scales should be fixed now.

Shouldn't it be ZOOM_MULTIPLIER^2 in tol?

If you have a square with side = 2 and if you apply a 3x zoom  then you get a 6by6 square = 36 = 4* (3x^2).

widdma:
@okomarov great. 

> Shouldn't it be ZOOM_MULTIPLIER^2 in tol?
If you want to apply `pixelate` and path simplification at the same scale, then yes. However, because the 'pixelate' routine is very crude in it's simplification I think it's better to apply it at a different zoom level to the path simplification.

okomarov:
@widdma Ok, then if we want to *ensure* a 4x zoom, then I'd say let's do a 4x*2 pixelate and a 4x path simplification.

widdma:
>  4x*2 pixelate and a 4x path simplification.

Do you mean 4^2 pixelate, the way it currently is?

I'd say the parameters are pretty good as they are. From my tests, with the current `tol`, the simplified plot is indistinguishable from the original at the given figure size. If the user desires larger images, say they want to zoom-in, they can increase the resolution.

For the `targetResolution` parameter, perhaps it's going to most intuitive for the user to simply specify a target size in pixels? Say `[W,H]` in pixels with a default of `[9000,5400]`.

For your scalar input syntax, where would you get the width and height from? Just the on-screen dimensions?

okomarov:
I think we have a different understanding about zoom, I have in mind http://en.wikipedia.org/wiki/Image_scaling#Comparison

As is, line simplification is not coherent since it does not square the multiplier in `tol`. From the example I posted above in the edit, you can see that at 300 ppc the plot is smooth (original size), but once you zoom in by 4x, the smoothness is lost. This does not (almost) happen if tol uses the same mult^2.
I must admit that I did not notice the incongruence at first, but only once you pointed it out.

For the scalar input syntax you can query the size of the figure by first setting the units to cm.

src/cleanfigure.m
```diff
@@ -298,73 +305,300 @@ function pruneOutsideBox(meta, handle)
...
+        xrange = (log10(a(2))-log10(a(1)));
+    end
+    if ~isYlog
+        yrange = (a(4)-a(3));
+    else
+        yrange = (log10(a(4))-log10(a(3)));
+    end
+    tol = xrange*yrange/(ZOOM_MULTIPLIER^2*prod(targetResolution));
+    nPixelsX = targetResolution(1)*sqrt(targetResolution(3));
+    nPixelsY = targetResolution(2)*sqrt(targetResolution(3));
```
widdma: I realise I made a mistake here. It's generally accepted to use resolution in pixels/points per (linear) length, not per area. We should change `sqrt(targetResolution(3))` to `targetResolution(3)` and `prod(targetResolution)` to `prod(targetResolution)*targetResolution(3)` to be consistent with standard use.

Alternatively, we go with as I suggested earlier and make `targetResolution` as simply the width and height of the target figure in pixels. This seems like a simpler interface.

widdma:
Ok, I think we have different ideas about what this parameter is controlling. 

In my original code, the factor of 4 in the calculation of `tol` was chosen to perform *maximum* simplification without loss of quality *at the target resolution*.  It was not chosen to, nor do I expect it to, allow zooming-in by a factor of 4 without loss of quality. 

If you want `ZOOM_MULTIPLIER^2` in the calculation of `tol`, I would suggest setting `ZOOM_MULTIPLIER = 2` so we are performing maximum simplification without loss of quality.

okomarov:
@widdma As a general approach I prefer to be safe when it comes to features that the user can potentially abuse. I still want to grant a good experience over a `state-of-the-art` implementation (which `featureArea` is btw, and in fact that part is untouched).

What I mean is, the 3 number input has a default that the user might be unaware of, or unsure on how to correctly use it (as earlier in my case). Hence, if the user goes full screen with his figure and sees piecewise he might be unhappy. In this context, a zoom multiplier is poor man solution of getting a buffer on thew simplification. It avoids extensive testing by just keeping it safe. 

Now, if the input was solely PPI, that is something that clicks immediately (at least for me). It is straightforward to just increase resolution. We don't need a zoom buffer and can use maximum simplification. If the user is unhappy, he just needs to increase one number and that is easy to convey in the help.

If you agree, I would give a try to the one scalar syntax and retrieve the inches measurement of the axis.

okomarov:
Also, I prefer the `[PPI]` over the `[W, H]` because I would like to keep the `cleanfigure` outside the workflow that decides the dimension of the figure which in principle belongs to the actual matlab plotting code even before `m2t`, with cleanfigure playing a more casual/sporadic role.

widdma:
Ok let's just go with a `PPI` option and query the figure for dimensions. It is certainly more clear in it's behaviour. Perhaps with a note saying the user may wish to set the figure size in physical units for consistent results. 
Eg. via 
```matlab
set(h, 'units', 'centimeters', 'pos', [0 0 15 9])
```

However, this will be somewhat cumbersome if you desire figures larger than your screen size. This is also tricky for me as I use a tiling window manager, so window size is constrained by default. These reasons make me desire a `[W,H]` syntax, but I understand they're both quite unusual.

PeterPablo:
@widdma, virtually doubling your screen size could then be done by doubling the `PPI`?! The easier the syntax, the better I'd say. 

Furthermore the above discussed *safety factor of permissible zoom* is a nice safety net. I happen to sometimes alter the axis limits afterwards. It would be a shame to redo the process of exporting... As @okomarov showed, the number of retained data points is still low, opposed to the raw data vector.

okomarov:
@widdma Let me know if the (second) syntax works as you would expected. 

> Perhaps with a note saying the user may wish to set the figure size in physical units for consistent results.
...
However, this will be somewhat cumbersome if you desire figures larger than your screen size. This is also tricky for me as I use a tiling window manager, so window size is constrained by default. These reasons make me desire a [W,H] syntax, but I understand they're both quite unusual. 

I am not totally sure what you mean (do you have a practical examples in mind?). 

Help sections are not yet updated to reflect the new syntax.

widdma:
@okomarov I'm happy with the second syntax. 

My example is this: due to the window manager I use, changing the window size for the figure is not straightforward, or at least not easily scriptable. Because of this, the dimensions returned by `get(h,'pos')` are not related to the eventual size I expect the figures to be rendered in TeX. 

Typically, I create full `\textwidth` figures on A4 paper with a 5:3 ratio, which means the figures come out around `15cm x 9 cm`. A resolution of 200 ppcm is plenty, so I can now simply call `cleanfigure('targetResolution',[15,9]*200)`

widdma:
Contrary to what I said earlier, it seems we can reduce `ZOOM_MULTIPLIER` to 1 without really affecting the final quality, and it certainly improves the speed.

Other than that, I'm happy with this PR.

okomarov:
@widdma The ZOOM_MULTIPLIER is only used by `pixelate()`, to perform the rough simplification at a higher precision and let featureArea do the precision work. We could reduce it to 2 (and maybe rename the coefficient?) which reduces the runtime in half from 1.6 to 0.8. Setting it to 1, takes 0.6 seconds, just slightly faster.

okomarov:
I added some refactorings to address:
* Egon's comments on the avriable `PPI=1` being misleading
* a fix for when line has markers.

   ```matlab
x = linspace(0,2*pi,5e6); 
h = plot(x,ones(1,numel(x)),'-x');
```
   It would have been simplified to two points only.

**Comment by okomarov** [comment]
This PR is ready to be merged.

**Comment by egeerardyn** [comment]
@okomarov Can you make sure that Travis passes the test? I don't mind taking on the MATLAB toolboxes that you don't have, but I don't have the time to do the complete test suite.

**Comment by okomarov** [comment]
I will produce Octave hashes but on Win platform unless there is some turnkey soution with Linux virtual machine + Octave that you know of.

**Comment by egeerardyn** [comment]
Hopefully, platform shouldn't matter (I use OS X mostly to generate hashes, so not even Linux). If it doesn't match, then the test should be marked unreliable for the time being

**Comment by okomarov** [comment]
@widdma I was giving one last graphical inspection to the `testGaphical` and Viswalingam's algorithm is weak on null-area point that which are single-coordinates moves (this problem affects `ACID(30) - roseplot`). 

The problem is illustrated below, i.e. it should not simplify. 

untitled


```matlab
subplot(122)
plot([1,2,2,2,3],[1,1,2,1,1]), hold on
plot([4,4,4,5],[1,2,1,1])
axis([0,6,0,3]);
cleanfigure
title 'after cleanfigure'

subplot(121)
plot([1,2,2,2,3],[1,1,2,1,1]), hold on
plot([4,4,4,5],[1,2,1,1])
axis([0,6,0,3]);
title original
```

I was unable to solve the problem with the current tools. What I have tried:
* After the first `featureArea` I take null-area points and its *skip-n* neighbours, i.e. not the points immediately in its vicinity but skipping one, two,... closest neighbours. This does not fix all illegit null-area points and I do not have a justification for how many neighbours to skip.
* I tried to mix the criterion on the area with the mask of an unzoomed pixelation. However, this method retains way too many points if the line is perfectly straight, i.e. all null-area points that fall on a new pixels will be retained instead of just 2.

egeerardyn:
@okomarov: Nicely spotted. I hadn't even considered that edge case, but it looks like a nasty one (well, not really impossible to come by in the wild, so an important one).

widdma:
@okomarov: good point! This does seem to be a pathological example for Viswalingam simplification which is quite hard to fix. 

Perhaps we should change simplification algorithm. I've look at a few others: 
  - The Ramer–Douglas–Peucker algorithm seems to be fairly common and will give comparable results to Viswalingam. The drawback is it's recursive, so will no have great performance in Matlab, and it's worst case is `O(n^2)`, although average `O(n log(n))`.
  - The Opheim or Reumann–Witkam algorithms are very easy to implement and are `O(n)` runtime, although their simplification will not be as optimal as Viswalingam or RDP. The edge case @okomarov mentioned above also seems to cause similar problems for these algorithms, however I have devised a way to rectify this.

I've implemented a modified version of the Opheim algorithm in the line-simp-opheim branch of my fork. I'll create a pull request to merge it into this branch.

While, in principle, this simplification algorithm will not give as good results as Viswalingam, I haven't noticed too much of a difference in practice. As a bonus, It's also somewhat quicker.

okomarov:
Continuing in https://github.com/matlab2tikz/matlab2tikz/pull/613

Issue #575, opened 2015-03-12
Make-like tool for testing


egeerardyn:
As I suggested in the last few comments of  #470, i.e. https://github.com/matlab2tikz/matlab2tikz/issues/470#issuecomment-75934867 and https://github.com/matlab2tikz/matlab2tikz/issues/470#issuecomment-75953009 it would be useful to have a wrapper tool for ALL testing facilities that also automates some of the aspects. I propose to call this tool `m2tdo` for brevity (unless someone has a nice alternative name)

There are a few separate actions that need to be accessible:
 - changing directory to `./test` (and back after completion)
 - adding `m2t` and `./test/suites` to the path
 - running `make`(e.g. `make distClean` and `make acid`) in different locations (as originally #470)
 - run the headless tester
 - run the graphical tester
 - managing the hash table
 - show the PDF of the output
 - maybe even some git integration as suggested in #506 (that tool should eventually become callable from this one anyways)

The calling syntax I had in mind was something (readable) like:
```matlab
m2tdo make distclean
m2tdo test graphical
m2tdo make pdf
% etc.
```

Each of these different actions should be callable by themselves, but we should also provide logical combinations of these actions, such that the common things we need to do are easily available.
E.g. a command like `m2tdo run graphical` would be equivalent to running a script:

```matlab
m2tdo addpaths
m2tdo cd test
m2tdo make distclean
m2tdo test graphical
m2tdo makepdf
m2tdo open pdf
m2tdo restorepath
m2tdo cd back
```

I'm not 100% certain about the syntax yet, so all suggestions are welcome.

But this way, we could just instruct new developers to run `m2tdo run graphical`, m2tdo run headless` and maybe 2 or 3 other of the "script" commands instead of having to explain every step. Since this is mainly a wrapper, we can keep the responsibilities of the different tools separated (which was the problem in #470 and we still have a few similar things that should be addressed).

PR #595, opened 2015-03-23
Reinitialize hashes as indicated in `getReferenceHash()`

PeterPablo:
see https://github.com/egeerardyn/matlab2tikz/commit/583251956adf3804577c27865dd90e15ee8043f8
I had the same problem, so went ahead and fixed this.
Personally I think that https://github.com/matlab2tikz/matlab2tikz/issues/575 is unrelated, but want to add the reference, in case this should be reviewed during the restructuring.

### DIFF
--- a/test/private/execute_hash_stage.m
+++ b/test/private/execute_hash_stage.m
@@ -28,7 +28,7 @@ function hash = getReferenceHash(status, ipp)
     % By storing the hash table in a persistent variable, the amount of disk
     % operations is minimized (i.e. reading the file, parsing it and storing its
     % data in a MATLAB struct), as this is only done once a new test suite is
-    % executed. To clear this persistent storage, run |clear functions|.
+    % executed. To clear this persistent storage, run `clear getReferenceHash`.
 
     if isempty(hashTable) || ~isequal(hashTable.suite, ipp.Results.testsuite)
         hashTable = loadHashTable(ipp.Results.testsuite);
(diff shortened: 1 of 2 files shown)


egeerardyn:
I like this very much. At the end I just ended up doing `clear functions; testHeadless(n)`. This is better, but as you indicated, should eventually go into #575.

test/private/execute_hash_stage.m
```diff
@@ -28,7 +28,7 @@
     % By storing the hash table in a persistent variable, the amount of disk
     % operations is minimized (i.e. reading the file, parsing it and storing its
     % data in a MATLAB struct), as this is only done once a new test suite is
-    % executed. To clear this persistent storage, run |clear functions|.
+    % executed. To clear this persistent storage, run `clear getReferenceHash`.
```
egeerardyn: Maybe we should decide on which markdown we chose: GitHub flavor (currently) or MATLAB flavor (before this change). I have no real preference, but we might be able to generate documentation using MATLAB flavored markdown in MATLAB itself.

PeterPablo: I wasn't aware that this is MATLAB markup -- never have seen it before and thus changed it. Thus I certainly don't have hard feelings about this being changed to something else.

egeerardyn: See http://nl.mathworks.com/help/matlab/matlab_prog/marking-up-matlab-comments-for-publishing.html for MATLAB Markdown. I'm not particularly fond of that notation either (I like GFM better in general), but it might be more convenient in the long run perhaps.

PR #603, opened 2015-04-08
improve behavior of testsuite on Windows octave

PeterPablo:
See https://github.com/matlab2tikz/matlab2tikz/pull/602 for discussion.

### DIFF
--- a/test/private/testMatlab2tikz.m
+++ b/test/private/testMatlab2tikz.m
@@ -73,10 +73,18 @@ function status = testMatlab2tikz(varargin)
   ipp = sanitizeInputs(ipp);
 
   % -----------------------------------------------------------------------
-  if strcmpi(env, 'Octave') && ~ipp.Results.figureVisible
-      % Use the gnuplot backend to work around an fltk bug, see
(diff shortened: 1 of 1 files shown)


PeterPablo:
@okomarov, could you please test this? If you agree, please go forward and merge. This change does not influence Travis, as it only matches Windows+Octave.

okomarov:
It improves things quite a bit! From 13 unmatching to 7. No effect on the 27 unreliable ones.

PeterPablo:
The other six match now? That is nice, considering this is such a minor change -- thanks to your investigation and `DIFF`ing.

okomarov:
Yes, 6 hashes match now. I reckon the remaining 7 do not match due to numerical differences.

PeterPablo:
Please go ahead and upload your output for one of the failing ones in the discussion of PR #604. Please choose a testcase with comparatively small output and once with the PR applied and once without.
PR #604 should solve numerical differences for numbers close to `eps`, though does NOT solve problems related to different figure dimensions (we need a separate solution).

PR #612, opened 2015-04-09
do not store hashTable as persistent

PeterPablo:
* oddly this results in a strong speedup of testsuite execution
* this eases the developers life, since now updated hashes in the file are always noticed
* `clear getReferenceHash` only worked in Matlab
* `clear functions` has unwanted side effects (`getEnvironment` could not be found anymore)

see https://github.com/matlab2tikz/matlab2tikz/commit/05520b8850b0e1a1a9d42f476d8ef5e06a7a04a6#commitcomment-10642865

### DIFF
--- a/test/private/execute_hash_stage.m
+++ b/test/private/execute_hash_stage.m
@@ -22,17 +22,9 @@ function [status] = execute_hash_stage(status, ipp)
 end
 % ==============================================================================
 function hash = getReferenceHash(status, ipp)
-    % retrieves a reference hash from a persistent hash table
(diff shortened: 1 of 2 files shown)


PeterPablo:
@egeerardyn: You seem to agree. Please go ahead and merge.

egeerardyn:
:+1: Only thing, I would add is that maybe you can add a bit of comments there that `persistent` is actually slower than just reading this, such that I'm not tempted to make the same mistake again :-)

PR #604, opened 2015-04-08
change float format

PeterPablo:
`%8.3g` ensures:
* same format as before
* less precision to prevent possible issues between different environments (see PR #602 for discussion)

Together with PR #603, this should improve stability of the test suite on various environments. I will update the hashes, once the other developers agree with the approach.

PR #590, opened 2015-03-23
introduce font encoding

PeterPablo:
as discussed in issue #587

Issue #552, opened 2015-02-27
Resolve hash mismatches between different platforms and program versions


PeterPablo:
See #485; #542 is related

Let's first collect the tex output from the `converted` folder and then start a systematic approach to eliminate the issues leading to md5 hash mismatches.

- [x] Color definitions are present in R2014b --> we might set the `defaultAxesColorOrder` (`get(0,'defaultAxesColorOrder')`, ref [here](http://de.mathworks.com/help/matlab/creating_plots/defining-the-color-of-lines-for-plotting.html)), fix at #641 
- [x] `width`, `height` and `at` don't match concerning their values, influenced by `ScreenPixelsPerInch`, fix at #641 
- [ ] `xtick` values differ; this is expected since octave here simply behaves differently. Course of action?
- [ ] `xlabel style`. `title style` and `legend style` is set explicitly/differently for R2014b (#542 could resolve this)
- [ ] `xlabel` is parsed differently. One time newlines are printed and once `\\[1ex]` is added

If we want to stick to md5sums for comparison, we have to tackle all of those bullet points. How should we proceed with the differing `xtick` values?

okomarov:
With reference to the mismatching hashes on R2014b - Win7 64b reported in https://github.com/matlab2tikz/matlab2tikz/pull/485#issuecomment-76407920, find attached the zipped converted folder: https://www.dropbox.com/s/6scbblpjiemsv4v/Win7_R2014b_converted.zip?dl=0

PeterPablo:
Did you test with https://github.com/egeerardyn/matlab2tikz/commit/a713c6656537c57427c90da2fdea84f6c6723db2?

PeterPablo:
Alright. Since the updated test environment is now in use, we may begin...
I ran `testGraphical(1)` on ubuntu with both `octave 3.8.2` and `MATLAB R2014b` and diffed the output:
```
--- test1_R2014b.tex	2015-03-03 10:05:04.811623309 +0100
+++ test1_octave.tex	2015-03-03 10:03:11.491625188 +0100
@@ -9,33 +9,31 @@
 \usepackage{amsmath}
 
 \begin{document}
-\definecolor{mycolor1}{rgb}{0.00000,0.44700,0.74100}%
-\definecolor{mycolor2}{rgb}{0.85000,0.32500,0.09800}%
-%
 \begin{tikzpicture}
 
 \begin{axis}[%
-width=4.568421in,
-height=3.463657in,
[...]

PR #614, opened 2015-04-10
remove function calls to SVD

### LINKED ISSUE #414: Use look-up table for "random" data in test functions

PeterPablo:
* introduce sub function that returns a small data table
* all svd calls are replaced
* this should hopefully partially resolve issues, observed e.g. in #590 and #604

Should partially fix #414.


PeterPablo:
Failing tests:

```
--- mine/test25-converted.tex
+++ (clipboard)
@@ -12,7 +12,7 @@
 \begin{tikzpicture}
 
 \begin{axis}[%
-width=0.95092\figurewidth,
+width=0.75\figurewidth,
 height=\figureheight,
 at={(0\figurewidth,0\figureheight)},
 scale only axis,
```
Unfortunately `width` is determined differently. I raised this issue in #https://github.com/matlab2tikz/matlab2tikz/issues/552#issuecomment-76918634.

PeterPablo:
How should we proceed?

okomarov:
I reckon there are two things to do first:
* Define width and height explicitly as in (but without the float format change) https://github.com/PeterPablo/matlab2tikz/commit/9e31d1bf57d9ca632fc935e8b46752476f5a9608
* Remove instances of `hist()`. TMW is deprecating it and one reason is [inconsistent behaviour](http://uk.mathworks.com/help/matlab/creating_plots/replace-discouraged-instances-of-hist-and-histc.html) (wathever that means) 

Basically, this PR will then address partially #552 and #414.

PeterPablo:
Sounds reasonable. Good point especially about removing `hist()`.
In order to introduce explicit `width` and `height` we have to adapt https://github.com/PeterPablo/matlab2tikz/commit/9e31d1bf57d9ca632fc935e8b46752476f5a9608 to not pass latex variables, but rather directly supply dimensions -- otherwise the width still contains a calculated value.
So, should I first focus on this PR?

okomarov:
I think so, but I am afraid is gonna be hard :sweat_smile: now that I realize what it entails.

PeterPablo:
:-/ A lot more hash calculation fun... I really hope it is worth the effort and we are rewarded by testsuite stability. Actually I would not remove calls to `hist()` here, but rather do this as a second PR, to not introduce too many changes in one go. I will though (re)label them as *unreliable* and link to your comment.
Furthermore I will adapt the above linked commit of mine to incorporate fixed dimensions. * crossing fingers this does not introduce new issues *

PeterPablo:
Oh, I'll wait for a short comment of @egeerardyn, so we agree on course of action.

egeerardyn:
I'm on holiday until Sunday, so I will try to have a look early next week.

egeerardyn:
With regards to `hist`, I'm on the one hand very happy that TMW is deprecating it (it has caused a lot of indeed inconsistent behavior, also in our codebase: the internals  of `hist` are a terrible mess).

Anyhow, I think we should keep our limited `hist` support at least for one more year. Most people I know have not migrated to R2014b, let alone R2015a (where `hist` has been deprecated), so we still need to keep their condition in mind. E.g. here in my lab it is unfeasible to migrate completely to R2014b since some of our own toolboxes are still not compatible and I can imagine that other labs have similar issues.

When (not if) we remove `hist`, it should happen in a different PR, regardless of the actual time when we do this.

Regarding the axes dimensions: I think that we always calculate their size, as long as it is specified in a valid LaTeX format, so that course of action is probably not helping anything. But let's discuss this in a separate issue report and keep this one focussed on the current changes.

PeterPablo:
Let's keep this PR's scope limited to `SVD`. `width` and `hist()` will be dealt with separately.

I restored the `unreliable` status of two tests, though added a comment about why this test is unreliable.
I removed the `unreliable` flag of `bars` (= `ACID(34)`). Please test, if hashes match for you!

If no further issues arise, I consider this ready.

PeterPablo:
Please let's first get #590 in. I will then merge `develop` here and update the hashes. Afterwards somebody has to verify `ACID(34)` and this is then ready to go in.

PeterPablo:
I rebased on `HEAD`, updated the hashes and marked one more test unreliable and linked to the issue/discussion. Please test `ACID(34)` on Matlab R2014(a|b) and merge, as soon as Travis ran successfully.

PR #616, opened 2015-04-14
Change files root

### LINKED ISSUE #600: Clean up top-level files
I think at the moment we have a bit too many files in our file tree:

```
.
├── AUTHORS
├── COPYING
├── ChangeLog
├── Makefile
├── README.md
├── THANKS
├── logos/
├── src/
├── test/
└── version-0.6.0
```

I think the following should be done:

* [x] `Makefile` is used for releases, which are just a ZIP file from the latest master branch. I think the same can be done from within the [GitHub interface](https://github.com/matlab2tikz/matlab2tikz/releases/new).
* [x] I think `version-0.6.0` seems to be unused since #476, so it should disappear.
* [x] `THANKS` and `AUTHORS` should just be merged as `AUTHORS` and could probably use some updating. E.g. at https://github.com/django/django/blob/master/AUTHORS is an example of how this might look.
* [x] `ChangeLog` should be renamed to `CHANGELOG.md`
* [x] `COPYING` should be renamed to `LICENSE.md` and updated
* [x] Probably all those extra files should be written in MarkDown (and maybe even with the `.md` extension to signal this).

egeerardyn:
This fixes #600.


PeterPablo:
:+1: Looks good! Go ahead and merge, if you think it should go on. Travis is happy, since no code was changed.

egeerardyn:
I forgot to disable Travis for this PR, since it is indeed nonsensical in this case.

PeterPablo:
No problem. Better be safe than sorry.

PR #620, opened 2015-04-15
Error handler should just output message

### LINKED ISSUE #611: Warning: "unrecognized escape sequence"
I get this when running `testHeadless` in octave on windows. This spams the output tremendously. [Here](http://octave.1599824.n4.nabble.com/Oct-File-helo-world-sample-td4657087.html) an explanation for the cause is given and possible solutions.

We should identify the location that causes the warning and fix it.

egeerardyn:
There is no need in interpreting it in any environment, as there
should not be any escape sequences in there anyway.
This fixes #611.
There has been no formal testing, only visual since this happens within an error path and not within normal execution.

### DIFF
--- a/test/private/errorHandler.m
+++ b/test/private/errorHandler.m
@@ -33,15 +33,9 @@ end
 % ==============================================================================
 function disp_error_message(msg)
     stderr = 2;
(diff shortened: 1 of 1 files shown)


okomarov:
:+1:

PR #629, opened 2015-04-24
Added 'mesh/ordering=y varies' option in drawSurface

josombio:
When plotting a surface, both the mesh/rows and mesh/ordering have to be specified.
http://tex.stackexchange.com/questions/198639/how-can-i-import-matrix-data-2d-function-from-matlab-into-pgfplots

If not, in some cases the resulting surface is incorrect. See the following example:
pre_fix
And now with the fix
post_fix

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -2234,6 +2234,9 @@ function [m2t,env] = drawSurface(m2t, handle)
     % data writing to one single sprintf() call.
     opts{end+1} = sprintf('mesh/rows=%d', numrows);
 
+    % Add mesh/ordering=y varies for specifying the ordering of the data
+    opts{end+1} = 'mesh/ordering=y varies';
+    
     opts = join(m2t, opts, ',\n');
     str = [str, sprintf(['\n\\addplot3[%%\n%s,\n', opts ,']'], plotType)];
 


PeterPablo:
@josombio, thank you very much for your contribution!

I  can see the improvement on the screenshot. Could you please supply us with some minimal matlab code, so we can reproduce the issue? Ideally we incorporate this in one of the unit tests. Probably in a short while Travis will fail on some test. This is then a good candidate to add the `stat.issue` field.

Did you check the testsuite for regressions?

PeterPablo:
One more thing -- could you please create the pull request against the `develop` branch? Thank you.

PeterPablo:
We certainly need to be cautious when applying this option. I manually added it to one of the plots I created with `m2t` and it was distorted afterwards. As such, it seems the current table creation is suited for `x varies`.
Did you encounter that issue while exporting with `m2t`?

egeerardyn:
I have the impression that both `x varies` and `y varies` will result in some plots that are going to be off. I.e. neither is inherently better than the other.

As such, I'd vote to keep the default value provided by pgfplots until we can determine programmatically which of both is appropriate. Unfortunately, this means we should not merge this code and that manual editing will still be needed in the meanwhile.

PeterPablo:
I also would not merge this submission, as long as we can make a clear distinction when this is necessary. Thus, @josombio please provide minimal Matlab code that results in a problematic plot, so we can further investigate the underlying issue.

egeerardyn:
**Conclusion:** @josombio We really appreciate your PR, but unfortunately I think it is unwise to forcefully set this setting. Let's continue the discussion in #635 to see whether we can come up with a smarter way to set these things.

PR #630, opened 2015-04-26
fix3Daxes

LJeub:
sets `plot box ratio` only for 3D plots and merges changes from develop

(see also PR #560)

I haven't had a chance to deal with the hashes yet. I haven't had to deal with that particular part of the code before, so it would take me a bit of time to get up to speed. Could anyone else check what needs to be done there?


egeerardyn:
I will have a look at this, somewhere tonight or tomorrow and merge in the changes and make sure the hashes are adapted.

Issue #553, opened 2015-02-27
pgfplots 1.12 is incompatible with lualatex


PeterPablo:
See #550, especially this [comment](https://github.com/matlab2tikz/matlab2tikz/pull/550#issuecomment-76357216)

Switching to `xelatex` would require to remove the `\pdfminorversion` statement and review `ACID(47) zplanePlot1`.

okomarov:
Also, color-interpolated patches do not render with Lualatex.

PeterPablo:
Do you have a pointer to a test case at hand, so we can collect specific issues in a list?

egeerardyn:
I think the problem is with the new Lua back-end in `pgfplots 1.12`.
That new back-end is supposed to get the calculations more precise and fast; but it seems that it doesn't support the whole gamut of features in pgfplots.

If I add `\pgfplotsset{lua backend=false}` to the document, `lualatex` is able to compile everything without a problem.

PeterPablo:
It says
```However, it only covers parts of the feature set:
some features of pgfplots are unsupported by the lua backend. In this case, it will be switched off
automatically and the TEX implementation will be used as fallback.```

As such, I'd say this is a bug.
`lua backend=false` is a viable workaround. I'll add the label to this issue.

egeerardyn:
I've reported this bug to the pgfplots project as [bug 78](https://sourceforge.net/p/pgfplots/bugs/78/).

PeterPablo:
What is our course of action? Stick with `lualatex` as the next release of `pgfplots` has this problem fixed?

egeerardyn:
We should stick with `lualatex`. Once there is a stable [`pgfplots`](http://pgfplots.sourceforge.net) release with this fix available, we can close this issue.

PeterPablo:
What about the [comment](https://github.com/matlab2tikz/matlab2tikz/issues/553#issuecomment-76416410) of @okomarov? (I did not check this myself.)

If we intend to stick to `lualtex`, with what I am fine, then why not close right away? Search will direct here anyhow.

egeerardyn:
I prefer to keep it open (i.e. more visible to new users) as long as the dependencies are not easily available to those users. An unstable TDS release is available, but I consider that a bit too difficult to incorporate by most users.

I will have a go at reproducing the bug Oleg talks about and report it to pgfplots.

egeerardyn:
[...]

PR #631, opened 2015-04-28
Fix 3D Axis (rebased)

egeerardyn:
I've rebased both previous PRs on top of develop, i.e. this PR should fix #630 and #560. 

For R2014a and R2014b, I've ran the headless test suite and I updated the hashes accordingly. Double checks are of course appreciated (also on the code, due to the rebase).
My non-matching outputs are here: 
[R2014b](https://gist.github.com/egeerardyn/fea3b1a6f39f550775b8): only `[56 82]`non-matching reliable tests
[R2014a](https://gist.github.com/egeerardyn/d451a9bde99340df6a5e): only `[9]` was non-matching

For Octave 3.8, I'm running into problems since `view()` does not return a transformation matrix there, but just the `[azimuth elevation]` vector. I will inspect whether I can transform this reliably into the right form. But right now, this is not-working in Octave due to this issue.

e.g. in Octave `ACID([56])` fails with
```
error: operator *: nonconformant arguments (op1 is 1x2, op2 is 4x1)
error: Octave:nonconformant-args
error: called from:
error:   /home/travis/build/matlab2tikz/matlab2tikz/test/../src/matlab2tikz.m at line 4855, in function matlab2tikz>getRelativeAxesPosition
```
due to this different behavior.

TODO list:
 - [x] convert azimuth, camera vector, elevation, ... into transformation matrix
 - [x] check on Octave and update hashes
 - [x] double check output of previously failing tests


egeerardyn:
Octave non-matching (reliable) tests are the `[29:31]` as in https://gist.github.com/5dd3f87a6cc5319ec0ff.

I just need to verify the output by hand for all environments to ensure the data seems valid.

egeerardyn:
I've executed the graphical tests in
 - [R2014a](https://dl.dropboxusercontent.com/u/452005/m2t/fix-axis-3d/acid2014a.pdf)
 - [R2014b](https://dl.dropboxusercontent.com/u/452005/m2t/fix-axis-3d/acid2014b.pdf)

In Octave 3.8, I had some hangs, but I'm unsure it is due to our code or just stability issues of my Octave. In any case, Travis does not complain, so for me that's enough.

To me, the output looks fine. Only bad result seems `47` (which is compiled with LuaLaTeX and unpatched pgfplots 1.12). All the others are compiled with pdflatex to avoid the same bug (i.e. #553).

PR #643, opened 2015-05-06
Require pgfplots 1.12 for `z buffer=sort`

### LINKED ISSUE #504: Mesh legend does not support zbuffer sort and ACID(58) fails
When double-checking current feature branches, I noticed that `ACID(58)` in MATLAB R2014a produces output that cannot be compiled by a LaTeX compiler by our latest develop branch (i.e. 285dc83 )

In the error log, one of the symptoms is:

```log
! Undefined control sequence.
<argument> ...ots@current@point@x ,\pgfplots@current@point@y 
                                                  ,\pgfplots@current@point@z 
l.257 \end{axis}
```

I haven't inspected this further, this issue report is just to make sure we do not forget this.

egeerardyn:
This fixes #504.

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -2561,7 +2561,7 @@ function [m2t,str] = drawSurface(m2t, h)
     [dx, dy, dz, numrows] = getXYZDataFromSurface(h);
     m2t = jumpAtUnboundCoords(m2t, [dx(:); dy(:); dz(:)]);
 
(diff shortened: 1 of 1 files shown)


PeterPablo:
So this should simply show a warning to the user? Anything I should test in particular?

egeerardyn:
I tested on R2014a without any error messages during execution. So indeed, I also assume that it works fine in the wild. There's not a lot that can be tested :-)

Issue #649, opened 2015-05-08
testGraphical on octave


PeterPablo:
at 7a6b248655f555c1411e7397e47bfa3e50eb8050

This is difficult to accomplish, since the reference output does not seem to be saved, as long as figures are invisible. I think @okomarov had the same problem yesterday.

Unfortunately I seem to be unable to pass `'figureVisible','true'` to `testGraphical()`. @egeerardyn, can you please verify?

By this small patch, I was able to generate the output. I reduced the pause to speed things up:
```
diff --git a/test/private/execute_save_stage.m b/test/private/execute_save_stage.m
index f92a236..ced2a1b 100644
--- a/test/private/execute_save_stage.m
+++ b/test/private/execute_save_stage.m
@@ -18,8 +18,9 @@ function [status] = execute_save_stage(status, ipp)
 
             case 'Octave'
                 % In Octave, figures are properly cropped when using  print().
+                % FIXME: 'figureVisible' need to be 'true'?!
                 print(reference_pdf, '-dpdf', '-S415,311', '-r150');
-                pause(1.0)
+                pause(0.2)
             otherwise
                 error('matlab2tikz:UnknownEnvironment', ...
                      'Unknown environment. Need MATLAB(R) or GNU Octave.')
diff --git a/test/private/testMatlab2tikz.m b/test/private/testMatlab2tikz.m
index dbfa217..31746f2 100644
--- a/test/private/testMatlab2tikz.m
+++ b/test/private/testMatlab2tikz.m
@@ -61,7 +61,7 @@ function status = testMatlab2tikz(varargin)
 
   ipp = ipp.addOptional(ipp, 'testFunctionIndices', [], @isfloat);
   ipp = ipp.addParamValue(ipp, 'extraOptions', {}, @iscell);
-  ipp = ipp.addParamValue(ipp, 'figureVisible', false, @islogical);
+  ipp = ipp.addParamValue(ipp, 'figureVisible', true, @islogical);
   ipp = ipp.addParamValue(ipp, 'actionsToExecute', @(varargin) varargin{1}, @isFunction);
   ipp = ipp.addParamValue(ipp, 'testsuite', @ACID, @isFunction );
```

## Tests whose output looks very wrong:
53, 55, 77
## Tests that give errors:
74,81,83
Unfortunately those tests need to be manually commented out in `acid.tex`, otherwise compilation fails. :cry: 

I will output the `acid.pdf` shortly.

[...]

Issue #326, opened 2014-03-05
ACID: file formats


egeerardyn:
For the ACID test, we might want to consider different file formats for the reference (MATLAB) images.

In #321 @nschloe  suggested to use PNG instead of EPS/PDF. While I'm not in favor for that option, we might want to add the option to do so as PNG files are more compatible with Github. It is only a bit of work to parametrize this: adding an option to `matlab2tikz_acidtest` to define the file type: `PNG` versus `EPS/PDF`.

Moreover, we might also want to compile the TikZ figures as standalone. That way we get rid of quite some loose couplings (e.g. the required LaTeX preamble in the test might diverge from the m2t preamble). Moreover, this would allow for easy parallelization of the compilation progress in `make`. A pitfall that I expect to pop up is relative paths (see #291).

okomarov:
I do find problems in printing to pdf with Octave on Windows. I think it is a font related problem. Also, the I never managed to print ACID(3) which creates a growing .pdf (I aborted after it reached ~100Mb). Thus, `.png` might be an alternative.

PeterPablo:
@okomarov, `ACID(3)` runs fine here with octave on a mac. Do you still face this issue? The single tests are output as standalone figures by now.

I don't see a strong advantage in switching to `png`, though I would be fine with changing to that (or another format). Preferably, I would not introduce (yet) another parameter to define the extension of the reference output.

okomarov:
@PeterPablo I have not investigated further. The Octave's `print()` function does not work on Win. I just do not run the whole ACID on Octave.

PeterPablo:
Just started up my virtual machine and gave [this Octave](http://mxeoctave.osuv.de/) a go (version 3.8.2-5). Indeed `print()` does not create output. I played around a little and this seems to be due to the figure not being visible by default.
@okomarov, please try calling the testsuite with additional paramters `'figureVisible',true` and report back.
E.g. `testGraphical(3,'figureVisible',true)` with current PR #485 works fine and gives the expected output.

PeterPablo:
From my understanding `'figureVisible', false` sets `graphics_toolkit gnuplot`. This correctly works around the `print()` bug of invisible figures but triggers a different bug that yields infinitely increasing file output.

```
graphics_toolkit fltk
testMatlab2Tikz('figureVisible',true)
```
works fine though, since `fltk` is ensured that works correct, as long as figures are visible.

[...]

PR #662, opened 2015-05-13
fix testGraphical()

### LINKED ISSUE #664: ACID(21) zoom: warning: division by zero
Executing `ACID(21)` in octave yields many `warning: division by zero` and might crash octave (and thus `testGraphical()`).
Tested on 37d139129f59ba73273ad3826b0a7e59856326db.

PeterPablo:
For `testGraphical()` and octave:
* also output to eps; `make clean` would otherwise delete `reference` data
* remove unecessary `pause`
* always use `gnuplot` backend

Discussion in #649.

Remaining issue, but was present before: `ACID(21)`  might crash octave (even after fixed #664).

### DIFF
--- a/test/private/execute_save_stage.m
+++ b/test/private/execute_save_stage.m
@@ -18,8 +18,8 @@ function [status] = execute_save_stage(status, ipp)
 
             case 'Octave'
                 % In Octave, figures are properly cropped when using  print().
-                print(reference_pdf, '-dpdf', '-S415,311', '-r150');
-                pause(1.0)
+                print(reference_eps, '-depsc');
+
             otherwise
                 error('matlab2tikz:UnknownEnvironment', ...
                      'Unknown environment. Need MATLAB(R) or GNU Octave.')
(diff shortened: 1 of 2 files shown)


PeterPablo:
I updated the first comment slightly to reflect the current situation on `ACID(21)`.

PeterPablo:
This can be reviewed and does not involve hashes.

PeterPablo:
hm, it's unfortunate that you don't receive proper output using EPS. For me it's the other way round. Currently I don't get any output at all.
For now I will cherry-pick my two commits in order to test the graphical output of octave.

egeerardyn:
I agree this is very unfortunate. Perhaps we should parametrize that part, such that we can define what picture format to use: EPS, PDF, PNG, ... and then both of us can work happily by changing a switch :-)

PeterPablo:
Sounds like a reasonable workaround. :-)

egeerardyn:
I'll close this here, since this is really related to an old topic, i.e. #326. I'll add it to the 2.0 milestone such that we keep an eye on this.

PR #663, opened 2015-05-13
testsuite: change from lualatex to pdflatex

### LINKED ISSUE #661: ACID(66) won't compile with lualatex, but works with pdflatex
```
%
\usepackage[T1]{fontenc}
\usepackage[utf8]{inputenc}
\usepackage{pgfplots}
\usepackage{grffile}
\pgfplotsset{compat=newest}
\usetikzlibrary{plotmarks}
\usepackage{amsmath}

\begin{document}
\begin{tikzpicture}

\begin{axis}[%
width=4.525in,
height=3.569in,
at={(0.759in,0.482in)},
scale only axis,
xmin=0,
xmax=6,
ymin=-1,
ymax=1,
]
\addplot [color=blue,solid,forget plot]
  table[row sep=crcr]{%
0	0\\
1	0.841470984807897\\
2	0.909297426825682\\
3	0.141120008059867\\
4	-0.756802495307928\\
5	-0.958924274663138\\
6	-0.279415498198926\\
};
\node[left, align=right, text=black]
at (axis cs:5.75,0.598) {Each cell is a quoted string\\You can specify how the string is aligned\\$\text{You can use LaTeX symbols like }\pi\text{ }\chi\text{ }\Xi$\\$\text{\bf{}Or use bold \rm}\text{\it{}or italic font\rm}$\\and use umlauts like äöüßÄÖÜ and accents éèêŐőŰűç};
\end{axis}
\end{tikzpicture}%
\end{document}
```

lualatex: `! Package inputenc Error: Unicode char \u8:äöü not set up for use with LaTeX.`
pdflatex compiles fine

This is the last issue I see with compiling the `testGraphical('figureVisible',true)`, followed by `make -j4` on my Linux machine on tip of `develop` with all currently open PR merged locally.

PeterPablo:
fixes #661 

Tested on tip of `develop` with all other currently pending PRs merged locally. Compiles the output of `testGraphical()` fine, using `make -j4`. Tested with octave.

### DIFF
--- a/test/data/converted/Makefile
+++ b/test/data/converted/Makefile
@@ -1,7 +1,7 @@
 # ./Makefile
 
 ECHOCMD:=/bin/echo -e
-LATEX:=lualatex --shell-escape -interaction=batchmode
+LATEX:=pdflatex --shell-escape -interaction=batchmode
 
 TEST_SRCS:=$(wildcard test*-converted.tex)
 TEST_PDFS:=$(TEST_SRCS:.tex=.pdf)
--- a/test/tex/Makefile
+++ b/test/tex/Makefile
@@ -1,7 +1,7 @@
 # ./Makefile
 
 ECHOCMD:=/bin/echo -e
-LATEX:=lualatex --shell-escape
+LATEX:=pdflatex --shell-escape
 TARGET:=acid
 
 main:


egeerardyn:
I have the impression this will depend heavily on the simplification algorithm doing it's job, so I think we should hold off such changes.

In the meanwhile, I don't know whether you are aware that you can change such settings locally on your install by doing the following from the command line (assuming bash or compatible shells):
```bash
export LATEX=pdflatex --shell-escape
make
```

From within MATLAB, you can get similar results using `setenv`.

PeterPablo:
I did not know that I can influence `make` like this. Thanks for the hint! On my linux dev machine I can again compile with lualatex, since Christian Feuersänger recently fixed a bug in the lua detection that I reported.

We might reach more developers with `pdflatex` and it might also be desireable to have the output being compileable with `pdflatex` from an end-users perspectice, since that is more common. I don't have hard feelings, if we would drop this though. 

Please note, that `lualatex` struggles with umlauts, as we currently set up the output. I have this documented in issue #661.

PeterPablo:
To allow error free compilation with all common compilers we could use the `iftex` package, as suggested by @egeerardyn.

PeterPablo:
Since the new `cleanfigure` algorithm is in place and it doesn't seem to cause major trouble, we can reconsider the course of action here.
**stick with lualatex**
* easy, since no change required
* works for larger `tex` files, without running into memory issues

**change to pdflatex**
* allows compilation on machines without lualatex and outdated lua backend (pgfplots 1.12)

I guess I am the minority of not having an up-to-date lua backend. So shall we close here and keep it the way it currently is?

PeterPablo:
Oh, your instruction `export LATEX="pdflatex --shell-escape"` followed by `make` does not work for me. Compilation still uses `lualatex`.

egeerardyn:
I am retesting this on top of the current develop. It passes in R2014b, I'm currently testing R2014a and will have a go at Octave (not that I really trust that). If it compiles in MATLAB at least, I will merge this.
For Octave, we can do the reverse workaround given above if would prove to be a problem.

PeterPablo:
Thanks! This makes it easier for me on my Xubuntu machine.

PR #669, opened 2015-05-18
Guess plot type

PeterPablo:
**octave only**
Guess octave plot type, since function `handle()` is not available.

* [x] makes scatter plots work in octave `ACID(53)` and `ACID(55)`
* [x] allows errorbar plots
* [x] errorbar plot `ACID(74)`: legend is not handled in native implementation; comment is added -- the issue probably is close to [this line](https://github.com/matlab2tikz/matlab2tikz/blob/octave_guessType/src/matlab2tikz.m#L723); #760 should solve the issue
* [x] `ACID(83)`: tex output is not compileable -- `cData` was empty

image


PeterPablo:
Does the assignment imply something specifically? Here I kept it open to indicate that somebody else can free to take over, as stated in my first comment.
Especially for `ACID(83)` I need help of somebody familiar with the scatter plots. My change seems to have unveiled a bug that I couldn't fix in a reasonable amount of time.

egeerardyn:
Oh, I thought this was the all PRs combined, so I assigned you to mark it that you are curating that part.
Also, I'm not sure about the "do not merge" pseudo-label. For me "work-in-progress" on a PR conveys the same meaning actually.

PeterPablo:
Fine, I will remove that in the title. This was meant as a red lantern, if somebody fails to notice the wip label.

PeterPablo:
@okomarov, you are knowledgeable about octave. As you see as it currently stands we can detect scatter plots and errorbar plots in octave (your hint `udata`). Do you know further properties that identify some plot types? If not, we can always go the manual route, as I did with scatter plots. There I actually looked up, what properties are used by the *to be called* m2t function. (No pressure, since I labelled this for 1.1)

PeterPablo:
I'd like to get some feedback on this. My main concern is, whether this PR should fix the missing legend of the error bar plot and whether further native plot types should tried to be identified?
I fear that if we would want to do this, that it will take quite some time, until we can merge this PR. We could also merge this as a beginning and then work on the other plot types separately.

src/matlab2tikz.m
```diff
@@ -2503,6 +2503,28 @@ case size(cData,2) % specific y-data is given
     end
 end
 % ==============================================================================
+% Function `handle` is not yet implemented in Octave.
+% Consequently the plot type needs to be guessed. See #645.
+% If the type can not be determined reliably, 'unknown' will be set.
+function cl = guessOctavePlotType(h)
+    properties = get(h);
+    % scatter plots
+    if isfield(properties,'marker') && ...
```
egeerardyn: Just a few minor things I think of
 - Doesn't Octave have the `type` property as MATLAB does in the HG1 framework?
 - I think we can make this code clearer by introducing a function `hasProperties(h, propertynames)` that would check the presence of those properties and preferably do so case-insensitively. Then the fragment would become something like:

```matlab
if hasProperties(h, {'Marker', 'SizeData', 'CData'})
    cl = 'specgrap.scattergroup';
elseif ...
    ...
end
```

egeerardyn:
If we don't introduce new regressions (i.e. if the legend was wrong before and it still is wrong), then it doesn't need to be fixed in this PR.  If it was fine before, it should be fixed before the next release: this PR or another one, that doesn't matter too much for me.
The same goes for other native formats: we don't have to tackle all of them in this PR.

If you think the PR is ready for further review, please remove the WiP label.

PeterPablo:
I introduced the helper function you suggested. Unfortunately I didn't see an easy way to make it work case-insensitive. The helper function in it's current form has the downside that it can't deal with negations, e.g. that it can check for `has property X` and  `doesn't have property Y`. It certainly can be extended later on, if it turns out that this is needed though.

Unfortunately the errorbar legend handling is a regression, as we now take a completely different code path. Since I am very unexperienced at how to properly debug octave, I'd prefer if somebody else could take a look (@okomarov may be? :-) ).

Review time :-)

PeterPablo:
One moment...

okomarov:
I've started working on it a few days back but have not progressed. Also, I haven't checked the progress of this PR yet.

However, a couple of things I found:
* Often, you cannot identify the type of hggroup from its properties, but you need to check for its children and their properties. 
* Expect many regressions, since code for hggroups was written to work on Matlab only. 

Also, bear in mind that `isprop()` is supported in Octave too.

PeterPablo:
I don't expect regressions apart from that missing legend, as this commit only allows native handling of scatter plots and errorbar plots. Both show improved behavior. Of course if other plot types are moved to the native approach, then their output needs to be carefully evaluated.

Do you see an advantage in using  `isprop()`? Would this make the check for a property case-insensitive?

egeerardyn:
I don't know about case-sensitivity and `isprop`. The easy way I would implement case-insensitivity is to do something like:
```matlab
fields = lower(fieldnames(get(h)));
all(ismember(fields, lower(fieldsExpectedPresent)));
~any(ismember(fields, lower(fieldsExpectedAbsent)));
```

There is, however, a gnawing feeling in the back of my head that there was a edge case of `ismember` with cells that isn't properly supported by Octave; but I can't remember exactly which.

PeterPablo:
Thanks for your suggestion. I had tried something similar without success. @egeerardyn, please feel free to review.
In my tests this worked :-)

PeterPablo:
After merging current `develop` ACID(83)` finally contains output. Unfortunately it is uncompileable code.

Extract:
```
%
\usepackage[T1]{fontenc}
\usepackage[utf8]{inputenc}
\usepackage{pgfplots}
\usepackage{grffile}
\pgfplotsset{compat=newest}
\usetikzlibrary{plotmarks}
\usetikzlibrary{arrows.meta}
\usepackage{amsmath}

\newlength\figureHeight \setlength{\figureHeight}{6cm}
\newlength\figureWidth \setlength{\figureWidth}{10cm}
\begin{document}
\begin{tikzpicture}

\begin{axis}[%
width=0.951\figureWidth,
height=\figureHeight,
at={(0\figureWidth,0\figureHeight)},
scale only axis,
colormap/jet,
xmin=   0,
xmax=  11,
xtick={   1,   2,   3,   4,   5,   6,   7,   8,   9,  10},
xmajorgrids,
ymin=   0,
ymax= 140,
ymajorgrids,
]
\addplot[scatter,only marks,scatter src=explicit,scatter/use mapped color={mark=x,mark options={},draw=mapped color},visualization depends on={\thisrowno{2} \as \perpointmarksize},scatter/@pre marker code/.append style={/tikz/mark size=\perpointmarksize}] plot table[row sep=crcr,meta index=3]{%
   1	  10	0.790569\\
   2	  10	0.940151\\
   3	  10	1.04045\\
   4	  10	1.11803\\
   5	  10	1.18218\\
   6	  10	1.23731\\
   7	  10	1.28592\\
   8	  10	1.32957\\
   9	  10	1.36931\\
  10	  10	1.40585\\
};
\end{axis}
\end{tikzpicture}%
\end{document}
```

Any idea on how to solve this?

PeterPablo:
HG1:
```
\addplot[scatter,only marks,color=blue,mark=x,mark options={},scatter/use mapped color=blue,visualization depends on={\thisrowno{2} \as \perpointmarksize},scatter/@pre marker code/.append style={/tikz/mark size=\perpointmarksize}] plot table[row sep=crcr]{%
```

octave:
```
\addplot[scatter,only marks,scatter src=explicit,scatter/use mapped color={mark=x,mark options={},draw=mapped color},visualization depends on={\thisrowno{2} \as \perpointmarksize},scatter/@pre marker code/.append style={/tikz/mark size=\perpointmarksize}] plot table[row sep=crcr,meta index=3]{%
```

The definition is mixed up...

egeerardyn:
The `meta index=3` is certainly wrong: at most it could be `meta index=2`; this causes the file to be uncombilable. 
But there is certainly a lot more going on, since the `mapped color` stuff is also wrong and there should a lot more points in the figure.

PeterPablo:
The data is present in the output -- I simply stripped it down so it is easier to see what is going on. It is only the `\addplot` definition that is borked. Those are the issues that @okomarov was warning about -- the functions that are now called haven't been tested with octave before.

egeerardyn:
@PeterPablo Have you been able to look into this further over the past few days?

PeterPablo:
Unfortunately I haven't found time to dig into the code branch that is responsible for `ACID(83)` containing uncompileable code. There some adjustments for octave are necessary. Since this test wouldn't have produced any output at all before the current state certainly is an improvement, though worsens the experience of calling `make` to compile the graphical testsuite.

PeterPablo:
Issues with scatter plots are solved. Only missing thing is literally the legend of errorbar plots.

PeterPablo:
I tried to debug why the legend is not shown for `ACID(74)`. @okomarov, can you help? Probably the issue is close to [this line](https://github.com/matlab2tikz/matlab2tikz/blob/octave_guessType/src/matlab2tikz.m#L723).

@egeerardyn -- how should we proceed? Review the PR as is and file an issue for the missing legend entry? #702 does not help, as it specifically deals with MATLAB.

src/matlab2tikz.m
```diff
@@ -2433,17 +2433,16 @@ function assertRegularAxes(handle)
 end
 % ==============================================================================
 function [m2t, str] = drawHggroup(m2t, h)
-% Octave doesn't have the handle() function, so there's no way to determine
-% the nature of the plot anymore at this point.  Set to 'unknown' to force
-% fallback handling. This produces something for bar plots, for example.
+% Continue according according to the plot type. Since the function `handle` is
```
egeerardyn: `according according` -> `according`

egeerardyn:
If the legend issue was introduced here, we should fix it before the next release. Whether that happens in this PR or not, is of lesser importance (I prefer this PR, but I can live with either: we have done the same for quiver plots).

If a legend issue (in the same circumstances) was present before this PR, then we should address that issue elsewhere.

In any case: the code looks OK. I will run it through some visual inspections to verify.

PeterPablo:
Alright, then we need to fix the legend issue here. I am open to suggestions as to how to fix this. Debugging octave still is not my favorite task...

Quiver can wait, as I did not flip the switch for quiver plots to use the *native* approach.

egeerardyn:
Graphical output for everything but `ACID(74)` (i.e. legend of error bars) looks great. Great job! I will try to have a look at this as well (but I share your aversion of Octave debugging, I think we should one day write a simple overview of equivalent debugging commands for Octave and MATLAB).

PeterPablo:
Good to hear you like it :-)
I think I once temporarily hacked `matlab2tik.m` to return the structure. In case you force m2t to not close the figure window you then have the inners of the conversion -- and especially the handles -- available. In case you take a look, you probably can start of at the code line I linked in the issue description.

egeerardyn:
I think it really makes sense to return our internal data structure for debugging (I once filed a PR for that and an interface to extract function handles as well, but at the time Nico thought it was a bad idea). I will try to dig up that code again and refile that PR (it is necessary to unit-test our code anyways).

For debugging, I typically step through the code using a debugger (quite convenient in MATLAB, Octave not so much since I'm not used to its command line interface).

okomarov:
The Octave GUI works *pretty* much as the Matlab editor when debugging.

egeerardyn:
Using the renewed `figure2dot` tool and the Octave debugger, I was able to get some more information on this.

First, let's have a look at what the figure looks like to Octave:
test74-tree-reduced

I may have pruned some parts, but those shouldn't be relevant.
The bold nodes are the ones present in the legend (i.e. for those `hasLegend` would return `true` near line 723:
```matlab
        if ~isempty(m2t.gcaAssociatedLegend)
            delfun           = get(m2t.gcaAssociatedLegend,'deletefcn');
            legendEntryPeers = delfun{6} % See set(hlegend, "deletefcn", {@deletelegend2, ca, [], [], t1, hplots}); in legend.m  
            hasLegend        = ismember(child, legendEntryPeers);
            interpreter      = get(m2t.gcaAssociatedLegend, 'interpreter');
            legendString     = getOrDefault(child,'displayname','');
        end
```
However, we check the legend only for the blue nodes (maybe not only, but those are the relevant ones at least).

So as with MATLAB, we probably have to specify that not only the node itself, but maybe also its direct children should be checked. I haven't yet tried to implement this properly, so feel free to have a go.

okomarov:
What I remember from time ago is that there is no unified approach to test for a legend due to the differences in implementation between Matlab and Octave. 

Basically, we need a `hasLegend()` method and a `getLegendProperties()` that should work independently of Matlab/Octave. 

Suppose we have a line object:
* In Matlab, you can test **directly** if it has a legend entry through the `Annotation` property, but to retrieve the info, you will need to work **backwards** from the legend object and extract the handle of interest from the handles contained under the `PlotChildren` property.
* In Octave, you can only test and retrieve the properties by working **backwards** from the legend objects. 

Things will start to get complicated when you are trying to include legends for series-plots like stemseries etc... You do not want to detect a legend (that does not exist) for each sub-component of that group, especially when Octave cannot tell apart those series-plots from a regular handleAllChildren call. 

Hence, having a `hasLegend()` method would solve this problem. First, you travel down all the sub-componenets of the composite object and check if the parent has a legend. Then you only retrieve the legend entry properties for the Parent.

egeerardyn:
@okomarov I agree. Those legends are a pain and doing it right is not trivial. That's why I'm currently not even attempting to implement it. Probably one can write 3 lines to check the children/parent as well, but I think it's very likely that by doing so, we get a lot of spurious legend entries as well.

okomarov:
I came across this nifty undocumented feature in Matlab:

    children = graph2dhelper ('get_legendable_children', axisHandle);
children = 
  3x1 graphics array:

  Bar     (LMH)
  Line    ($High$)
  Line    ($Low$)

Not sure yet if it might be of help in testing for `hasLegend` but it does have the legend label info :)

egeerardyn:
@okomarov:

Nifty tool. This makes me wonder what other gems are hidden in `fullfile(matlabroot, 'toolbox','matlab','scribe','private')`.

PeterPablo:
I merged the last changes from `develop` so that this now works on both octave 3.8.2 and octave 4.0.0. As pointed out by @egeerardyn [here](https://github.com/matlab2tikz/matlab2tikz/pull/669#issuecomment-129929978) things look pretty, apart from the missing legend. @okomarov is currently working on this in #760 (thank you!). I updated the first comment to reflect the state of affairs.

Are there other things that need to be polished in this PR?

egeerardyn:
:+1: for the code, effort and great results. I haven't executed the test suite again, but I see no reason for regressions since the last time.

As far as I'm concerned, this can be merged since the troubles with `ACID(74)` can only be fixed by #760 and are also indicated by a FIXME in the code.

src/matlab2tikz.m
```diff
@@ -2475,17 +2475,16 @@ function assertRegularAxes(handle)
...
 % #COMPLEX: big switch-case
-% TODO: Introduce function `guessOctaveType()` to infer plot type by trying
-% to read certain properties, that are specific to a plot type (for example
-% `udata` should be unique to errorbarseries). See #645 for more details.
-    try
+    if strcmpi(getEnvironment, 'MATLAB')
         cl = class(handle(h));
-    catch %#ok
-        cl = 'unknown';
+    elseif strcmpi(getEnvironment, 'octave')
```
egeerardyn: Hmm, only nit-pick, maybe change this to:
```
switch getEnvironment()
    case 'MATLAB'
       ...
    case 'Octave'
       ...
    otherwise
       errorUnknownEnvironment()
end
```
?

PeterPablo:
Thanks! Let me take here of that comment and then I will merge this. As you noted I already put a note inside the code. This way it is also easy for @okomarov to test his PR, as it should fix this issue.

Issue #671, opened 2015-05-26
Native quiverplots


egeerardyn:
Quiver plots should be implemented as native quiver plots in `pgfplots`, i.e. similar to what is mentioned in its manual in Section 4.5.8 "Quiver Plots (Arrows)":
![quiver](https://cloud.githubusercontent.com/assets/177360/7807544/c2ec8748-038c-11e5-87c8-94f3695e8770.png)
Section 4.6.10 "3D quiver plots (arrows)" states that 3D is supported by supplying a `w` option as well.

Our current output (`ACID(22)`) is:
```latex
\begin{tikzpicture}[%
arrow1/.style={->,color=mycolor1,solid}
]

\begin{axis}[%
width=0.951\figureWidth,
height=\figureHeight,
at={(0\figureWidth,0\figureHeight)},
scale only axis,
xmin=   0,
xmax=   1,
ymin=   0,
ymax=0.45,
axis background/.style={fill=white}
]
\addplot [arrow1] coordinates{(   0,   0) (0.45,0.45)};
\addplot [arrow1] coordinates{(   1,   0) (0.55,0.45)};
\end{axis}
\end{tikzpicture}%
```

egeerardyn:
This has been addressed in #683 and #690. There still is a bit of work to be done, i.e. #691, but I think we can close this one already.

PR #679, opened 2015-06-19
Now using pgfplots' native quiver functionality.

jam4375:
This pull request is in response to this [issue](https://github.com/matlab2tikz/matlab2tikz/issues/671). In addition to now using the native quiver functionality, the arrow heads are scaled with the arrow length similar to [here](http://tex.stackexchange.com/a/210219/29042).


src/matlab2tikz.m
```diff
@@ -3692,19 +3694,20 @@ case size(cData,2) % specific y-data is given
...
-    % - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
-    % return the vector field code
-    str = [str, ...
-        sprintf(['\\',name,' [arrow',num2str(m2t.quiverId), '] ', ...
-        'coordinates{(',format,') (',format,')};\n'],...
-        data)];
-    %FIXME: external
+    
+    tabOpts = 'row sep=crcr';
+    [m2t, table] = makeTable(m2t, repmat({''}, size(data,2)), data);
```
egeerardyn: If you call it as 
```matlab
variables = {'x','y','z','u','v','w'}; % 3D
variables = {'x',y','u','v'}; % for 2D
[m2t, table] = makeTable(m2t, variables, data);
```

there is no need to specify `x y u v` in the output code below as the table will include them for you.

src/matlab2tikz.m
```diff
@@ -3692,19 +3694,20 @@ case size(cData,2) % specific y-data is given
...
+    str = sprintf(['\\%s[%%\npoint meta={sqrt((\\thisrow{u})^2+(\\thisrow{v})^2)},\n'...
```
egeerardyn: Most of these things only account for 2D arrows, so we should also account for 3D.

egeerardyn:
Thank you for your contribution!
There are, however, a few things that need to be addressed before we should merge this.
If you need any help to address my comments (or if you prefer that I have a go at it), just give a shout.

In any case, I will take care of the failed build for you, as I presume this is mostly due to the line you've added to the LaTeX preamble.

src/matlab2tikz.m
```diff
@@ -3678,11 +3677,14 @@ case size(cData,2) % specific y-data is given
 
     arrowOpts = opts_new();
     if showArrowHead
-        arrowOpts = opts_add(arrowOpts, '->');
+        arrowOpts = opts_add(arrowOpts, '-Straight Barb');
```
egeerardyn: I think we should warn the user that he should load the TikZ library `arrows.meta`, but I'd have to look up how we do this with other dynamic dependencies to give a snippet of code for this.

jam4375:
Hi Egon,

Thank you for your suggestions! I have made changes to the code to address them. I am not sure how you would like to warn the user though. I have tested the code for both 2D and 3D quiver plots on MATLAB R2011a (I don't have access to a newer version). I look forward to hearing from you.

-Jason Monschke

egeerardyn:
Hi Jason,

Thank you for carrying out the changes. It looks fine: I will look up that warning stuff and add it, along with the changes to keep our test suite happy :+1:

egeerardyn:
This is to be merged in #683.

PR #683, opened 2015-06-24
Native quiver

### LINKED ISSUE #671: Native quiverplots

egeerardyn:
This is a continuation of the work of @jam4375 in #679 to fix #671.

Some more personal TODOs before I consider this finished:
 - [x] use `opts_*` framework to set most of these options
 - [x] move arrow options from figure level to the plot level (this patch makes this more convenient)
 - [x] perform graphical tests in all environments (hashes should pass already)


egeerardyn:
The output looks okay, not great for `ACID(22)`, but that one seems like an unnatural test case anyway. Some TODOs have been added to the test and code to signal this for other devs.

egeerardyn:
@PeterPablo : Could you review this PR, please?

PeterPablo:
Will do now.

PeterPablo:
Okay...
* Hashes match on R2014a/b and octave
* Graphical output compilation struggles on my Linux machine. It might be caused by the old Lua version. Thus I temporarily changed to `pdflatex` in the `Makefile`:
```
There is no LUA colorspace for 'rgb' - skipping LUA color map definition of hot
There is no LUA colorspace for 'rgb' - skipping LUA color map definition of mymap
make[1]: *** [test10-converted.pdf] Error 1
```

R2014b:
image
image
* [ ] Shouldn't this be better now?

R2014a:
*will edit*

egeerardyn:
I noticed `ACID(22)` as well, but since the example is quite contrived, I didn't inspect it any further. But I think that `ACID(24)` is unacceptable, I seem to have missed that one. So we should fix the scaling before merging this PR.

src/matlab2tikz.m
```diff
@@ -3676,35 +3673,56 @@ case size(cData,2) % specific y-data is given
...
+        arrowHeadOpts = opts_new();
+        arrowHeadOpts = opts_add(arrowHeadOpts, 'scale length', ...
+                                 '{max(0.01,\pgfplotspointmetatransformed/1000)}');
+        arrowHeadOpts = opts_add(arrowHeadOpts, 'scale width', ...
+                                 '{0.5*max(0.01,\pgfplotspointmetatransformed/1000)}');
+        headStyle = ['-{Straight Barb[' opts_print(m2t, arrowHeadOpts, ',') ']}'];
+        quiverOpts = opts_add(quiverOpts, 'every arrow/.append style', ...
+                              ['{' headStyle '}']);
+    else
+        %TODO: check what we can do for scaling non-barbed arrows
```
PeterPablo: Is there even anything we *need* to do? If `ShowArrowHead` is `false`, there is no arrow head to scale/style?!

egeerardyn: Indeed, there is nothing we can do. I must have misinterpreted the width option to mean the line width, but it's really the width of the arrow. Will remove that TODO.

src/matlab2tikz.m
```diff
@@ -3676,35 +3673,56 @@ case size(cData,2) % specific y-data is given
...
-        sprintf('arrow%d/.style', m2t.quiverId), ...
-        ['{', arrowOptions, '}']);
-    % - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
-    % return the vector field code
-    str = [str, ...
-        sprintf(['\\',name,' [arrow',num2str(m2t.quiverId), '] ', ...
-        'coordinates{(',format,') (',format,')};\n'],...
-        data)];
-    %FIXME: external
+    if showArrowHead
```
PeterPablo: This is, where the octave codepath diverges. (octave output of `testGraphical([22])` looks good)

PeterPablo:
The octave graphical output is fine as well and code looks good to me. So why not merge this now, so that the other PRs can already be merged with the updated hashes/preamble and fix the remaining MATLAB issues (remove TODO note, fix MATLAB scaling) separately?

src/matlab2tikz.m
```diff
@@ -3676,35 +3673,54 @@ case size(cData,2) % specific y-data is given
...
+        %TODO: scale the arrows more rigorously to match MATLAB behavior
+        %There is a "MaxHeadSize" property (at least in R2014b) that plays a
+        %role in determining the quiver size.
+        arrowHeadOpts = opts_new();
+        arrowHeadOpts = opts_add(arrowHeadOpts, 'scale length', ...
```
PeterPablo: This leads to wrong output of `ACID(22)`. Manually setting `scale width` and `scale length` to 10 results in better output. Consequently there must be a bug in the two formulas. Rereading the pgfmanual concerning `\pgfplotspointmetatransformed` might be a good start (didn't do this myself).

PR #684, opened 2015-06-24
Silently ignore `light` objects

### LINKED ISSUE #680: Light support
It would be nice if the `light` function was supported. Up to now `I don't know how to handle this object: light` error is returned.

egeerardyn:
This fixes #680 by ignoring, since support for lighting is unlikely to be introduced into `pgfplots`


PeterPablo:
:+1:

PR #690, opened 2015-07-07
Fix quiver

### LINKED ISSUE #689: Scaling of arrows in quiver3-plots
When making a quiver3 plot in Matlab and exporting them using matlab2tikz, I noticed that the "scale" option set to 0 did not have any effect. Scale set to 0 should disable automatic scaling. 

From the [documentation](http://se.mathworks.com/help/matlab/ref/quiver3.html?refresh=true):
> `quiver3(...,scale)` automatically scales the vectors to prevent them from overlapping, and then multiplies them by scale. scale = 2 doubles their relative length, and scale = 0.5 halves them. Use scale = 0 to plot the vectors without the automatic scaling.

I guess I can use scale = 1 if my vectors don't overlap

KInd regards
Mathias

egeerardyn:
This should fix #689.

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -1075,7 +1075,7 @@ end
 % ==============================================================================
 function m2t = drawBoxAndLineLocationsOfAxes(m2t, h)
 % draw the box and axis line location of an axes object
-    isBoxOn       = strcmpi(get(h, 'box'), 'on');
+    isBoxOn       = isOn(get(h, 'box'));
     xLoc          = get(h, 'XAxisLocation');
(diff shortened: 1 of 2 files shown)


PeterPablo:
Should a test be added to cover this code path? Changes look good.

egeerardyn:
@PeterPablo You are very right. I will add a TODO such that we can do that after this one and #671 are in `develop`, such that we don't have to meddle too much with the hashes.

PeterPablo:
:+1: Feel free to merge, as soon as you added the TODO note.

Issue #584, opened 2015-03-19
Updater should be more verbose


egeerardyn:
The updater should be more verbose towards the user, more specifically.

 - [x] before upgrading to another major version (see #583 and [Semantic Versioning](http://www.semver.org)) the user should be warned about the practical implications (i.e. some parameters may not work anymore or do different things than before)
 - [x] before an upgrade, show the change log (preferably just the relevant part of it)
 - [x] show a message on how to disable updating (if we don't do so already)
 - [x] :new: check for upgrades after conversion (possibly in an `onCleanup` construct to allow for patches in case an error is introduced)
 - ? :new: perhaps we should add a parameter `installUpdates` that allows to choose whether an update should be carried out. If the update is disallowed, we should just display a message and maybe even the changelog. Possible values would be:
    * `'major'` (or `'yes'`): major (and minor and patch) versions are installed automatically
    * `'minor'`: minor (and patch) versions are installed automatically 
    * `'patch'`: patch  versions are installed automatically
    * `'ask'`: just show the prompt as we do now (default behavior)
    * `'no'`: don't update automatically

egeerardyn:
Let's put the "auto-install" feature in the fridge. I'm not quite sure there is a lot of demand for that: it would be nice to have from our side, but I'm unsure whether we really need it at the moment.

PeterPablo:
I also think that this isn't any of our *core* competences and from the first implementation tests I remember that implementing this feature is more involved than it seems at first. So let's leave things the way they are for now.

PeterPablo:
Implemented by #697

Issue #583, opened 2015-03-18
Implement Semantic Versioning


egeerardyn:
[Semantic Versioning](http://www.semver.org) is an understandable way to designate a version number. Briefly:

> Given a version number MAJOR.MINOR.PATCH, increment the:
>
> - MAJOR version when you make incompatible API changes,
> - MINOR version when you add functionality in a backwards-compatible manner, and
> - PATCH version when you make backwards-compatible bug fixes.
> 
> Additional labels for pre-release and build metadata are available as extensions to the MAJOR.MINOR.PATCH format.

Our 1.0 release is way overdue, so this is an ideal time to adhere to this "standard" in the future. It will mean our version numbers will increase at a higher pace than currently (but that's not a bad thing).

okomarov:
Should the next release be then 1.0.0? 
Is there an easy way to count number of patches, minor and major mods applied when closing issues and merging PR?

PeterPablo:
Fine by me to bump the version. Since API should not change right afterwards, I think that especially #471 (filename/filehandle) should be dealt with and the new simplification algorithm #571 introduced.
@okomarov -- you can see the number of commits between tags and e.g. a branch: [example](https://github.com/matlab2tikz/matlab2tikz/compare/0.6.0...develop)

nschloe:
> Since API should not change right afterwards

I wouldn't mind bumping to 2.* for the next release. Otherwise, we'll never catch up with Chrome.
For the release, I thought I'll definitely wait for #571 and #560. If there's #471 on top, fine.

egeerardyn:
@okomarov 
> Is there an easy way to count number of patches, minor and major mods applied when closing issues and merging PR?

Not that I'm aware of. That's a really good point you bring up. I suggest we have non-firm numbers in the `develop` branch and regularly check for the amount of features that have changed. While our current releases are quite far apart (a few months) and contain a lot of new things, maybe we can just release more minor versions i.e. after each $smallNumberOf PRs, we have a release PR (i.e. merge `develop` back into `master`) with a bumped version number, change log added, etc.

@okomarov @PeterPablo @nschloe 
With regard to #471 and deprecating old features, that is explicitly mentioned in [SemVer](http://semver.org) (emphasis mine):

> #### How should I handle deprecating functionality?
>
[...]

PR #697, opened 2015-07-14
Improve updater

egeerardyn:
This addresses most of #584, i.e. the updater
 - is Semantic Versioning-aware, 
 - will give a basic interpretation of the upgrade size (i.e. major vs minor vs patch),
 - will show the relevant part of the changelog to the user if they it ask for it,
 - will perform the upgrade *after* TikZ conversion instead of before.

As a cherry on top, I made the code a bit more modular.

What is in #583, but isn't implemented here is the automatic options or the options to upgrade certain versions automatically. I don't think those are nice-to-haves, but not crucial.

### DIFF
--- a/src/private/errorUnknownEnvironment.m
+++ b/src/private/errorUnknownEnvironment.m
@@ -1,0 +1,5 @@
+function errorUnknownEnvironment()
+% Throw an error to indicate an unknwon environment (i.e. notMATLAB/Octave/...).
+    error('matlab2tikz:unknownEnvironment',...
+          'Unknown environment "%s". Need MATLAB(R) or Octave.', getEnvironment);
+end
--- a/src/private/errorUnknownEnvironment.m
+++ b/src/private/errorUnknownEnvironment.m
@@ -1,5 +1,5 @@
 function errorUnknownEnvironment()
-% Throw an error to indicate an unknwon environment (i.e. notMATLAB/Octave/...).
+% Throw an error to indicate an unknwon environment (i.e. not MATLAB/Octave).
     error('matlab2tikz:unknownEnvironment',...
           'Unknown environment "%s". Need MATLAB(R) or Octave.', getEnvironment);
 end
--- a/src/private/m2tUpdater.m
+++ b/src/private/m2tUpdater.m
@@ -33,8 +33,8 @@ function m2tUpdater(about, verbose)
 
   mostRecentVersion = determineLatestRelease(version);
   if askToUpgrade(mostRecentVersion, version, verbose)
-    tryToUpgrade(name, fileExchangeUrl, verbose);
-    userInfo(verbose, '');
+      tryToUpgrade(name, fileExchangeUrl, verbose);
+      userInfo(verbose, '');
   end
 end
 % ==============================================================================
@@ -170,23 +170,22 @@ function mostRecentVersion = determineLatestRelease(version)
 end
 % ==============================================================================
 function isBelow = isVersionBelow(versionA, versionB)
-  % Checks if version string or vector versionA is smaller than
-  % version string or vector versionB.
-  env = getEnvironment;
-  vA = versionArray(versionA);
-  vB = versionArray(versionB);
+% Checks if version string or vector versionA is smaller than
+% version string or vector versionB.
+    env = getEnvironment;
+    vA = versionArray(versionA);
+    vB = versionArray(versionB);
 
-  isBelow = false;
-  for i = 1:min(length(vA), length(vB))
-    if vA(i) > vB(i)
-      isBelow = false;
-      break;
-    elseif vA(i) < vB(i)
-      isBelow = true;
-      break
+    isBelow = false;
+    for i = 1:min(length(vA), length(vB))
+        if vA(i) > vB(i)
+            isBelow = false;
+            break;
+        elseif vA(i) < vB(i)
+            isBelow = true;
+            break
+        end
     end
-  end
-
 end
 % =========================================================================
 function arr = versionArray(str)
@@ -203,22 +202,17 @@ function arr = versionArray(str)
   else
     arr = str;
   end
-
 end
 % ==============================================================================
 function userInfo(verbose, message, varargin)
   % Display usage information.
+  if verbose
+      mess = sprintf(message, varargin{:});
 
-  if ~verbose
-      return
+      % Replace '\n' by '\n *** ' and print.
+      mess = strrep( mess, sprintf('\n'), sprintf('\n *** ') );
+      fprintf( ' *** %s\n', mess );
   end
-
-  mess = sprintf(message, varargin{:});
-
-  % Replace '\n' by '\n *** ' and print.
-  mess = strrep( mess, sprintf('\n'), sprintf('\n *** ') );
-  fprintf( ' *** %s\n', mess );
-
 end
 % =========================================================================
 function list = rdirfiles(rootdir)
--- a/src/private/m2tUpdater.m
+++ b/src/private/m2tUpdater.m
@@ -45,7 +45,8 @@ function shouldUpgrade = askToUpgrade(mostRecentVersion, version, verbose)
       userInfo(verbose, 'New version available! (%s)\n', mostRecentVersion);
       userInfo(verbose, '**********************************************\n');
 
-      userInfo(verbose, 'By upgrading you may lose any custom changes.\n');
+      warnAboutUpgradeImplications(version, mostRecentVersion, verbose);
+      askToShowChangelof(version, mostRecentVersion);
       reply = input([' *** Would you like ', name, ' to self-upgrade? y/n [n]:'],'s');
       shouldUpgrade = strcmpi(reply(1),'y');
   end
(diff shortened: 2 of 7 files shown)


PeterPablo:
Do you have an suggestion on how to test this? Your changes look nice, but I'd like to give it a spin to feel more comfortable.

PeterPablo:
hmmm, how often does the updater check for a new version? Privacy concerned user's might dislike the automated polling?! ([some context](https://lists.debian.org/debian-devel/2015/07/msg00226.html))

egeerardyn:
My own approach was to place a breakpoint inside `matlab2tikz` on the conditional right before `m2tUpdater` is called. Then I call the subsequent line by hand using <kbd>F10</kbd>.
Most of what I tested, was manually changing the `version` string to something different and see what the response was.

But I agree to your implicit objection: the updater is hardly tested. We should really unit-test it in the future. One of the things that might make it easier for developers to test it, is to change how difficult it is to call that code. I do have a small idea how to make it a bit more accessible and will try to implement it in this PR as well.

egeerardyn:
> hmmm, how often does the updater check for a new version? Privacy concerned user's might dislike the automated polling?! (some context)

Every time you launch `matlab2tikz`, since we don't keep any global variable to keep track of this. This certainly is the easiest way to do everything. But as you mention: some users may be concerned about privacy. In any case: we don't get any privacy-sensitive data about the user. TMW and GitHub are the only parties that could do something with it. However, GitHub gives us [aggregated results](https://github.com/matlab2tikz/matlab2tikz/graphs/traffic), but it seems that there are very few hits on the releases page. So either that's not tracked or few people are using 1.0.

But privacy is certainly something we can look into further in the future and warn our users about.

PeterPablo:
I will take a look probably on monday.

src/private/errorUnknownEnvironment.m
```diff
@@ -0,0 +1,5 @@
+function errorUnknownEnvironment()
+% Throw an error to indicate an unknwon environment (i.e. notMATLAB/Octave/...).
```
PeterPablo: space before `MATLAB` ;-)

src/private/m2tUpdater.m
```diff
@@ -27,7 +27,115 @@
...
+          currentFolderFiles = [currentFolderFiles; fullfile(targetPath, topZipFolder{1})];
+      end
+
+      cleanupOldFiles(currentFolderFiles, unzippedFilesTarget);
+
+      userInfo(verbose, 'Upgrade has completed successfully.');
+  catch
+      userInfo(verbose, ...
+               ['Upgrade has failed.\n', ...
+                'Please install the latest version manually from %s !'], ...
```
PeterPablo: space before `!` unneeded

src/private/m2tUpdater.m
```diff
@@ -58,180 +171,165 @@
...
+    if ~isempty(idxVersion)
+        changelog = changelog(1:idxVersion-1);
+    else
+        % Just show the whole changelog if we don't find the old version.
+    end
+end
+% ==============================================================================
+function warnAboutUpgradeImplications(currentVersion, latestVersion, verbose)
+% This warns the user about the implications of upgrading as dictated by
+% Semantic Versionning.
```
PeterPablo: semantic versioning.

src/private/m2tUpdater.m
```diff
@@ -27,7 +27,115 @@
...
+  shouldUpgrade = false;
+  if ~isempty(mostRecentVersion)
+      userInfo(verbose, '**********************************************\n');
+      userInfo(verbose, 'New version (%s) available!\n', mostRecentVersion);
+      userInfo(verbose, '**********************************************\n');
+
+      warnAboutUpgradeImplications(version, mostRecentVersion, verbose);
+      askToShowChangelog(version);
+      reply = input(' *** Would you like to upgrade? y/n [n]:','s');
+      shouldUpgrade = strcmpi(reply(1),'y');
```
PeterPablo: Here you need to check for an empty response, otherwise this fails. (User directly presses enter.)

src/private/m2tUpdater.m
```diff
@@ -58,180 +171,165 @@
...
+    currentVersion = versionString(currentVersion);
+
+    % Header is "# YYYY-MM-DD Version major.minor.patch [Manager](email)"
+    % Just match for the part until the version number. Here, we're actually
+    % matching a tiny bit too broad due to the periods in the version number
+    % but the outcome should be the same if we keep the changelog format
+    % identical.
+    pattern = ['\#\s*[\d-]+\s*Version\s*' currentVersion];
+    idxVersion = regexpi(changelog, pattern);
+    if ~isempty(idxVersion)
```
PeterPablo: If `idxVersion` is empty, because no changelog entry was created for a given version then the user is presented with the complete changelog.
Example: User is on 0.7.0 and could update to 1.0.0, I'd say it would make sense to only show everything below 1.0.0 but larger than 0.6.0, which in this case is the next entry.
Please decide yourself, whether you want to change the current behavior. I am fine with both.

src/private/m2tUpdater.m
```diff
@@ -27,7 +27,115 @@
...
+      userInfo(verbose, '**********************************************\n');
+      userInfo(verbose, 'New version (%s) available!\n', mostRecentVersion);
+      userInfo(verbose, '**********************************************\n');
+
+      warnAboutUpgradeImplications(version, mostRecentVersion, verbose);
+      askToShowChangelog(version);
+      reply = input(' *** Would you like to upgrade? y/n [n]:','s');
+      shouldUpgrade = strcmpi(reply(1),'y');
+      if ~shouldUpgrade
+        userInfo(verbose, ['To disable the self-updater in the future, add ' ...
```
PeterPablo: Please consider adding a line break to the output. I am uncertain, whether it is better with or without linebreak.

src/private/m2tUpdater.m
```diff
@@ -27,7 +27,115 @@
...
+end
+% ==============================================================================
+function cleanupOldFiles(currentFolderFiles, unzippedFilesTarget)
+% Delete files that were there in the old folder, but that are no longer
+% present in the new release.
+    newFolderStructure = [getFolders(unzippedFilesTarget);  unzippedFilesTarget];
+    deleteFolderFiles  = setdiff(currentFolderFiles, newFolderStructure);
+    for ii = 1:numel(deleteFolderFiles)
+        x = deleteFolderFiles{ii};
+        if exist(x, 'file')
```
PeterPablo: The folders in the unzipped directory were not removed. I am unsure why...
I received:
```
Warning:
'.../matlab2tikz/matlab2tikz-matlab2tikz-33667a0'
is a directory.  Use rmdir to delete directories. 
```

src/private/m2tUpdater.m
egeerardyn: I know it's unneeded, it's just a personal deviation from that rule since there is an URL before: that makes it easier to copy-paste.

PeterPablo: Sounds reasonable. Ok!

src/private/m2tUpdater.m
egeerardyn: I agree. When the user is on a version without a proper changelog entry, the whole changelog is shown. But in general, that situation should not happen (outside of manual testing).

I slightly prefer the current approach, where we just show the whole thing. Most importantly because it is the least complicated way (hence, less chance for bugs) and because the feature is just a convenience, not an actual must-have feature.

PeterPablo: :+1: This is my opinion as well. Wanted to double check, whether you intended it to work like this.

src/private/m2tUpdater.m
```diff
@@ -58,180 +171,165 @@
...
+    for ii = 1:numel(description)
+        if latestVersion(ii) > currentVersion(ii)
+            cls = description{ii};
+            return
+        end
+    end
+    cls = 'unknown';
+end
+% ==============================================================================
+function isBelow = isVersionBelow(versionA, versionB)
```
PeterPablo: Not needed -- you created a private function?!

src/private/m2tUpdater.m
```diff
@@ -58,180 +171,165 @@
...
+% Converts a version array to string
+  if ischar(arr)
+      str = arr;
+  elseif isnumeric(arr)
+      str = sprintf('%d.', arr);
+      str = str(1:end-1); % remove final period
+  end
+end
+% ==============================================================================
+function arr = versionArray(str)
```
PeterPablo: Not needed -- you created a private function?!

PeterPablo:
Please take a look at my comments. The update is functional in R2014b. I did not check octave, but don't see a reason why this should not work. :+1:

src/private/m2tUpdater.m
egeerardyn: Nice find!

src/private/m2tUpdater.m
egeerardyn: I didn't really touch those mechanics here, I just reorganized them. Anyway, what I think is a bit weird is the fact that your path starts with `...`: often that is not interpreted as a valid path. Let's investigate this in a different issue.

src/private/m2tUpdater.m
egeerardyn: Very true!

src/private/m2tUpdater.m
egeerardyn: Very true!

src/private/m2tUpdater.m
PeterPablo: Sorry for not being more precise. My path is actualy something like `/Users/myaccount/blabla/blub/matlab2tik/....` on Mac. You decide whether changing here or tackling this separately is the way to go.

PeterPablo:
:+1: Feel free to merge -- if you do, please open an issue concerning the not working deletion of the zip folder. Did you test the code on octave?

PeterPablo:
Updater does not work in Octave (on Xubuntu):
```
 *** Downloading and unzipping to '/home/longPath/matlab/0_addons/figures/matlab2tikz/src/private/../..' ...
 *** Upgrade has failed.
```

I checked out d62655d and manually changed the `version` to `0.6.0`. Afterwards I execute:
`ACID(2); matlab2tikz('test.tex','standalone',true,'checkForUpdates',true)`

Updating from MATLAB still works, yet gives the warning and the zip folder is not removed:
```
Warning: '/home/longPath/matlab/0_addons/figures/matlab2tikz/matlab2tikz-matlab2tikz-33667a0' is a
directory.  Use rmdir to delete directories. 
> In src/private/m2tUpdater>cleanupOldFiles at 131
  In src/private/m2tUpdater>tryToUpgrade at 112
  In src/private/m2tUpdater at 35
  In matlab2tikz at 342 
```

@egeerardyn, could you please double check?

egeerardyn:
The MATLAB bug is fixed, Octave on the other hand is not.
Since I did not change that part, I will open up a bug report for that with all my findings up to now.

egeerardyn:
@PeterPablo Can you verify that the bug in MATLAB is resolved?

PeterPablo:
It's not resolved on 29ec4dd. Let me have a look. It is fine that you open a bug report for octave and we merge this, once MATLAB is fixed.

PeterPablo:
```matlab
        if exist(x, 'dir') == 7
            % First check for directories since
            % `exist(x, 'file')` also checks for directories!
            rmdir(x,'s');
        elseif exist(x, 'file') == 2
            delete(x);
        end
```
works. The `kind` setting doesn't work reliably for directories it seems.

But there is another warning `File not found or permission denied.` Let me check and report back as well.

PeterPablo:
```matlab
function cleanupOldFiles(currentFolderFiles, unzippedFilesTarget)
% Delete files that were there in the old folder, but that are no longer
% present in the new release.
    newFolderStructure = [getFolders(unzippedFilesTarget);  unzippedFilesTarget];
    deleteFolderFiles  = setdiff(currentFolderFiles, newFolderStructure);
    for ii = 1:numel(deleteFolderFiles)
        x = deleteFolderFiles{ii};
        if exist(x, 'dir') == 7
            % First check for directories since
            % `exist(x, 'file')` also checks for directories!
            rmdir(x,'s');
        end
    end
    for ii = 1:numel(deleteFolderFiles)
        x = deleteFolderFiles{ii};
        if exist(x, 'file') == 2
            delete(x);
        end
    end
end
```
Will first delete all directories (which is fast) and afterwards delete the remaining filesThis reduces the warnings down to a single on (before numerous). I have the impression that on my NFS file system it takes a short while until it notices that a file is really deleted.

If you like this approach, you might want to beautify it a bit.

PeterPablo:
The last warning is caused by temporary file `/home/longPath/matlab/0_addons/figures/matlab2tikz/test/data/converted/.nfs0000000003025314000007c1`

PeterPablo:
hmpfff -- the updater deleted some of my private test files I had stored in the `m2t` directory. That is very unkind!

okomarov:
@PeterPablo There is a trade-off, either you leave old `m2t` files in, or you remove everything that doesn't belong to the new `m2t`. To avoid deleting private files, you need to:
* carry the info about the whole file structure of the `m2t` release or...
* avoid cleaning the test folder for example

PeterPablo:
Yeah, I totally understand. It's sufficient to introduce a warning that is more explicit than ` *** Would you like to upgrade? y/n [n]:y`
No problem -- I can restore my files (I think)

egeerardyn:
We have such a warning: `By upgrading you may lose any custom changes.`. But we could make it a bit more visible, since by showing the changelog and everything, we are now quite verbose.

Another option (which removes quite some code from this PR as well: i.e. a good thing), would be to just show a link to the changelog on GitHub (and make it clickable in MATLAB). It's less directly accessible (i.e. few people will use it), but it may mitigate this downfall. What do you guys think?

PeterPablo:
I think it is nice to show the changelog. We are all interested people and this way might attract more developers?!
Making the warning more visible should be sufficient.

egeerardyn:
@PeterPablo With regard to the matlab fix (i.e. 29ec4dd), are you sure that doesn't work because the code you provided is equivalent (anything non-zero is regarded as `true` and so `if exist(..., kind) == kindInt` should be equivalent to `if exist(..., kind)`.

Can you verify, because I would like to avoid putting those magic numbers in our code: I hate them with a passion. In any case, I will make the warning stand out more.

PeterPablo:
Please join me in the chat. Yes, I did check it (I can recheck, if necessary). Did you notice this is already done exactly like I suggested somewhere else in that function?

PR #699, opened 2015-07-15
Warn about faceted interp shader

### LINKED ISSUE #693: Hist3 under Matlab2013b: Problems with v.1.0.0
Hi all,

I think there is an issue in v.1.0.0 with how hist3 plots from MATLAB are handled, at least under 2013b. The issue results in huge PDF files (30 Mb per image) being created from the .tikz (using pdflatex).

By comparing the output with my previous saved .tikz files using v.0.6.0 (and 2012b which might be an irrelevant coincidence) I found out that the relevant difference is very simple:

Previously:
```
\addplot3[%
surf,
faceted color=black,
shader=faceted,
colormap={mymap}{[1pt] rgb(0pt)=....},
mesh/rows=75]
table[row sep=crcr,header=false] {%
%
2.18965243889418	1.04390933333333	0\\
< DATA>
```

New:
```
\addplot[%
surf,
shader=faceted interp,faceted color=black,colormap={mymap}{[1pt] rgb(0pt)=....},
mesh/rows=75]
table[row sep=crcr, point meta=\thisrow{c}] {%
%
x	y	c\\
2.18965243889418	1.04390933333333	0\\
<DATA>
```

Additionally, the view={0}{90}, command is missing in the new output, probably because this is not relevant for addplot() anyay in contrast to addplot3

If I change the above differences back, the file is working fine again (size of the created PDF is less than 100k).

UPDATE: Minimum example of the MATLAB code.
```
% Generate data
n_points  = 1E3;
DATA      = repmat([2, 5],n_points,1).*randn(n_points,2);
n_grid    = 15;

colormap(jet)
hist3([DATA(:,1), DATA(:,2)], [n_grid, n_grid]);
set(get(gca,'child'),'FaceColor','interp','CDataMode','auto');
view(2)
```

And yes, you are correct, changing the 'FaceColor' property from 'interp' to 'flat' solves the issue. So basically it is not a bug but a feature in 1.0.0 which was not present in 0.6.0 :)

In case  you know better way on how to create coloured 2D histograms than the use of hist3() that would be most appreciated :)

egeerardyn:
This fixes #693 by just issuing a warning and referring back to the issue where a solution is provided.

I think it is more convenient to refer to the issue than to provide a full explanation, since there we are more at liberty to write a detailed explanation online than in the code.

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -3088,6 +3088,7 @@ function [m2t, opts, s] = shaderOptsSurfPatchEdgeFlat(m2t, handle, opts, s)
     if strcmpi(s.faceColor, 'flat')
         opts = opts_add(opts,'shader','flat corner');
     elseif strcmpi(s.faceColor, 'interp')
+        warnFacetedInterp(m2t);
         opts = opts_add(opts,'shader','faceted interp');
     else
         s.hasOneFaceColor = true;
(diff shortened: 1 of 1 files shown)


PeterPablo:
Code looks good to me. Did not runtime test. Feel free to merge.

PR #701, opened 2015-07-17
Fix #682: patch with no edge color

### LINKED ISSUE #682: Edge of a shaded area is drawn, even when set to 'none'
Dear all,

I am experiencing a problem relating to the edge of shaded areas in my plot. In matlab, I produce a plot like this:

shadedmatlab

This is the code I used to produce the figure:

```matlab
year=1982:2008;

Peak    = [1990.50 2001.00];
Trough  = [1991.00 2001.75];

rng(123)
x0=randn*5;
x=zeros(length(year),1);
x(1)=x0;
for ii=2:length(year)
    x(ii)=0.8+x(ii-1)+randn;
    
end
figure
plot(year,x,'k')
hold on
ylims    = get(gca,'ylim');
for i=1:size(Peak,2);
    patch([Peak(i) Trough(i) Trough(i) Peak(i)],[ylims(1) ylims(1) ylims(2) ylims(2)],[0 0 0 0],[0.7 0.7 0.7],'edgecolor','none');
end
plot(year,x,'k')
xlim([year(1) year(end)])
ylim(ylims)
title('Output','fontweight','normal')
hold off

cleanfigure;
matlab2tikz('testshaded.tex', 'standalone',true); 
```

However, when I compile the .tex file, I arrive at the following plot
shadedtex

Although, I specified the 'edgecolor' to 'none' the edge is there again. If I put 'edgecolor' to 0.7 0.7 0.7 (same as the facecolor), then the edge disappears, but unfortunately, there are gaps in the box of the figure.

How could I solve this issue?

Thanks for your help in advance!

egeerardyn:
This fixes #682 , but I'm currently running the test suite to check regressions.

**edit:** Test suite passes in all environments on Mac.

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -1854,7 +1854,7 @@ function [m2t, str] = drawPatch(m2t, handle)
         cycle  = conditionallyCyclePath(Vertices);
 
         [m2t, drawOptions] = setColor(m2t, handle, drawOptions, 'draw', ...
-                                         s.edgeColor);
+                                         s.edgeColor, 'none');
         [m2t, drawOptions] = setColor(m2t, handle, drawOptions, 'fill', ...
                                          s.faceColor);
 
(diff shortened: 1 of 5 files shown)


PeterPablo:
Should we adapt a test to cover that code path in our testsuite? I'll test this soon.

egeerardyn:
Probably we should, yes. If we can adapt a test case to cover this branch
as well, that would be the best.

PeterPablo:
Ok, I'll test once you updated a test and removed the wip label.

PR #702, opened 2015-07-19
Fixed #700, legends of bodeplots not showing

### LINKED ISSUE #700: Legend of bode plot
As indicated by @tomlankhorst in #667:

> This seems not to work in my case, legend entries still missing:
>```matlab
> bode(tf([1],[1 1]));
> legend('G0');
> cleanfigure
> matlab2tikz('legendbode.tex', 'standalone', true)
> !pdflatex legendbode.tex
>```
> Results in PDF without the text 'G0' in the legend.

Personally, I can confirm the behavior on R2014b: bode plot legends are not OK.

tomlankhorst:
From a first analysis: MATLAB bodeplots consist of groups of two lines (Magnitude, Phase). The lines have no 'DisplayName' but their common parent has one

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -751,7 +751,7 @@ function [legendString, interpreter, hasLegend] = findLegendInfoMATLAB(m2t, chil
     legendString = '';
     interpreter  = '';
     hasLegend = false;
-    isGroup   = false;
+    legendRefersToParent = false;
     %FIXME: this part (e.g. fall back objects) should be restructured.
     for legendHandle = m2t.legendHandles(:)'
         ud = get(legendHandle, 'UserData');
@@ -768,9 +768,7 @@ function [legendString, interpreter, hasLegend] = findLegendInfoMATLAB(m2t, chil
             % legends refer to the specgraph.errorbarseries
             % handle which is 'Parent' to the line handle.
             k = find(get(child,'Parent') == plotChildren);
-            if(~isempty(k))
-                isGroup = true;
-            end
+            legendRefersToParent = ~isempty(k);
         end
         if ~isempty(k)
             % Legend entry found. Add it to the plot.
@@ -779,11 +777,11 @@ function [legendString, interpreter, hasLegend] = findLegendInfoMATLAB(m2t, chil
             if ~isempty(ud) && isfield(ud, 'lstrings')
                 legendString = ud.lstrings{k};
             else
-                legendString = get(child, 'DisplayName');
-                % If there is no DisplayName it might be in the parent
-                if isempty(legendString) && isGroup
-                    parent       = get(child, 'Parent');
+                if legendRefersToParent
+                    parent       = get(child,'Parent');
                     legendString = get(parent, 'DisplayName');
+                else
+                    legendString = get(child, 'DisplayName');
                 end
             end
         end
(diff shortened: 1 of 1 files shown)


src/matlab2tikz.m
```diff
@@ -767,6 +768,9 @@ function addCustomCode(fid, before, code, after)
             % legends refer to the specgraph.errorbarseries
             % handle which is 'Parent' to the line handle.
             k = find(get(child,'Parent') == plotChildren);
+            if(~isempty(k))
```
egeerardyn: I think we should put a comment here to clarify what a "group" is in this context. Some variation on 

> MATLAB bodeplots consist of groups of two lines (Magnitude, Phase). The lines have no 'DisplayName' but the containing Group has one.

would be fine.

Also just as a tiny remark: this statement can be written more concisely:
```matlab
isGroup = ~isempty(k);
```

tomlankhorst:
@egeerardyn Thanks for reviewing. I somewhat reformulated the fix: 
If the 'legendHandle' refers to the parent of a child in 'findLegendInfoMATLAB', the same parents 'DisplayName' should be used instead of the childs one. Seems to work for normal plots, bode plots, errorbar plots. Would like to test somewhat more extensively, should I run 'runMatlab2TikzTest' and manually inspect all results?

tomlankhorst:
Test 20 shows only a single legend entry instead of two as in the reference image. This however seems not to be due to the proposed fix. I'll try to put some effort in fixing these kind of things in the near future.

egeerardyn:
@tomlankhorst : if you can, you can get by by only running `testHeadless` (but `testGraphical` never hurts). A bit more information on the workflow is in `test/README.md` of the repository.

Anyways, I ran `testHeadless` here without any regressions. So as far as I'm concerned, we can merge this unless you want to test this further?

tomlankhorst:
@egeerardyn I am going to test it somewhat further and I'll post the result here.

PeterPablo:
I tested this on R2014a and R2014b and I can confirm that the legend is now present in both cases. Did you notice though that the title is also missing in R2014b (HG2)? Chances are good that it is the same root cause. If you are interested feel free to fix this as well :smiley: 
You can easily verify this by running `testHeadless(42)` and compiling the generated tex file. I find it easy to compare it to the Matlab figure directly, which is generated with `ACID(42)`.

(The change affects the hash of our testcase `bodePlots`, which is good. This is not noticed easily though since, we still mark this test as unreliable.)

PeterPablo:
@tomlankhorst, do you have an update for us? :-)

tomlankhorst:
Hey guys. I am going to sort this out and fix it for sure but
coincidentally I had already planned a short holiday to lovely Greece so
things aren't going as swift as they do normally. I'll catch up on this as
soon as I come back home!
Op 27 jul. 2015 10:52 schreef "Peter Pablo" <notifications@github.com>:

tomlankhorst:
Okay this seems to work fine. 
Regarding the problem with the title of the bode plot. The title is stored in a third axis that has the property 'Visibility' set to off. Therefore it is not added to the list of 'relevantPlots' so that the title isn't added to the output tex. Will create another issue for this. Lastly, looking at test 20 (moreLegends) there is something wrong with the legend in a 'plotyy'. This however seems unrelated to this problem. Will create another issue.

egeerardyn:
Since you're happy with the result and I'm happy with the result, I shall merge this into `develop`.

PR #705, opened 2015-07-24
SublimeText workspace

egeerardyn:
This PR adds a project file for [Sublime Text (ST3)](http://www.sublimetext.com/3).
It's not very high on our priority list, but since the testing of our updater, I had noticed my file went missing, so I had an incentive to rebuild my file properly.

I took that opportunity to write some build systems (see also e9adcf8) that allow to:
 - run the CI test suite using Octave,
 - run the CI test suite using MATLAB (default / HG1 / HG2 versions),
 - make the (graphical) ACID test,
 - distclean the ACID test.

While this does not yet allow to do everything m2t-related from within ST3, I think those are the most common actions. Anyway, for more interactive work with MATLAB/Octave, I think [SublimeREPL](https://sublimerepl.readthedocs.org/en/latest/) is needed anyway.

I expect those build systems to work on Linux as well (given the instructions in e9adcf8), but I would be very surprised if Windows would work. However, I certainly think we could adapt everything to work on Windows as well, if there is some demand.

Since I had to change the `.gitignore` files anyway to ignore the workspace file, I also cleaned those files up a bit.

Anyway: I'm eager to hear your suggestions and comments.


.gitignore
```diff
@@ -1,7 +1 @@
-test*-reference.pdf
-test*-reference.eps
-test*-converted*
-# exclude files generated by test suite
-!.gitignore
-# git cannot presently store empty directories, so committing the
-# .gitignore file makes sure the directory isn't actually empty
+*.sublime-workspace
```
PeterPablo: How about adding `octave-workspace` as well?

PeterPablo:
Nice in general -- thank you! Having the different build systems at hand is great!

Unfortunately git notices my `matlab2tikz.sublime-project` to be different, since the `FuzzyFilePath` addon adds a section and the file is reordered. Consequently `git` asks to commit those changes, which I of course do not want...
Adding `matlab2tikz.sublime-project` to `.gitignore` does not help it seems. Any ideas on how we can resolve this unpleasant effect?

The octave build systems throws an error during setting up the global state. Let me check and edit.

Feel free to merge, as soon as you considered my line comment.

egeerardyn:
@PeterPablo As far as I'm concerned, we can have the settings of `FuzzyFilePath` in there. Do you have any specific settings for those fields or not? Normally, ST2/3 don't care about extra settings that it doesn't recognize.
Only downside is that the autoformatting of that JSON file also removes comments. I'll try to see if it works any better if it already has the right settings in there.

PeterPablo:
Here is the diff:
```diff
diff --git a/matlab2tikz.sublime-project b/matlab2tikz.sublime-project
index f4042aa..0c7656f 100644
--- a/matlab2tikz.sublime-project
+++ b/matlab2tikz.sublime-project
@@ -1,48 +1,95 @@
 {
+	"build_systems":
+	[
+		{
+			"cmd":
+			[
+				"octave",
+				"-q",
+				"--eval",
+				"runMatlab2TikzTests.m"
+			],
+			"env":
+			{
+				"CONTINUOUS_INTEGRATION": "true"
+			},
+			"name": "CI tests (Octave)",
+			"working_dir": "${project_path}/test"
+		},
+		{
+			"cmd":
+			[
+				"matlab",
+				"-nosplash",
+				"-r",
+				"cd('${project_path}/test'); runMatlab2TikzTests()"
+			],
+			"env":
+			{
+				"CONTINUOUS_INTEGRATION": "true"
+			},
+			"name": "CI tests (MATLAB)",
+			"working_dir": "${project_path}/test"
+		},
+		{
+			"cmd":
+			[
+				"matlab-hg1",
+				"-nosplash",
+				"-r",
+				"cd('${project_path}/test'); runMatlab2TikzTests()"
+			],
+			"env":
+			{
+				"CONTINUOUS_INTEGRATION": "true"
+			},
+			"name": "CI tests (MATLAB HG1)",
+			"working_dir": "${project_path}/test"
+		},
+		{
+			"cmd":
+			[
+				"matlab-hg2",
+				"-nosplash",
+				"-r",
+				"cd('${project_path}/test'); runMatlab2TikzTests()"
+			],
+			"env":
+			{
+				"CONTINUOUS_INTEGRATION": "true"
+			},
+			"name": "CI tests (MATLAB HG2)",
+			"working_dir": "${project_path}/test"
+		},
+		{
+			"cmd":
+			[
+				"make",
+				"-j8"
+			],
+			"name": "ACID: make",
+			"working_dir": "${project_path}/test/tex"
+		},
+		{
+			"cmd":
+			[
+				"make",
+				"distclean"
+			],
+			"name": "ACID: distclean",
+			"working_dir": "${project_path}/test/tex"
+		}
+	],
 	"folders":
 	[
 		{
 			"path": "."
 		}
 	],
-    "build_systems":
-     [
-        {
-            "name": "CI tests (Octave)",
-            "env": {"CONTINUOUS_INTEGRATION":"true"},
-            "working_dir": "${project_path}/test",
-            "cmd":  ["octave","-q","--eval","runMatlab2TikzTests.m"]
-        },
-        {
-            "name": "CI tests (MATLAB)",
-            "env": {"CONTINUOUS_INTEGRATION":"true"},
-            "working_dir": "${project_path}/test", // this is not respected by MATLAB R2015a depending on the user settings!
-            "cmd":  ["matlab", "-nosplash",
-                     "-r","cd('${project_path}/test'); runMatlab2TikzTests()"]
-        },
-        {
-            "name": "CI tests (MATLAB HG1)",
-            "env": {"CONTINUOUS_INTEGRATION":"true"},
-            "working_dir": "${project_path}/test", // this is not respected by MATLAB R2015a depending on the user settings!
-            "cmd":  ["matlab-hg1", "-nosplash",
-                     "-r","cd('${project_path}/test'); runMatlab2TikzTests()"]
-        },
-        {
-            "name": "CI tests (MATLAB HG2)",
-            "env": {"CONTINUOUS_INTEGRATION":"true"},
-            "working_dir": "${project_path}/test", // this is not respected by MATLAB R2015a depending on the user settings!
-            "cmd":  ["matlab-hg2", "-nosplash",
-                     "-r","cd('${project_path}/test'); runMatlab2TikzTests()"]
-        },
-        {
-            "name": "ACID: make",
-            "working_dir": "${project_path}/test/tex",
-            "cmd":  ["make","-j8"]
-        },
-        {
-            "name": "ACID: distclean",
-            "working_dir": "${project_path}/test/tex",
-            "cmd":  ["make","distclean"]
-        }
-    ]
+	"settings":
+	{
+		"FuzzyFilePath":
+		{
+		}
+	}
 }
```

egeerardyn:
Ah, I see. It's just the empty placeholder anyway. I will add it, because it seems that when it is present, the file doesn't get rewritten by that plugin. I will have a look at it, to see whether we can include more sane settings in the project file.

egeerardyn:
@PeterPablo Could you check on your end as well that the file doesn't get overwritten? Afterwards I'll merge.

PeterPablo:
Works for me.
Unrelated: Do the build processes give identical output for you, compared to executing them manually in the given environment?

egeerardyn:
The build environments work fine here: no failing tests whatsoever. Are you sure that you aren't calling Octave 4.0 by any chance (you might be able to diagnose that by removing the `-q` switch in there).

The only other thing that I can think of, is that maybe we should have the same workaround on Octave as  MATLAB, just in case somebody sets the PWD from their startup script (`.octaverc` for Octave, or something very similarly named).

PeterPablo:
Nope, it was was definitely the MATLAB build environment and R2014b got started (my default). But nevermind -- this is certainly unrelated and yet another aspect of the odyssey to reliable tests...

PeterPablo:
Oh, I just noticed that I cant press `ctrl+b` on a `*.tex` file anymore, since now the previous build system is chosen (e.g. matlab). Previously it would have executed `pdflatex`, since I have the [`LatexTools` addon](https://github.com/SublimeText/LaTeXTools) installed.

egeerardyn:
@PeterPablo That's true: it will always use the latest build system. Normally, you should be able to set it back to "automatic" or "pdflatex" from the menus. 

I am currently testing to make them filetype-specific: i.e. only run MATLAB/Octave when an m-file is active and the other ones can then still use LaTeX, I hope.

PeterPablo:
You should give the `ScopeHunter` a try to determine the correct selector, e.g. `text.tex.latex` for latex files. MATLAB seems to be `source.matlab`. Don't put too much energy into this. If you have something, feel free to merge. It certainly an improvement and I saw myself that I can choose `LaTex` from the Build menu manually.

egeerardyn:
@PeterPablo I have it working, but I first want to make sure that I can map a shortcut to the build menu, such that one can easily change the build system used (since we need HG1/HG2/Octave all the time).

PeterPablo:
`strg+shift+b` is the default shortcut to present you the menu, where you can select the build system

egeerardyn:
For one reason or another (and not <kbd>cmd</kbd> vs <kbd>ctrl</kbd>), that doesn't seem to work here (not even by redefining those manually). Anyway: I think this is ready, and until I fix my issue, I will click through the menus.

PeterPablo:
:+1:

PR #707, opened 2015-07-29
Fix page border of logo

egeerardyn:
This makes sure that other SVG editors show the whole logo, not only a part.
Not high-priority, but it bothered me every time I opened the file.

I don't know how well most SVG editors support font fallbacks, so I didn't have a go at that to also make sure that rendering is better when Liberation Sans is not installed (e.g. Illustrator CS6 dislikes it when not all fonts are installed on your system).

### DIFF
--- a/logos/matlab2tikz.svg
+++ b/logos/matlab2tikz.svg
@@ -9,12 +9,12 @@
    xmlns="http://www.w3.org/2000/svg"
    xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
    xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
-   width="744.09448819"
-   height="1052.3622047"
+   width="2078.8535"
+   height="846.6734"
    id="svg3051"
    version="1.1"
-   inkscape:version="0.48.5 r10040"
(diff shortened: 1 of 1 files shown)


PeterPablo:
I tested this by opening [this link](https://github.com/matlab2tikz/matlab2tikz/pull/707/files?short_path=9e440ab#diff-9e440abea8ec59225d27ac940894f328) in Chrome. It seems it is (still) not wide enough -- the `2` is cropped?!

egeerardyn:
I hope that is just a font issue. I have opened the file in both Inkscape and Affinity Designer (Illustrator CS6 refuses most Inkscape files due to font issues). You can have a look at what I see:

screen shot 2015-07-30 at 16 40 17
screen shot 2015-07-30 at 16 40 24

There, the borders seem to be fine (those were set using Inkscape's function to use the bounding box as borders), only in Inkscape it seems that a tiny part of the 2 is truncated.

I dont' really know what is going in, but I added a few pixels of margin to hopefully fix the issue in most renderer?

PeterPablo:
I started up Inkscape 0.91 and see
image
... there I also saw a font that I do not have Liberation Sans installed on my machine, so this is why...

:+1: feel free to merge

Issue #477, opened 2014-12-04
Clarify usage of `m2t` container object


PeterPablo:
Let's discuss the usage of the `m2` container object. It's history and some of it's purposes are described by @egeerardyn in comment https://github.com/matlab2tikz/matlab2tikz/commit/08dbd7c52197949a1a9abbdc093255e78ba84b89#commitcomment-8832480.

I understand now that it is used to remove the unwanted use of global variables ( :+1: ). Ideally it should at least be documented somewhere what kind of features make use of it and what general purpose it serves. I am unsure whether the code or the wiki here is better suited.

Moving away from this approach is certainly a difficult task and I would not suggest doing it right now. But I propose that we have a look at all the (new) subfunctions and check, if it is needed to pass the `m2t` object (check both for the **parameter** `m2t` and the **returned object** `m2t`).
## Todo list

@egeerardyn: I'm hijacking the first post to add a little TODO list for clarity.
- [x] [Group some m2t fields into an `About` sub-struct](https://github.com/matlab2tikz/matlab2tikz/issues/477#issuecomment-67893096) (this is tackled in a private branch for the updater)
- [ ] [Maybe even remove some of those fields](https://github.com/matlab2tikz/matlab2tikz/issues/477#issuecomment-67897064)
- [ ] [Put `axes`-related fields in `axesContainers{end}`](https://github.com/matlab2tikz/matlab2tikz/issues/477#issuecomment-69473097)
- [ ] Remove `m2t` input/output parameter in some functions where this is not necessary (beware: in some functions `m2t` might be needed in the future. See below for discussion).

okomarov:
I also noted that lots of axis specific stuff is in `m2t` rather than under `m2t.axesContainers{end}`:
```matlab
             axesBoundingBox: [0 0 1 1]
              axesContainers: {[1x1 struct]}
         gcaAssociatedLegend: []
                  unitlength: [1x1 struct]
               xAxisReversed: 0
               yAxisReversed: 0
               zAxisReversed: 0
```

okomarov:
I would also suggest to localize the color definitions with the suggestion in: 
http://tex.stackexchange.com/questions/24434/how-to-specify-a-fill-color-in-rgb-format-in-a-node-in-tikzpicture

In other words, get rid of the global `m2t.extraRgbColorNames`, `m2t.extraRgbColorSpecs` and use the RGB specification directly inside the plot:
    \addplot[fill={rgb:red,4;green,2;yellow,1},...]

PR #712, opened 2015-08-05
Format URLs and make clickable (MATLAB)

egeerardyn:
This is a low-priority fix for something that bothered me a bit.
There are a few small things that happen:
 - URLs to issues/wiki/github are now stored in `m2t.about`
 - these URLs are now clickable in MATLAB
 - I mainly refer to GitHub for interaction, since FEX is not the place to encourage bug reports IMO.

The console output looks as follows in MATLAB:
screen shot 2015-08-05 at 15 13 23

For Octave, it looks mostly like the old output:
screen shot 2015-08-05 at 15 12 58


src/matlab2tikz.m
```diff
@@ -3120,26 +3129,35 @@ function warnFacetedInterp(m2t)
         ['A 3D plot with "shader = faceted interp" is being produced.\n', ...
         'This may produce big and sluggish PDF files.\n', ...
         'See %s and Section 4.6.6 of the pgfplots manual for workarounds.'], ...
-        issueUrl(693, true));
+        issueUrl(m2t, 693, true));
 end
 % ==============================================================================
-function url = issueUrl(number, forOutput)
+function url = issueUrl(m2t, number, forOutput)
```
PeterPablo: I would pass `m2t.about.issues`, instead of the whole structure `m2t` (like it is done for the calls `clickableUrl` above).

PeterPablo:
Runtime tested on our supported environments. Feel free to merge. I leave it up to you to take my suggestion into account.
:+1:

src/matlab2tikz.m
egeerardyn: I think both have their merit. 

Yours is without any doubt cleaner in the amount/scope of data that is passed around: not the whole `m2t` god object is needed, just the issue part. :+1:

However, I slightly prefer my approach: in many functions we just pass `m2t` (which is not a best practice, I will admit, but at least we are aware of it: #477 ).  And it's a pattern in the code, so it's not surprising anymore to pass around `m2t`.

Indexing into `m2t` before calling `issueUrl`, just moves that logic (`m2t.about.issue`) to ALL call sites of the function, hence littering those places with unrelated/duplicate logic and possibility for error. Since my approach has that piece of logic in a single place, that cannot happen.

With #477, I think we can certainly have a look at this again. Since we're talking about (global) constants being passed around, I think there are quite a few options to pass those behind the scenes (e.g. wrap those in a function that can be called where needed).

Issue #715, opened 2015-08-06
Title is not showing in bodeplot


tomlankhorst:
The title is stored in an axis that has the property 'Visibility' set to off. It is not added to the list of 'relevantPlots' altough it has a visible 'title' or 'label'.

egeerardyn:
Should be fixed by #716/#753.

PR #716, opened 2015-08-06
Fixed missing title when visibility is set to off

tomlankhorst:
Potential fix for #715

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -686,8 +686,8 @@ function [m2t, pgfEnvironments] = handleAllChildren(m2t, h)
                 % supported by matlab2tikz or pgfplots.
 
             case ''
-                % No children found for handle.
-                % Carrying on as if nothing happened
+                % No children found for handle. (It has only a title and/or
+                % labels). Carrying on as if nothing happened
 
             otherwise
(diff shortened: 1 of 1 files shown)


tomlankhorst:
Hi Egon, in case of the bodeplot the Title of the figure is stored in an axis that has visibility set to Off. Despite the axis being set to invisible, MATLAB does show the title. Please not that this is still W.I.P. created this PR to discuss some things with you guys. 

``` MATLAB
G = tf(1,[1 1]);
H = tf([1 1], [1 1 1]);

bode(G,H);
legend('G(s)','H(s)');
t = title('Bode Test Title');

% The title
t.Parent.Title

% The visibility
t.Parent.Visible
```

egeerardyn:
Could you comment a bit on why/how this works, because I can't really understand from your code why we need to check for presence of a title to determine whether an axes is relevant to determining the bounding box or not.

Preferably, this should be made clear in the source code.

tomlankhorst:
You're right, this is too vague so I spent some more time to trace the problem down. The thing is, invisible axes containers are checked for visible children but not for visible titles or labels. I added this now. However, this can cause axis having no children (but a title and or label) to be processed resulting in a warning.

egeerardyn:
See #753.

PR #720, opened 2015-08-10
Save hash table selective

### LINKED ISSUE #634: saveHashTable when test is not executed
As noted in, e.g. in #605 by @okomarov, how `saveHashTable(status)` behaves when a test is not executed is not exactly as desired. However, given the simple file formats, it is always a feasible approach to manage the hashes by hand.

**Current behavior:**
for each test in the `status` array, write out the computed hash value to the file unless the test doesn't have a computed hash value (e.g. skipped tests, empty output, ...).

Downsides are:
 - skipped tests (e.g. due to missing toolboxes) have to be handled by hand
 - omitted tests (i.e. ones explicitly not executed) also have to be handled by hand

**What I'd propose instead:**
- first read in the existing hash table
- for each executed test in the `status` array, write out the hash
- for skipped tests, just retain the old hash and issue a warning

That would get rid of the first downside. The second downside, IMHO, is necessary to also handle the case when we remove test cases. To handle that more cleanly, we could always add a flag to `saveHashTable` to, either:
 - remove tests not found in `status` (i.e. current behavior and needed for removal of tests)
 - to retain hashes in the file that have not been specified in `status` (i.e. useful when only a part of the tests has been executed to generate `status`)

egeerardyn:
This fixes #634. By default, I just ask what action should be performed on tests for which there is a reference hash, but no corresponding test in the status array. The action can be set using the `removedTests` parameter.

I'm not completely satisfied with that name, so suggestions are more than welcome!

This PR can [skip ci] since Travis doesn't cover that part of our code.

### DIFF
--- a/test/saveHashTable.m
+++ b/test/saveHashTable.m
@@ -1,42 +1,135 @@
-function saveHashTable(status)
+function saveHashTable(status, varargin)
 % SAVEHASHTABLE saves the references hashes for the Matlab2Tikz tests
 %
 % Usage:
 %  SAVEHASHTABLE(status)
 %
+%  SAVEHASHTABLE(status, 'dryrun', BOOL, ...) determines whether or not to
+%  write the constructed hash table to file (false) or to stdout (true).
+%  Default: false
+%
+%  SAVEHASHTABLE(status, 'removedTests', CHAR, ...) specifies which action to
+%  execute on "removed tests" (i.e. test that have a hash recorded in the file,
+%  but which are not present in `status`). Three values are possible:
+%   - 'ask' (default): Ask what to do for each such test.
+%   - 'remove': Remove the test from the file.
+%               This is appropriate if the test has been removed from the suite.
+%   - 'keep': Keep the test hash in the file.
+%             This is appropriate when the test has not executed all tests.
+%
 % Inputs:
 %   - status: output cell array of the testing functions
 %
 % See also: runMatlab2TikzTests, testMatlab2tikz
+    ipp = m2tInputParser();
+    ipp = ipp.addRequired(ipp, 'status', @iscell);
+    ipp = ipp.addParamValue(ipp, 'dryrun', false, @islogical);
+    ipp = ipp.addParamValue(ipp, 'removedTests', 'ask', @isValidAction);
+    ipp = ipp.parse(ipp, status, varargin{:});
+
+    %% settings
     suite = status{1}.testsuite; %TODO: handle multiple test suites in a single array
     filename = hashTableName(suite);
     FILEFORMAT = '%s : %s\n';
+
+    %% process the hash table
     oldHashes = readHashesFromFile(filename);
+    newHashes = updateHashesFromStatus(oldHashes, status);
+    writeHashesToFile(filename, newHashes);
 
-    % sort by file names to allow humans better traversal of such files
-    funcNames = cellfun(@(s) s.function, status, 'UniformOutput', false);
-    [dummy, iSorted] = sort(funcNames); %#ok
-    status = status(iSorted);
+    % --------------------------------------------------------------------------
+    function hashes = updateHashesFromStatus(hashes, status)
+        % update hashes from the test results in status
+        oldFunctions = fieldnames(hashes);
+        newFunctions = cellfun(@(s) s.function, status, 'UniformOutput', false);
 
-    % write to file
-    fid = fopen(filename,'w+');
-    finally_fclose_fid = onCleanup(@() fclose(fid));
-
-    for iFunc = 1:numel(status)
-        S = status{iFunc};
-        thisFunc = S.function;
-        if isfield(S.hashStage,'found')
-            thisHash = S.hashStage.found;
-        else
-            warning('SaveHashTable:NoHashFound',...
-                    'No hash found for "%s". Assuming empty.', S.function);
-            thisHash = ''; % FIXME: when does this happen??
+        % add hashes from all executed tests
+        for iFunc = 1:numel(status)
+            S = status{iFunc};
+            thisFunc = S.function;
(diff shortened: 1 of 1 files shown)


test/saveHashTable.m
```diff
@@ -1,37 +1,159 @@
...
+            fprintf(fid, FILEFORMAT, func, hashes.(func));
         end
-        if ~isempty(thisHash)
-            fprintf(fid, '%s : %s\n', thisFunc, thisHash);
+    end
+    function hashes = readHashesFromFile(filename)
+        % read hashes from a file
+        if exist(filename','file')
+            fid = fopen(filename, 'r');
+            closeFileAfterwards = onCleanup(@() fclose(fid));
```
PeterPablo: :+1: in general, but you  might want to make the naming consistent, e.g., `finally_fclose_fid`.

test/saveHashTable.m
```diff
@@ -1,37 +1,159 @@
...
+        % ask which action to carry out on a removed test
+        action = lower(ipp.Results.removedTests);
+        validActions = {'remove', 'keep'};
+        while ~ismember(action, validActions)
+            query = sprintf('Remove or keep "%s"? [Rk]:', testName);
+            answer = strtrim(input(query,'s'));
+
+            if isempty(answer) || strcmpi(answer(1), 'R')
+                action = 'remove';
+            elseif strcmpi(answer(1), 'K')
```
PeterPablo: Doesn't it make more sense to make `keep` the default action? A developer without some of the toolboxes might accidentally remove tests he can't run?!

test/saveHashTable.m
```diff
@@ -1,37 +1,159 @@
...
+                    fprintf(1, 'Kept hash for "%s"\n', thisTest);
+
+            end
+        end
+    end
+    function action = askActionToPerformOnRemovedTest(testName)
+        % ask which action to carry out on a removed test
+        action = lower(ipp.Results.removedTests);
+        validActions = {'remove', 'keep'};
+        while ~ismember(action, validActions)
```
PeterPablo: did you intend to call `isValidAction` here?

egeerardyn: No, not really. This function gets rid of the action `'ask'` in favor of the actual actions.

test/saveHashTable.m
egeerardyn: It certainly makes more sense. It's just that the current behavior (before this PR) is `remove`. But you're absolutely right. On the user-facing side, I try to be conservative with changing those things. But dev-facing we can indeed be more pragmatic and just do the more logical thing.

PeterPablo: :+1: 
and :+1: for the more logical thing

test/saveHashTable.m
PeterPablo: My fault. Now I recognize the "recursive" approach. You might want to add a one line comment near `ask` in the loop.

test/saveHashTable.m
```diff
@@ -1,37 +1,159 @@
...
+                case 'keep'
+                    % useful when not all tests were executed by the tester
+                    fprintf(1, 'Kept hash for "%s"\n', thisTest);
+
+            end
+        end
+    end
+    function action = askActionToPerformOnRemovedTest(testName)
+        % ask which action to carry out on a removed test
+        action = lower(ipp.Results.removedTests);
```
PeterPablo: I did not yet runtime test, but am wondering: why is `ipp` accessible in the scope of that function?

egeerardyn: That's because most of these functions are [nested functions](http://nl.mathworks.com/help/matlab/matlab_prog/nested-functions.html) (not sub-functions!).

Regarding variables, scoping of a nested function = parameters + variables of parent + own variables.
In Octave there are some things to watch out for (i.e. you can't pass handles to nested functions around), which are acceptable in MATLAB.
Recent versions of MATLAB (2009b or newer, I think) also highlight variables that cross scopes like that (by default in a blue-greenish color).

The same mechanism is used for `FILEFORMAT` by the way.

PeterPablo: ah! Didn't notice this while looking at the code using GitHub. Alright then :-)

test/saveHashTable.m
```diff
@@ -1,37 +1,163 @@
...
+    ipp = m2tInputParser();
+    ipp = ipp.addRequired(ipp, 'status', @iscell);
+    ipp = ipp.addParamValue(ipp, 'dryrun', false, @islogical);
+    ipp = ipp.addParamValue(ipp, 'removedTests', 'ask', @isValidAction);
+    ipp = ipp.parse(ipp, status, varargin{:});
+
+    %% settings
     suite = status{1}.testsuite; %TODO: handle multiple test suites in a single array
     filename = hashTableName(suite);
+    FILEFORMAT = '%s : %s\n';
```
okomarov: In Win this is a problem, it should be `\r\n`. However, you can actually skip it altogether and let textscan deal with it internally, i.e. `'%s : %s'` should work OS independently.

test/saveHashTable.m
```diff
@@ -1,37 +1,163 @@
...
+            % data is now a cell array with 2 elements, each a (row) cell array
+            %  - the first is all the function names
+            %  - the second is all the hashes
+
+            % Transform `data` into {function1, hash1, function2, hash2, ...}'
+            % First step is to transpose the data concatenate both fields under
+            % each other. Since MATLAB indexing uses "column major order",
+            % traversing the concatenated array is in the order we want.
+            dataTransposed = cellfun(@transpose, data, 'UniformOutput', false);
+            allValues = vertcat(dataTransposed{:});
```
okomarov: I find the following easier to read:

```
data = [data{:}]';
allValues = data(:)';
```

PR #725, opened 2015-08-11
Figure2dot improvements

egeerardyn:
When I was trying to debug some stuff in Octave, I noticed that `figure2dot` doesn't work in Octave. I took the opportunity to clean up the code a bit and add some new features.

This `figure2dot` function allows to translate a HG object into a DOT (GraphViz) file that represents the parent/children relationship and shows some of the relevant properties for each node. This allows to construct figures such as

test74-tree

### DIFF
--- a/src/figure2dot.m
+++ b/src/figure2dot.m
@@ -74,6 +74,7 @@ function figure2dot(filename, varargin)
             catch
                 % don't do anything
             end
+            label = addProperty(label, 'Handle', sprintf('%g', double(h)));
             label = addHGProperty(label, h, 'Title', '');
             label = addHGProperty(label, h, 'String', '');
             label = addHGProperty(label, h, 'Tag', '');
--- a/src/figure2dot.m
+++ b/src/figure2dot.m
@@ -69,7 +69,7 @@ function figure2dot(filename, varargin)
             label = {};
             label = addHGProperty(label, h, 'Type', '');
             try
-                hClass = class(handle(child));
+                hClass = class(handle(h));
                 label = addProperty(label, 'Class', hClass);
             catch
                 % don't do anything
--- a/src/figure2dot.m
+++ b/src/figure2dot.m
@@ -114,7 +114,7 @@ function label = addHGProperty(label, h, propName, default)
 
     if isprop(h, propName)
         propValue = get(h, propName);
-        if ishghandle(propValue) && isprop(propValue, 'String')
+        if numel(propValue) == 1 && ishghandle(propValue) && isprop(propValue, 'String')
             % dereference Titles, labels, ...
             propValue = get(propValue, 'String');
         elseif iscellstr(propValue)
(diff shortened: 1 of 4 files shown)


egeerardyn:
@matlab2tikz/developers 
Could somebody take a look at this? Otherwise I merge this PR tomorrow night:
 - it works here,
 - I think most interesting properties are included,
 - this is just a debugging tool (hence the development label) and no other code relies on this: so our users are not likely to notice anything anyway.

src/figure2dot.m
```diff
@@ -29,127 +29,108 @@ function figure2dot(filename)
...
+        children = allchild(h);
+
+        for h = children(:)'
+            if shouldSkip(h), continue, end;
+            node_number = node_number + 1;
+
+            label = {};
+            label = addHGProperty(label, h, 'Type', '');
+            try
+                hClass = class(handle(child));
```
okomarov: This will always fail since `child` is not defined anymore, but should be `h`. For Octave you could retrieve the 'type' property instead.

egeerardyn: Thanks for spotting this!

src/private/m2tstrjoin.m
```diff
@@ -0,0 +1,34 @@
...
+            error('matlab2tikz:join:NotCellstrOrNumeric',...
+                'Expected cellstr or numeric.');
+        end
+    end
+
+    % inspired by strjoin of recent versions of MATLAB
+    newstr = cell(2,nElem);
+    newstr(1,:)         = reshape(cellstr, 1, nElem);
+    newstr(2,1:nElem-1) = {delimiter}; % put delimiters in-between the elements
+    newstr = [newstr{:}];
```
PeterPablo: Please ensure compatibility with Octave 4. See https://github.com/PeterPablo/matlab2tikz/commit/f2479095f9e29a6ee4af04d4989284841b1ca1f7

okomarov:
Tested on Octave 4.0.0, no errors, and graph visualized on http://www.webgraphviz.com/ 
Good to go :+1:

src/figure2dot.m
```diff
@@ -29,127 +29,108 @@ function figure2dot(filename)
...
+    function plot_children(fh, h, parent_node)
+
+        children = allchild(h);
+
+        for h = children(:)'
+            if shouldSkip(h), continue, end;
+            node_number = node_number + 1;
+
+            label = {};
+            label = addHGProperty(label, h, 'Type', '');
```
okomarov: This fails in Matlab

egeerardyn: What version of MATLAB and figure are you using? Is this still present after my last commit (just now)?

egeerardyn:
In the last commit, I fixed some troubles with `cell` values.
For me, it runs fine on all ACID figures in Octave 3.8.0, R2014a and R2014b using the following snippet
```matlab
for i = 1:numel(ACID(0))
  figure;
  ACID(i)
  figure2dot(tempname)
  close all
end
```

okomarov:
Checked, all good now.

egeerardyn:
Thank you for reviewing and forcing me to test the code more diligently :+1:

PR #727, opened 2015-08-13
Fix bar width

RPeschke:
Hi, 

I had the problem that the bar width was not correct in the latex output. I found that the problem was just that the bar with in matlab is in relative units where it is in absolut units in latex. 

another problem I encountered was on the latex side it only uses axis units and not pixel units if you use:
```latex
\pgfplotsset{compat=newest}
```
or at least one has to use: 
```latex
\pgfplotsset{compat=1.7}
```


egeerardyn:
Thank you for looking into this. I will try to review this tonight.
We should warn about that minimal pgfplots version. I thought we did, but we don't seem to do so for bar plots. Thank you for noticing!

src/matlab2tikz.m
```diff
@@ -3424,6 +3424,16 @@ case size(cData,2) % specific y-data is given
     end
 end
 % ==============================================================================
+function BarWidth=getCorrectBarWidth(h)
```
PeterPablo: Could you please add short comment to describe the goal of the function.

src/matlab2tikz.m
```diff
@@ -3424,6 +3424,16 @@ case size(cData,2) % specific y-data is given
...
+XData_bar=get(h,'XData');
+length_bar = length(XData_bar);
+BarWidth= get(h, 'BarWidth');
+if length_bar > 1
+    BarWidth = min(diff(XData_bar))*BarWidth;
```
PeterPablo: I am wondering why you use the `min`imal difference. Would `mean` make more sense?

PeterPablo:
@RPeschke thank you for your patch! Do you have a minimal working example that is fixed with your change?

src/matlab2tikz.m
okomarov: The *Correct* part is not informative in the name of the function. I suggest to either drop it or add a more specific adjective but such that the name starts with `getBar...`

RPeschke:
@PeterPablo

I tried it with the “mean” first but it seems that matlab internally uses the “min” function I reverse engineered it by creating Plots with unevenly spaced data sets. 


I added also an example file to the pull request which makes a plot and exports it to latex.

example/example_bar_plot.m
```diff
@@ -0,0 +1,53 @@
+function example_bar_plot()
```
egeerardyn: Please either move the `example` folder into the `test` folder to avoid cluttering the root folder. You could also make this into a test suite (see the other functions in `test/suites`), but I think in this case that might be more work and less convenient to execute.

egeerardyn:
Thank you for contributing. :+1: I have merged this into `develop`, please file future pull requests against that branch.

stiffsen:
The above fix doesn't work for me. I patched my local copy of matlab2tikz with these changes, because it is apparently not included in the master, yet. However, the newly created function getBarWidthInAbsolutUnits() returns the same value as before, because my bars are placed in a distance of 1. That lead to the "width=0.8" parameter in the tikz file, resulting in way too thin bars. I fixed the issue for me, but I am not sure if that fix is universally applicable:
I changed the line

    drawOptions = opts_add(drawOptions, 'bar width', ...
        formatDim(barWidth, ''));

to

    drawOptions = opts_add(drawOptions, 'bar width', ...
        formatDim(barWidth/(numel(get(h,'XData'))+1), '\figurewidth'));

which effectively relates the barwidth to the figurewidth.

I am on Matlab R2013b.

RPeschke:
HI,

maybe the problem is inside your latex file. if you don't set 
\pgfplotsset{compat=newest}
the width of the bar is given in different units.

egeerardyn:
There are still bugs in determining the bar width, e.g. #785. In any case, as @RPeschke says, the current code depends strongly on using pgfplots 1.7 or newer. Given your description, I suspect your version of pgfplots might be outdated.

Unfortunately, the fix is not as easy as the one you provide:
 - that only works for a single subplot,
 - when one specifies `\figurewidth`, and
 - I suspect when there are no grouped bars.

But we will look into fixing this behavior.

stiffsen:
Oh yeah, I am new to this whole pgfplots thing, so I must've missed that. 

And by the way, thanks a bunch. adding that line right there fixed a whole bunch of issues for me!! :-)

PR #730, opened 2015-08-13
bugfix for #728

### LINKED ISSUE #728: drawAnnotations fails: "Invalid handles passed to findall."
I have a MATLAB fig file which I opened and then called `matlab2tikz` (latest version from development branch)

It failed with:
<pre>
Error using findall (line 24)
Invalid handles passed to findall.

Error in matlab2tikz>drawAnnotations (line 2538)
        annotHandles = findall(allchild(annotPanes),'Visible','on');

Error in matlab2tikz>saveToFile (line 454)
    m2t = drawAnnotations(m2t);

Error in matlab2tikz (line 347)
saveToFile(m2t, fid, fileWasOpen);
</pre>

This happened in R2015a on Linux.
The fig-file which caused the bug is available at https://gist.github.com/dachziegel/df4f74df9fca63556119#file-test-fig

dachziegel:
fixes #728

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -2535,7 +2535,12 @@ function m2t = drawAnnotations(m2t)
     % Get annotation handles
     if isHG2
         annotPanes   = findall(m2t.currentHandles.gcf,'Tag','scribeOverlay');
-        annotHandles = findall(allchild(annotPanes),'Visible','on');
+        children = allchild(annotPanes);
+        if all(cellfun(@isempty,children))
+            annotHandles = {};
+        else
+            annotHandles = findall(allchild(annotPanes),'Visible','on');
+        end
     else
         annotHandles = findall(m2t.scribeLayer,'-depth',1,'Visible','on');
     end
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -2539,7 +2539,7 @@ function m2t = drawAnnotations(m2t)
         if all(cellfun(@isempty,children))
             annotHandles = {};
         else
-            annotHandles = findall(allchild(annotPanes),'Visible','on');
+            annotHandles = findall(children,'Visible','on');
         end
     else
         annotHandles = findall(m2t.scribeLayer,'-depth',1,'Visible','on');
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -2536,11 +2536,10 @@ function m2t = drawAnnotations(m2t)
     if isHG2
         annotPanes   = findall(m2t.currentHandles.gcf,'Tag','scribeOverlay');
         children = allchild(annotPanes);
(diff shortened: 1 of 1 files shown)


okomarov:
It's a bit of a weird edge case. Could you elaborate on the workflow that's given you that type of figure, i.e. two annotation layers without any annotations?

I am asking because I didn't know Matlab allowed multiple annotation layers, but since it does, your fix might not generalize to e.g. one layer with annotations and another empty one.

dachziegel:
Unfortunately all I have left is the MATLAB fig-file and I wanted to pretty-print it using `matlab2tikz`.

src/matlab2tikz.m
```diff
@@ -2535,7 +2535,12 @@ function assertRegularAxes(handle)
     % Get annotation handles
     if isHG2
         annotPanes   = findall(m2t.currentHandles.gcf,'Tag','scribeOverlay');
-        annotHandles = findall(allchild(annotPanes),'Visible','on');
+        children = allchild(annotPanes);
+        if all(cellfun(@isempty,children))
+            annotHandles = {};
+        else
+            annotHandles = findall(allchild(annotPanes),'Visible','on');
```
egeerardyn: You can use `children` here instead of calling `allchild(annotPanes)` again.

egeerardyn:
There are a few cases that are not handled well here:
 - if there are 0 or 1 annotation panes, `cellfun` will fail because `children` is not a cell. You can check this by running `testHeadless` (as indicated in [test/README.md](https://github.com/matlab2tikz/matlab2tikz/blob/develop/test/README.md)). That case is extremely common, so it should be dealt with properly.
 - less importantly, if one of the annotation panes does contain annotations (e.g. manually add an arrow to your FIG file), this throws an error again.

It's a shame MATLAB doesn't handle those silly cases itself, but I think the following fixes all cases:
```matlab
children = allchild(annotPanes);
if iscell(children)
   children = cell2mat(children);
end
```

dachziegel:
@egeerardyn I am not sure how the bugfix would look like with your suggestion I just tried it with my figure and I got 

<pre>
Error using cell2mat (line 52)
CELL2MAT does not support cell arrays containing cell arrays or objects.

Error in matlab2tikz>drawAnnotations (line 2540)
          children = cell2mat(children);

</pre>

egeerardyn:
Hmm, I think I see what happened with my "fix": it never was called in the test cases I ran.
If we replace `cell2mat(children)` with `[children{:}]`, however, it should be fine. I checked your MWE and the same with an annotation added. Our regular test cases don't trigger that `iscell` branch, so they should remain fine.

dachziegel:
@egeerardyn I changed to your suggestion and it worked for my figure

egeerardyn:
Thank you for your work :+1:

Issue #723, opened 2015-08-11
contourf with iso-lines that are not closed


egeerardyn:
The current implementation of handling filled contour plots in HG2 assumes that all isolines are closed. However, this is not always the case, e.g.

```matlab
N = 20;
[X, Y] = meshgrid(1:N, 1:N);
Z = sqrt(X.^2 ./ Y.^2);
contourf(X+sin(Y), Y, Z, 50, 'EdgeColor', 'k');
matlab2tikz('test.tex','standalone',true);
!pdflatex test.tex
```

This produces the following output:

<img width="1064" alt="screen shot 2015-08-11 at 15 36 47" src="https://cloud.githubusercontent.com/assets/177360/9198787/cf80eb5e-403e-11e5-8406-1d3bf4933293.png">

What we see is that the iso-lines cross the boundary, but are cycled in a straight line by TikZ. Instead, it should be cycled over the bounds of the domain. A possible hurdle that I see is: what part of the boundary: the clockwise or counterclockwise completion?

sacraig:
I'll point out that immediately following the issue I submitted that led you to discover this issue, I have now run into it. I'd actually hazard a guess that in the scientific world, these kinds of contours are actually the most common.

okomarov:
@egeerardyn This example is quite complicated. I would start with something simpler, i.e. when the contours domain covers the whole figure. 

Then there are two cases for open contours:
1 open contour that end on same edge of the box. 
2. open contour that goes from one edge to another

The first case does not involve detecting which direction to cycle, you just need to repeat the first coordinate.

The second can be taken care by using the info on whether it's a hill or a valley and comparing on which side is the next contour of the group.

egeerardyn:
@okomarov you are right this is immediately the full complexity case. In a first step we could cycle over the bounding box around the domain and later on refine that to the actual domain as you said.

During development, your suggestion is a great breakdown to start writing, but for the implementation, we only need the second case (as the first case is also covered by that). My conclusion: let's come up with a contour plot that covers your cases in a simple way.

egeerardyn:
This is just a minor update on what I think is needed to implement this:

 - Determine the boundary of the plot (or first its bounding box, as we do now)
 - For each iso-line:
    * Check if it's non-closed (i.e. first point ~= last point), just to be sure.
[...]

PR #737, opened 2015-08-20
Vectorize segmentsIntersect and segmentVisible functions.

miscco:
modified:   src/cleanfigure.m

So here is the vectorized version of segmentsIntersect and segmentVisible. Some things to note:
1. The function which mow takes up the main chunk of segmentsIntersect was crossLines. This function is used later on in move to Box. I havent had time to check whether this can be vectorized too.
2. So for my personal use cases (my own figures) This speeds up cleanfigure from about 35s to 7s. However, this doesnt seem to be the case for the testcase with with a simple sinus. I cant really tell why.
3. I only have R2015 so some additional tests fail due to that, here the output from travis:
# MATLAB 8.5 (Linux 3.13.0-62-generic) commit 6fa89d1547 

## Unreliable tests
These do not cause the build to fail.

|   Testcase |                   Name |                       OK |                                                                      Status |
| :--------- | :--------------------- | :----------------------: | :-------------------------------------------------------------------------- |
|  `ACID(1)` |     `multiline_labels` | :heavy_exclamation_mark: | hash 21a208748863224355861184cf9910f7 != (01d4eeae27254d91a0f8316897e23e81) |
|  `ACID(6)` | `sine_with_annotation` | :heavy_exclamation_mark: | hash 995cd2a901446c59f85552d48025ba27 != (7d415c014eec35c249802eb3a3ee519e) |
| `ACID(10)` |       `peaks_contourf` | :heavy_exclamation_mark: | hash 1241fa010c4c53ab967f1e61931b596f != (77bfa3a9f07e49b9061d919551ad0a0f) |
| `ACID(16)` |              `logplot` |       :white_check_mark: |                                                                             |
| `ACID(18)` |           `legendplot` | :heavy_exclamation_mark: | hash d37d19ba36d911a0db60095db48b08ef != (746423d1c0d403421667221a6b0528bf) |
| `ACID(24)` |          `quiver3plot` |       :white_check_mark: |                                                                             |
| `ACID(34)` |                 `bars` |       :white_check_mark: |                                                                             |
| `ACID(38)` |          `subplot2x2b` |       :white_check_mark: |                                                                             |
| `ACID(40)` |        `subplotCustom` |       :white_check_mark: |                                                                             |
| `ACID(42)` |            `bodeplots` | :heavy_exclamation_mark: | hash ca7b2029a08a2a311a94b73339e2128a != (8f6cafc1771cc3a506c8fc46c15ec43e) |
| `ACID(43)` |           `rlocusPlot` |       :white_check_mark: |                                                                             |
| `ACID(48)` |          `zplanePlot2` | :heavy_exclamation_mark: | hash 4c2f1fea92ce3c969475e3f6b40f87cf != (fb3ba1a4db46c9975cd00e6da99ebe48) |
| `ACID(58)` |            `surfPlot2` | :heavy_exclamation_mark: | hash 520bba1c7a536efb15c82dfcc8ce6a23 != (eb8e6d18db994ba4a90e177bd44b5db3) |
| `ACID(94)` | `stackedBarsWithOther` |       :white_check_mark: |                                                                             |
| `ACID(97)` |     `overlappingPlots` |       :white_check_mark: |                                                                             |

## Reliable tests
Only the reliable tests determine the build outcome.
Passing tests are not shown (only failed and skipped tests).

|   Testcase |              Name |                       OK |                                                                      Status |
| :--------- | :---------------- | :----------------------: | :-------------------------------------------------------------------------- |
|  `ACID(8)` |   `peaks_contour` | :heavy_exclamation_mark: | hash 4c0f04124b13d5c12351f4875455a67d != (8885c9a4185aaca594ce5d0139aa578b) |
|  `ACID(9)` |    `contourPenny` | :heavy_exclamation_mark: | hash 1bbca20d0029e2e510eb5d7c46222ee7 != (3ee200ec93371666499d72f790ee9bb3) |
| `ACID(12)` | `double_colorbar` | :heavy_exclamation_mark: | hash 44a0d0b4b35ed634d2f2f568e51f2c8a != (6355484ba42b84205cd6a7d6de76711d) |
| `ACID(14)` |     `double_axes` | :heavy_exclamation_mark: | hash a053ed16097389ed6d6cfe6f353a46ab != (331441239c1bb077aebaf5ce8abe55f1) |
| `ACID(15)` |    `double_axes2` | :heavy_exclamation_mark: | hash ec8ccb7d178fd3e0227fd2cf3b210fa0 != (739edbf88cbf57187bea11ef93ebbf9a) |
| `ACID(23)` |      `quiverplot` | :heavy_exclamation_mark: | hash 6f6723cafaf167b8e13d62f56b5a0aff != (308f4ce197791f2fd5862768e1157ca8) |
| `ACID(62)` |         `spectro` | :heavy_exclamation_mark: | hash e871a0b852a68fe12d78eead39e3de3b != (4426bb06bb44d08d016595f585fb01da) |
| `ACID(71)` |   `parameterSurf` | :heavy_exclamation_mark: | hash e64030978c54cc260847bf78a8c6ea6a != (12522a58d7e3d8dc7f4e38a0970bf047) |
| `ACID(17)` | `colorbarLogplot` |          :grey_question: |                                                                     SKIPPED |

## Test summary
Test results for m2t commit 6fa89d1547 running with MATLAB 8.5 on Linux
 3.13.0-62-generic.

```matlab
suite = @ACID;
alltests = [1:99];
reliable = [2:5 7:9 11:15 17 19:23 25:33 35:37 39 41 44:47 49:57 59:93 95:96 98:99];
unreliable = [1 6 10 16 18 24 34 38 40 42:43 48 58 94 97];
failReliable = [8:9 12 14:15 23 62 71];
passUnreliable = [16 24 34 38 40 43 94 97];
skipped = [17];
```

| Unreliable |    8 |    7 |    0 |    15 |
|   Reliable |   75 |    8 |    1 |    84 |
|      Total |   83 |   15 |    1 |    99 |

Build fails with 8 errors. :heavy_exclamation_mark:

### DIFF
--- a/src/cleanfigure.m
+++ b/src/cleanfigure.m
@@ -320,7 +320,7 @@ function out = segmentVisible(data, dataIsInBox, xLim, yLim)
     end
 end
 % =========================================================================
-function out = segmentsIntersect(data, dataDiff, X3, X4)
+function out = segmentsIntersect(X1, diffX2X1, X3, X4)
   % Checks whether the segments X1--X2 and X3--X4 intersect.
   % Given four points X_k=(x_k,y_k), k\in{1,2,3,4}, and the two lines defined
   % by those,
@@ -336,22 +336,18 @@ function out = segmentsIntersect(data, dataDiff, X3, X4)
   %
   % for lambda1 and lambda2.
 
-  % NOTE: We could vectorize this function. Now data is a matrix containing
-  % all points X1 and dataDiff is a matrix containing the differences
+  % NOTE: We could vectorize this function. Now X1 is a matrix containing
+  % all points X1 and diffX2X1 is a matrix containing the differences
   % X2-X1 for all points X1 and X2
   % n is the number of segments (not points in the plot!)
-  n   = size(dataDiff, 1);
+  n   = size(diffX2X1, 1);
   out = false(n, 1);
-  
-  % Rotational matrix with sign flip. It transforms a given vector [a,b] by 
-  % Rotate * [a,b] = [-b,a] as required for calculation of invA and detA  
-  Rotate = [0, -1; 1, 0];   
   
   % Calculate the determinant of A = [X2-X1, -(X4-X3)];
   % detA = -(X2(1)-X1(1))*(X4(2)-X3(2)) + (X2(2)-X1(2))*(X4(1)-X3(1))
   % NOTE: Vectorized this is equivalent to the matrix multiplication
   % [nx2] * [2x2] * [2x1] = [nx1]
-  detA = dataDiff * Rotate * (X4-X3);  
+  detA = -diffX2X1(:,1) .* (X4(2)-X3(2)) + diffX2X1(:,2) .* (X4(1)-X3(1));  
   
   % Get the indexes for nonzero elements
   id_detA = detA~=0;
@@ -360,16 +356,20 @@ function out = segmentsIntersect(data, dataDiff, X3, X4)
       % rhs = X3(:) - X1(:)
       % NOTE: Originaly this was a [2x1] vector. However as we vectorize the 
       % calculation it is beneficial to treat it as an [nx2] matrix rather than a [2xn]
-      rhs = bsxfun(@minus, X3', data(1:end-1, :));
+      rhs = bsxfun(@minus, X3', X1(1:end-1, :));
       
       % Calculate the inverse of A and lambda
       % invA=[-(X4(2)-X3(2)), X4(1)-X3(1);...
       %       -(X2(2)-X1(2)), X2(1)-X1(1)] / detA 
       % lambda = invA * rhs
+  
+      % Rotational matrix with sign flip. It transforms a given vector [a,b] by 
+      % Rotate * [a,b] = [-b,a] as required for calculation of invA  
+      Rotate = [0, -1; 1, 0];   
+
 
       % Rather than calculating invA first and then multiply with rhs to obtain 
-      % lambda, directly calculate the respective terms
-
+      % lambda, directly calculate the respective termsq
       % The upper half of the 2x2 matrix is always the same and is given by:
       % [-(X4(2)-X3(2)), X4(1)-X3(1)] / detA * rhs
       % This is a matrix multiplication of the form [1x2] * [2x1] = [1x1]
@@ -383,7 +383,7 @@ function out = segmentsIntersect(data, dataDiff, X3, X4)
       % matrix multiplication leading to a [nx1] vector. Therefore, use the
       % elementwise multiplication and sum over it
       % sum( [nx2] * [2x2] .* [nx2], 2) = sum([nx2],2) = [nx1] 
-      lambda2 = sum(-dataDiff(id_detA, :) * Rotate .* rhs(id_detA, :), 2)./detA(id_detA);
+      lambda2 = sum(-diffX2X1(id_detA, :) * Rotate .* rhs(id_detA, :), 2)./detA(id_detA);
 
       % Check whether lambda is in bound
       out(id_detA) = 0.0 < lambda1 & lambda1 < 1.0 & 0.0 < lambda2 & lambda2 < 1.0;
--- a/src/cleanfigure.m
+++ b/src/cleanfigure.m
@@ -296,36 +296,93 @@ function out = segmentVisible(data, dataIsInBox, xLim, yLim)
     % in p and determine whether the line between the pair crosses the box.
     n = size(data, 1);
     out = false(n-1, 1);
-
-    [bottomLeft, topLeft, bottomRight, topRight] = corners(xLim, yLim);
-
-    for k = 1:n-1
-        this = data(k  , :);
-        next = data(k+1, :);
-
+    % Only check if there is more than 1 point    
+    if n>1
         % One of the neighbors is inside the box and the other is finite
-        nextVisible = (dataIsInBox(k+1) && all(isfinite(this)));
-        thisVisible = (dataIsInBox(k)   && all(isfinite(next)));
+        nextVisible = (dataIsInBox(2:end)   & all(isfinite(data(1:end-1)),2));
+        thisVisible = (dataIsInBox(1:end-1) & all(isfinite(data(2:end)),2));
 
-        % Check whether the line connecting this point and the next one
-        % intersects with any of the borders of the drawn axis
-        left   = segmentsIntersect(this, next, bottomLeft , topLeft);
-        right  = segmentsIntersect(this, next, bottomRight, topRight);
-        bottom = segmentsIntersect(this, next, bottomLeft , bottomRight);
-        top    = segmentsIntersect(this, next, topLeft    , topRight);
+        % Get the corner coordinates
+        [bottomLeft, topLeft, bottomRight, topRight] = corners(xLim, yLim);
 
-        % The segment is visible when any of the following hold:
-        %  - this point is visible in the axis and the next is finite
-        %  - this point is finite and the next is visible in the axis
-        %  - the segment connecting this and the next point crosses a border
-        out(k) = thisVisible || nextVisible || left || right || top || bottom;
+        % Calculate the difference between two consequtive points
+        dataDiff = diff(data);
+
+        left   = segmentsIntersect(data, dataDiff, bottomLeft , topLeft);
+        right  = segmentsIntersect(data, dataDiff, bottomRight, topRight);
+        bottom = segmentsIntersect(data, dataDiff, bottomLeft , bottomRight);
+        top    = segmentsIntersect(data, dataDiff, topLeft    , topRight);
+
+        % Check the result
+        out = any([thisVisible, nextVisible, left, right, top, bottom],2);
     end
 end
 % =========================================================================
-function out = segmentsIntersect(X1, X2, X3, X4)
+function out = segmentsIntersect(data, dataDiff, X3, X4)
   % Checks whether the segments X1--X2 and X3--X4 intersect.
-  lambda = crossLines(X1, X2, X3, X4);
-  out = all(lambda > 0.0) && all(lambda < 1.0);
+  % Given four points X_k=(x_k,y_k), k\in{1,2,3,4}, and the two lines defined
+  % by those,
+  %
+  %  L1(lambda) = X1 + lambda (X2 - X1)
+  %  L2(lambda) = X3 + lambda (X4 - X3)
+  %
+  % returns the lambda for which they intersect (and Inf if they are parallel).
+  % Technically, one needs to solve the 2x2 equation system
+  %
+  %   x1 + lambda1 (x2-x1)  =  x3 + lambda2 (x4-x3)
+  %   y1 + lambda1 (y2-y1)  =  y3 + lambda2 (y4-y3)
+  %
+  % for lambda1 and lambda2.
+
+  %TODO: why don't we use `\` instead of Cramer's rule?
+
+  n   = size(dataDiff,1);
+  out = false(n,1);
+
+  % Rotational matrix with sign flip 
+  Rotate    = [0, -1; 1, 0 ];   
+  
+  % Calculate the determinant of A = [X2-X1, -(X4-X3)];
+  % detA = -(X2(1)-X1(1))*(X4(2)-X3(2)) + (X2(2)-X1(2))*(X4(1)-X3(1))
+  % NOTE: Vectorized this is equivalent to the matrix multiplication
+  % [nx2] * [2x2] * [2x1] = [nx1]
+  detA   = dataDiff * Rotate * (X4-X3);  
+  
+  % Get the indexes for nonzero elements
+  id_detA = detA~=0;
+  
+  if any(id_detA)
+      % rhs = X3(:) - X1(:)
+      % NOTE: Originaly this was a [2x1] vector. However as we vectorize the 
+      % calculation it is beneficial to treat it as an [nx2] matrix rather than a [2xn]
+      rhs = bsxfun(@minus, X3', data(1:end-1,:));
+      
+      % Calculate the inverse of A and lambda
+      % invA=[-(X4(2)-X3(2)), X4(1)-X3(1);...
+      %       -(X2(2)-X1(2)), X2(1)-X1(1)] / detA 
+      % lambda = invA * rhs
+
+      % Rather than calculating invA first and then multiply with rhs to obtain 
+      % lambda, directly calculate the respective terms
+
+      % The upper half of the 2x2 matrix is always the same and is given by:
+      % [-(X4(2)-X3(2)), X4(1)-X3(1)] / detA * rhs
+      % This is a matrix multiplication of the form [1x2] * [2x1] = [1x1]
+      % As we have transposed rhs we can write this as:
+      % rhs * Rotate * (X4-X3) => [nx2] * [2x2] * [2x1] = [nx1]
+      lambda1 = (rhs(id_detA, :) * Rotate * (X4-X3))./detA(id_detA);
+
+      % The lower half is dependent on (X2-X1) which is a matrix of size [nx2]
+      % [-(X2(2)-X1(2)), X2(1)-X1(1)] / detA * rhs 
+      % As both (X2-X1) and rhs are matrices of size [nx2] there is no simple 
+      % matrix multiplication leading to a [nx1] vector. Therefore, use the
+      % elementwise multiplication and sum over it
+      % sum( [nx2] * [2x2] .* [nx2], 2) = sum([nx2],2) = [nx1] 
+      lambda2 = sum(-dataDiff(id_detA,:) * Rotate .* rhs(id_detA,:), 2)./detA(id_detA);
+
+      % Check whether lambda is in bound
+      out(id_detA) = all([lambda1, lambda2] > 0.0, 2) & all([lambda1, lambda2] < 1.0, 2);
+  end
 end
 % =========================================================================
 function simplifyLine(meta, handle, targetResolution)
--- a/src/cleanfigure.m
+++ b/src/cleanfigure.m
@@ -335,10 +335,13 @@ function out = segmentsIntersect(X1, diffX2X1, X3, X4)
   %   y1 + lambda1 (y2-y1)  =  y3 + lambda2 (y4-y3)
   %
   % for lambda1 and lambda2.
-
-  % NOTE: We could vectorize this function. Now X1 is a matrix containing
-  % all points X1 and diffX2X1 is a matrix containing the differences
-  % X2-X1 for all points X1 and X2
+  % See https://en.wikipedia.org/wiki/Line-line_intersection for reference.
+  
+  % NOTE: As X2 is only used in the difference (X2-X1), we directly pass
+  % that difference instead of X2 individually.
+  % Now X1 is a matrix containing all data points X1 and diffX2X1 is a
+  % matrix containing the differences X2-X1 between all consecutive data
+  % points X1 and X2
   % n is the number of segments (not points in the plot!)
   n   = size(diffX2X1, 1);
   out = false(n, 1);
--- a/src/cleanfigure.m
+++ b/src/cleanfigure.m
@@ -296,6 +296,7 @@ function out = segmentVisible(data, dataIsInBox, xLim, yLim)
     % in p and determine whether the line between the pair crosses the box.
     n = size(data, 1);
     out = false(n-1, 1);
+    
     % Only check if there is more than 1 point    
     if n>1
         % One of the neighbors is inside the box and the other is finite
@@ -305,16 +306,17 @@ function out = segmentVisible(data, dataIsInBox, xLim, yLim)
         % Get the corner coordinates
         [bottomLeft, topLeft, bottomRight, topRight] = corners(xLim, yLim);
 
-        % Calculate the difference between two consequtive points
+        % Calculate the difference between two consecutive points
         dataDiff = diff(data);
 
+        % Check if data points intersect with the borders of the plot
         left   = segmentsIntersect(data, dataDiff, bottomLeft , topLeft);
         right  = segmentsIntersect(data, dataDiff, bottomRight, topRight);
         bottom = segmentsIntersect(data, dataDiff, bottomLeft , bottomRight);
         top    = segmentsIntersect(data, dataDiff, topLeft    , topRight);
 
         % Check the result
-        out = any([thisVisible, nextVisible, left, right, top, bottom],2);
+        out = thisVisible | nextVisible | left | right | top | bottom;
     end
 end
 % =========================================================================
@@ -334,19 +336,21 @@ function out = segmentsIntersect(data, dataDiff, X3, X4)
   %
   % for lambda1 and lambda2.
 
-  %TODO: why don't we use `\` instead of Cramer's rule?
-
+  % NOTE: We could vectorize this function. Now data is a matrix containing
+  % all points X1 and dataDiff is a matrix containing the differences
+  % X2-X1 for all points X1 and X2
   n   = size(dataDiff,1);
   out = false(n,1);
-
-  % Rotational matrix with sign flip 
-  Rotate    = [0, -1; 1, 0 ];   
+  
+  % Rotational matrix with sign flip. It transforms a given vector [a,b] by 
+  % Rotate * [a,b] = [-b,a] as required for calculation of invA and detA  
+  Rotate = [0, -1; 1, 0 ];   
   
   % Calculate the determinant of A = [X2-X1, -(X4-X3)];
   % detA = -(X2(1)-X1(1))*(X4(2)-X3(2)) + (X2(2)-X1(2))*(X4(1)-X3(1))
   % NOTE: Vectorized this is equivalent to the matrix multiplication
   % [nx2] * [2x2] * [2x1] = [nx1]
-  detA   = dataDiff * Rotate * (X4-X3);  
+  detA = dataDiff * Rotate * (X4-X3);  
   
   % Get the indexes for nonzero elements
   id_detA = detA~=0;
@@ -381,7 +385,7 @@ function out = segmentsIntersect(data, dataDiff, X3, X4)
       lambda2 = sum(-dataDiff(id_detA,:) * Rotate .* rhs(id_detA,:), 2)./detA(id_detA);
 
       % Check whether lambda is in bound
-      out(id_detA) = all([lambda1, lambda2] > 0.0, 2) & all([lambda1, lambda2] < 1.0, 2);
+      out(id_detA) = lambda1 > 0.0 & lambda1 < 0.0 & lambda2 > 0.0 & lambda2 < 1.0;
   end
 end
 % =========================================================================
--- a/src/cleanfigure.m
+++ b/src/cleanfigure.m
@@ -300,8 +300,8 @@ function out = segmentVisible(data, dataIsInBox, xLim, yLim)
     % Only check if there is more than 1 point    
     if n>1
         % One of the neighbors is inside the box and the other is finite
-        nextVisible = (dataIsInBox(2:end)   & all(isfinite(data(1:end-1)),2));
-        thisVisible = (dataIsInBox(1:end-1) & all(isfinite(data(2:end)),2));
+        nextVisible = (dataIsInBox(2:end)   & all(isfinite(data(1:end-1)), 2));
+        thisVisible = (dataIsInBox(1:end-1) & all(isfinite(data(2:end)), 2));
 
         % Get the corner coordinates
         [bottomLeft, topLeft, bottomRight, topRight] = corners(xLim, yLim);
@@ -339,12 +339,13 @@ function out = segmentsIntersect(data, dataDiff, X3, X4)
   % NOTE: We could vectorize this function. Now data is a matrix containing
   % all points X1 and dataDiff is a matrix containing the differences
   % X2-X1 for all points X1 and X2
-  n   = size(dataDiff,1);
-  out = false(n,1);
+  % n is the number of segments (not points in the plot!)
+  n   = size(dataDiff, 1);
+  out = false(n, 1);
   
   % Rotational matrix with sign flip. It transforms a given vector [a,b] by 
   % Rotate * [a,b] = [-b,a] as required for calculation of invA and detA  
-  Rotate = [0, -1; 1, 0 ];   
+  Rotate = [0, -1; 1, 0];   
   
   % Calculate the determinant of A = [X2-X1, -(X4-X3)];
   % detA = -(X2(1)-X1(1))*(X4(2)-X3(2)) + (X2(2)-X1(2))*(X4(1)-X3(1))
@@ -359,7 +360,7 @@ function out = segmentsIntersect(data, dataDiff, X3, X4)
       % rhs = X3(:) - X1(:)
       % NOTE: Originaly this was a [2x1] vector. However as we vectorize the 
       % calculation it is beneficial to treat it as an [nx2] matrix rather than a [2xn]
-      rhs = bsxfun(@minus, X3', data(1:end-1,:));
+      rhs = bsxfun(@minus, X3', data(1:end-1, :));
       
       % Calculate the inverse of A and lambda
       % invA=[-(X4(2)-X3(2)), X4(1)-X3(1);...
@@ -382,10 +383,10 @@ function out = segmentsIntersect(data, dataDiff, X3, X4)
       % matrix multiplication leading to a [nx1] vector. Therefore, use the
       % elementwise multiplication and sum over it
       % sum( [nx2] * [2x2] .* [nx2], 2) = sum([nx2],2) = [nx1] 
-      lambda2 = sum(-dataDiff(id_detA,:) * Rotate .* rhs(id_detA,:), 2)./detA(id_detA);
+      lambda2 = sum(-dataDiff(id_detA, :) * Rotate .* rhs(id_detA, :), 2)./detA(id_detA);
 
       % Check whether lambda is in bound
-      out(id_detA) = lambda1 > 0.0 & lambda1 < 0.0 & lambda2 > 0.0 & lambda2 < 1.0;
+      out(id_detA) = 0.0 < lambda1 & lambda1 < 1.0 & 0.0 < lambda2 & lambda2 < 1.0;
   end
 end
 % =========================================================================
--- a/src/cleanfigure.m
+++ b/src/cleanfigure.m
@@ -306,21 +306,22 @@ function out = segmentVisible(data, dataIsInBox, xLim, yLim)
         % Get the corner coordinates
         [bottomLeft, topLeft, bottomRight, topRight] = corners(xLim, yLim);
 
-        % Calculate the difference between two consecutive points
-        dataDiff = diff(data);
+        % Define the vectors of data points
+        X1 = data(1:end-1,:);
+        X2 = data(2:end,:);
 
         % Check if data points intersect with the borders of the plot
-        left   = segmentsIntersect(data, dataDiff, bottomLeft , topLeft);
-        right  = segmentsIntersect(data, dataDiff, bottomRight, topRight);
-        bottom = segmentsIntersect(data, dataDiff, bottomLeft , bottomRight);
-        top    = segmentsIntersect(data, dataDiff, topLeft    , topRight);
+        left   = segmentsIntersect(X1, X2, bottomLeft , topLeft);
+        right  = segmentsIntersect(X1, X2, bottomRight, topRight);
+        bottom = segmentsIntersect(X1, X2, bottomLeft , bottomRight);
+        top    = segmentsIntersect(X1, X2, topLeft    , topRight);
 
         % Check the result
         out = thisVisible | nextVisible | left | right | top | bottom;
     end
 end
 % =========================================================================
-function out = segmentsIntersect(X1, diffX2X1, X3, X4)
+function out = segmentsIntersect(X1, X2, X3, X4)
   % Checks whether the segments X1--X2 and X3--X4 intersect.
   % Given four points X_k=(x_k,y_k), k\in{1,2,3,4}, and the two lines defined
   % by those,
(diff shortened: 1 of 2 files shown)


src/cleanfigure.m
```diff
@@ -296,36 +296,93 @@ function pruneOutsideBox(meta, handle)
...
+        out = any([thisVisible, nextVisible, left, right, top, bottom],2);
     end
 end
 % =========================================================================
-function out = segmentsIntersect(X1, X2, X3, X4)
+function out = segmentsIntersect(data, dataDiff, X3, X4)
   % Checks whether the segments X1--X2 and X3--X4 intersect.
-  lambda = crossLines(X1, X2, X3, X4);
-  out = all(lambda > 0.0) && all(lambda < 1.0);
+  % Given four points X_k=(x_k,y_k), k\in{1,2,3,4}, and the two lines defined
```
egeerardyn: You should adapt the comment here as well.
Have you checked that this provides a speedup?
Because passing the data and derivative, might be computationally advantageous, but it seems to introduce quite a bit of extra calculations (and hence complexity and possibility of bugs) here.

Before, the function was really simple and used functionality that was already present in `crossLines` instead of trying to recompute everything in a different way.

src/cleanfigure.m
```diff
@@ -296,36 +296,93 @@ function pruneOutsideBox(meta, handle)
...
+  %
+  % for lambda1 and lambda2.
+
+  %TODO: why don't we use `\` instead of Cramer's rule?
+
+  n   = size(dataDiff,1);
+  out = false(n,1);
+
+  % Rotational matrix with sign flip 
+  Rotate    = [0, -1; 1, 0 ];
```
egeerardyn: Where does this 180° rotation come from?

src/cleanfigure.m
```diff
@@ -296,36 +296,93 @@ function pruneOutsideBox(meta, handle)
...
+  % for lambda1 and lambda2.
+
+  %TODO: why don't we use `\` instead of Cramer's rule?
+
+  n   = size(dataDiff,1);
+  out = false(n,1);
+
+  % Rotational matrix with sign flip 
+  Rotate    = [0, -1; 1, 0 ];   
+
```
miscco: Should i add something along the line of

% Rotate * [a;b] = [-b, a] 

here?

src/cleanfigure.m
miscco: Hm, 
1. the comment is still correct, as it checks for intersection between those points, but for multiple in parallel
2. The computation is identical to the previous one.
3. As in my initial statement, for my usecase (figures for a journal paper) I get a speedup by a factor of 4

miscco: As a sidenote, the computation of diff is exactly not cheap if there are a large amount of points

src/cleanfigure.m
egeerardyn: I would certainly move it further towards the bottom, i.e. where you actually define the vectorized determinant.

miscco:
Yeah i will move it to Line  364 below the definition of invA.

As I am new, should I amend the last commit or make a new one?

src/cleanfigure.m
egeerardyn: With the last, do you mean that vectorized operation (i.e. `diff`) is faster than non-vectorized `minus` (that's my intuition, and typical old-style MATLAB rule-of-thumb)? Or exactly the opposite?

miscco: Two things:
1. diff is faster than for i=1:n X(n+1)-X(n)
2. We are using diff in every of the four invocations of segmentsIntersect. So precalculating is definitely beneficial

src/cleanfigure.m
miscco: Ah, i cant move it as it is used in the next line for detA

src/cleanfigure.m
miscco: I have added the following note below the block with the formulas:

  % NOTE: We could vectorize this function. Now data is a matrix containing
  % all points X1 and dataDiff is a matrix containing the differences
  % X2-X1 for all points X1 and X2

egeerardyn:
Please make a new commit in the same branch. When you push it, GitHub will put it here as well.

Amending, squashing, rebasing, are all nice features of `git` (except for `squash`, I regularly use them on my private branches), but they are quite advanced features of `git` that should not be used lightly on public code since they change the history and can break other people's repositories in nasty ways (I have done so before, it was a mess).

egeerardyn:
With regard to the use of `n`. I'm fine with how it currently is: you define `n` within the function to mean the number of segments (not the number of data points in the original data). If you want, you can always add a short comment, in the style of
```matlab
n = size(ddata, 1); % N is the number of segments (not points in the plot!)
```

okomarov:
@misco Can you please provide one of your .fig files if not confidential, or a snippet code that reproduces the speedup?

miscco:
Strange earlier the test with sin didnt work here what helps:
time = linspace(0, 1, 1E5);
x    = rand(1E5, 1);

figure(1)
clf

plot(time, x)
tic
cleanfigure
toc

I think it is because there is more heavy lifting required because many points overlap

egeerardyn:
I have tweaked our test suite a bit to return more fine-grained timings.
I've run this cherry-picked on top of `develop` (before) and this PR (after), I will provide a PR later on when I have my processing script a bit cleaned up.

These are graphical results in R2014b/Mac:

screen shot 2015-08-20 at 11 13 21

So there clearly is quite some improvement for some of the tests.

okomarov:
The spikes might be outliers that are not representative of the real runtime, but could be queue times.

A correct approach would be to timeit(), i.e. take the median of a few runs in the same test.



Sent from my iPhone

On 20 Aug 2015, at 10:16, "Egon Geerardyn" <notifications@github.com<mailto:notifications@github.com>> wrote:


I have tweaked our test suite a bit to return more fine-grained timings.
I've run this cherry-picked on top of develop (before) and this PR (after), I will provide a PR later on when I have my processing script a bit cleaned up.

These are graphical results in R2014b/Mac:

[screen shot 2015-08-20 at 11 13 21]<https://cloud.githubusercontent.com/assets/177360/9379884/a8519836-472c-11e5-8fb8-e361dd91d745.png>

So there clearly is quite some improvement for some of the tests.

miscco:
I have extended the previous approach to cover 3D data as well by checking the other 2 planes accordingly. The results are as follows:

# MATLAB 8.5 (Linux 3.13.0-62-generic) commit 163c8fb75f 

## Unreliable tests
These do not cause the build to fail.

|   Testcase |                   Name |                       OK |                                                                      Status |
| :--------- | :--------------------- | :----------------------: | :-------------------------------------------------------------------------- |
|  `ACID(1)` |     `multiline_labels` | :heavy_exclamation_mark: | hash 21a208748863224355861184cf9910f7 != (01d4eeae27254d91a0f8316897e23e81) |
|  `ACID(6)` | `sine_with_annotation` | :heavy_exclamation_mark: | hash 995cd2a901446c59f85552d48025ba27 != (7d415c014eec35c249802eb3a3ee519e) |
| `ACID(10)` |       `peaks_contourf` | :heavy_exclamation_mark: | hash 1241fa010c4c53ab967f1e61931b596f != (77bfa3a9f07e49b9061d919551ad0a0f) |
| `ACID(16)` |              `logplot` |       :white_check_mark: |                                                                             |
| `ACID(18)` |           `legendplot` | :heavy_exclamation_mark: | hash b15161cf25f700c68c15aeadb9b97d78 != (746423d1c0d403421667221a6b0528bf) |
| `ACID(24)` |          `quiver3plot` |       :white_check_mark: |                                                                             |
| `ACID(34)` |                 `bars` |       :white_check_mark: |                                                                             |
| `ACID(38)` |          `subplot2x2b` |       :white_check_mark: |                                                                             |
| `ACID(40)` |        `subplotCustom` |       :white_check_mark: |                                                                             |
| `ACID(42)` |            `bodeplots` | :heavy_exclamation_mark: | hash ca7b2029a08a2a311a94b73339e2128a != (8f6cafc1771cc3a506c8fc46c15ec43e) |
| `ACID(43)` |           `rlocusPlot` | :heavy_exclamation_mark: | hash 20e335e1c570ea7181e6fce6031b827c != (f372b9a4004fa45ac96a21014656e143) |
| `ACID(48)` |          `zplanePlot2` | :heavy_exclamation_mark: | hash 4c2f1fea92ce3c969475e3f6b40f87cf != (fb3ba1a4db46c9975cd00e6da99ebe48) |
| `ACID(58)` |            `surfPlot2` | :heavy_exclamation_mark: | hash 520bba1c7a536efb15c82dfcc8ce6a23 != (eb8e6d18db994ba4a90e177bd44b5db3) |
| `ACID(94)` | `stackedBarsWithOther` |       :white_check_mark: |                                                                             |
| `ACID(97)` |     `overlappingPlots` |       :white_check_mark: |                                                                             |

## Reliable tests
Only the reliable tests determine the build outcome.
Passing tests are not shown (only failed and skipped tests).

|   Testcase |                 Name |                       OK |                                                                      Status |
| :--------- | :------------------- | :----------------------: | :-------------------------------------------------------------------------- |
|  `ACID(7)` |  `linesWithOutliers` | :heavy_exclamation_mark: | hash 0ed33b1ff2c679913b2d039b005794ae != (4976c3f7b1e8e8cd32f7e17e34fcc99f) |
|  `ACID(8)` |      `peaks_contour` | :heavy_exclamation_mark: | hash 4c0f04124b13d5c12351f4875455a67d != (8885c9a4185aaca594ce5d0139aa578b) |
|  `ACID(9)` |       `contourPenny` | :heavy_exclamation_mark: | hash 1bbca20d0029e2e510eb5d7c46222ee7 != (3ee200ec93371666499d72f790ee9bb3) |
| `ACID(11)` | `many_random_points` | :heavy_exclamation_mark: | hash 6f676c75bc510bfa7b29b3c0332bb23f != (cb80dfbba6c43bbe4aeaaba5b9a52096) |
| `ACID(12)` |    `double_colorbar` | :heavy_exclamation_mark: | hash 44a0d0b4b35ed634d2f2f568e51f2c8a != (6355484ba42b84205cd6a7d6de76711d) |
| `ACID(14)` |        `double_axes` | :heavy_exclamation_mark: | hash 4c7a6f044691633fa3692df6e7013b8f != (331441239c1bb077aebaf5ce8abe55f1) |
| `ACID(15)` |       `double_axes2` | :heavy_exclamation_mark: | hash ec8ccb7d178fd3e0227fd2cf3b210fa0 != (739edbf88cbf57187bea11ef93ebbf9a) |
| `ACID(21)` |               `zoom` | :heavy_exclamation_mark: | hash 61f760a15c30e73eb656d462c9e53daf != (817b9999d832dc238b89573037cabc81) |
| `ACID(23)` |         `quiverplot` | :heavy_exclamation_mark: | hash 6f6723cafaf167b8e13d62f56b5a0aff != (308f4ce197791f2fd5862768e1157ca8) |
| `ACID(62)` |            `spectro` | :heavy_exclamation_mark: | hash e871a0b852a68fe12d78eead39e3de3b != (4426bb06bb44d08d016595f585fb01da) |
| `ACID(71)` |      `parameterSurf` | :heavy_exclamation_mark: | hash e64030978c54cc260847bf78a8c6ea6a != (12522a58d7e3d8dc7f4e38a0970bf047) |
| `ACID(17)` |    `colorbarLogplot` |          :grey_question: |                                                                     SKIPPED |

## Test summary
Test results for m2t commit 163c8fb75f running with MATLAB 8.5 on Linux
 3.13.0-62-generic.

```matlab
suite = @ACID;
alltests = [1:99];
reliable = [2:5 7:9 11:15 17 19:23 25:33 35:37 39 41 44:47 49:57 59:93 95:96 98:99];
unreliable = [1 6 10 16 18 24 34 38 40 42:43 48 58 94 97];
failReliable = [7:9 11:12 14:15 21 23 62 71];
passUnreliable = [16 24 34 38 40 94 97];
skipped = [17];
```

| Unreliable |    7 |    8 |    0 |    15 |
|   Reliable |   72 |   11 |    1 |    84 |
|      Total |   79 |   19 |    1 |    99 |

Build fails with 11 errors. :heavy_exclamation_mark:

okomarov:
@misco I see the 4x speedup now :+1:

egeerardyn:
@okomarov I completely agree with you that a proper timing analysis should also take uncertainty into account (e.g. use the median over a few runs to compare results).
As a quick-and-dirty doublecheck, I have re-executed both commits, and it gives fairly similar results (the large speedup spike is still present, although the runtime is quite a bit lower).

screen shot 2015-08-20 at 11 39 17

When I file my PR for that, I will make sure that provisions are in place to do some statistics on the individual runtimes to make sure we minimize the risk of seeing outliers.

egeerardyn:
From a readability point of view, I agree with @okomarov: the calls `(X1, X2, X3, X4)` is more straightforward and the same for computing `detA` over the vectorized counterpart (however, I would prefer not going through `det`, but rather solve the system of equations instead, but that's because I'm more familiar with the latter and those are quite close to the metal).

Anyway, for me it's fine if there is a speedup and the code is somewhat readable and documented (which, I think, it is by now). I've rerun the timing here (ten times) on R2014b and the improvements on our test suite are less than my other graphs:
screen shot 2015-08-20 at 20 33 05

Best speedup is about 2.5 (which is not that bad, actually, but not ten-fold as I saw before).

In any case: first priority now is to get the test suite passing as @okomarov said. Adding 3D support and other stuff is NOT in the scope of this PR and so we should put that off until we get the suite passing again.

Before this PR, everything passes without a problem. During my own timing runs (on top of this 339917b), I do get errors:

## Reliable tests
Only the reliable tests determine the build outcome.
Passing tests are not shown (only failed and skipped tests).

|   Testcase |                 Name |                       OK |                                                                      Status |
| :--------- | :------------------- | :----------------------: | :-------------------------------------------------------------------------- |
|  `ACID(7)` |  `linesWithOutliers` | :heavy_exclamation_mark: | hash 0ed33b1ff2c679913b2d039b005794ae != (4976c3f7b1e8e8cd32f7e17e34fcc99f) |
| `ACID(11)` | `many_random_points` | :heavy_exclamation_mark: | hash 6f676c75bc510bfa7b29b3c0332bb23f != (cb80dfbba6c43bbe4aeaaba5b9a52096) |
| `ACID(14)` |        `double_axes` | :heavy_exclamation_mark: | hash ffa4bc2007efb67103c8c672333940af != (331441239c1bb077aebaf5ce8abe55f1) |
| `ACID(21)` |               `zoom` | :heavy_exclamation_mark: | hash 61f760a15c30e73eb656d462c9e53daf != (817b9999d832dc238b89573037cabc81) |
| `ACID(17)` |    `colorbarLogplot` |          :grey_question: |                                                                     SKIPPED |

## Test summary
Test results for m2t commit 66bf74c73c running with MATLAB 8.4 on Mac OS 10.10.4.

```matlab
suite = @ACID;
alltests = [1:99];
reliable = [2:5 7:9 11:15 17 19:23 25:33 35:37 39 41 44:47 49:57 59:93 95:96 98:99];
unreliable = [1 6 10 16 18 24 34 38 40 42:43 48 58 94 97];
failReliable = [7 11 14 21];
passUnreliable = [];
skipped = [17];
```

| Unreliable |    0 |   15 |    0 |    15 |
|   Reliable |   79 |    4 |    1 |    84 |
|      Total |   79 |   19 |    1 |    99 |

Build fails with 4 errors. :heavy_exclamation_mark:

So we will have to compare `[7 11 14 21]` graphically, before and after this PR.

okomarov:
As bread for the mind, there is a clever way to test for intersection, that does not involve even setting up a system of equations, but it check whether line segments are in counter-clockwuse order (which would also solve the #723:

http://bryceboe.com/2006/10/23/line-segment-intersection-algorithm/

which is a specific case of http://www.geeksforgeeks.org/check-if-two-given-line-segments-intersect/ (but we do not care about collinearity)

miscco:
There is also a very related way using the inner product
http://martin-thoma.com/how-to-check-if-two-line-segments-intersect/

As a mental note, I believe we can actually ignore most of the computational load by levering the "or" in segmentVisible.

Given that in most cases the points will be in the box, we can pass only those line segments that have a point outside the box. But that will be another commit. I am on to testing the failing cases.

miscco:
So i personally cannot find any differences between resulting figures. Here is the test is used. It assumes that current git cleanfigure is given as cleanfigure_orig. The process is as follows:
1. Create normalized figure.
2. ACID test N
3. cleanfigure (orig or new)
4. Save as png

5. Compare resulting pngs pixel wise.

function [result] = ACID_compare(N)

h1 = figure('units','normalized','outerposition',[0 0 1 1]);
ACID(N);
cleanfigure_orig; 
print(h1, ['ACID',num2str(N),'_orig'], '-dpng');

h2 = figure('units','normalized','outerposition',[0 0 1 1]);
ACID(N); 
cleanfigure; 
print(h2, ['ACID',num2str(N),'_new'], '-dpng');
close all;

Image_orig= imread(['ACID',num2str(N),'_orig.png']);
Image_new = imread(['ACID',num2str(N),'_new.png']);

if(~isempty(find(Image_orig~=Image_new, 1)))
    figure(3);
    image(Image_orig+Image_new);
    result = false;
    disp(['Test ACID(',num2str(N),') failed']);
else
    result = true;    
    disp(['Test ACID(',num2str(N),') passed']);
end
    system(['rm ','ACID',num2str(N),'_orig.png ACID',num2str(N),'_new.png']);
end

okomarov:
We should have probably mentioned that we have testHeadless and testGraphical for the purpose.

Sent from my iPhone

On 21 Aug 2015, at 08:46, "Michael Schellenberger Costa" <notifications@github.com<mailto:notifications@github.com>> wrote:


So i personally cannot find any differences between resulting figures. Here is the test is used. It assumes that current git cleanfigure is given as cleanfigure_orig. The process is as follows:
1. Create normalized figure.
2. ACID test N
3. cleanfigure (orig or new)
4. Save as png

  1.  Compare resulting pngs pixel wise.

function [result] = ACID_compare(N)

h1 = figure('units','normalized','outerposition',[0 0 1 1]);
ACID(N);
cleanfigure_orig;
print(h1, ['ACID',num2str(N),'_orig'], '-dpng');

h2 = figure('units','normalized','outerposition',[0 0 1 1]);
ACID(N);
cleanfigure;
print(h2, ['ACID',num2str(N),'_new'], '-dpng');
close all;

Image_orig= imread(['ACID',num2str(N),'_orig.png']);
Image_new = imread(['ACID',num2str(N),'_new.png']);

if(~isempty(find(Image_orig~=Image_new, 1)))
figure(3);
image(Image_orig+Image_new);
result = false;
disp(['Test ACID(',num2str(N),') failed']);
else
result = true;

disp(['Test ACID(',num2str(N),') passed']);
end
system(['rm ','ACID',num2str(N),'_orig.png ACID',num2str(N),'_new.png']);
end

miscco:
Yes and those test give a false positive. Egon reported that on his machine the tests 7, 11, 14 and 21 fail.

However, if one plots them with the original cleanfigure and my git version they are pixel identical. That was the purpose of the test function. Not comparing the hash but rather the resulting figure.

egeerardyn:
@okomarov @miscco We have `testHeadless` and `testGraphical` for that, but we still don't have convenient tools to track regressions through `git`. Until now, I typically track those by 
 1. running `testHeadless`/`testGraphical` using the "before" commit
 2. compile everything, 
 3. move the files to a temp location
 4. checkout "after" commit
 5. repeat steps 1 and 2
 6. compare the files

In essence, that is equivalent to what @miscco is doing, so the comparison is fine by me :+1:

miscco:
Yes so in principle one could do some git checkout shenanigans like
git checkout develop
git reset --hard

git checkout MyFeature
git reset --hard

Therewith one could always use cleanfigure and would not have to manually move stuff

miscco:
Btw if someone could verify that the images doesnt change it would be greatly appreciated

egeerardyn:
@miscco I will have a got at it tonight and provide you with hashes if everything goes well.

okomarov:
A comparison of ef6a28f64464a340e01c6fdb0eb1ee10375ac02d (left) vs current develop 892a4ec100ea67b0d6e220627c91c3846bccf248 (right) shows that some NaNs spur (which do not affect the graphical appearance of the plot)

capture

miscco:
Ah yeah that one is one me,
that was on my last commit, to simplify cutting. I though the original code included NaNs for every omitted data point. Obviously I was wrong I can change that easily;

okomarov:
I'm did not include those commits

Sent from my iPhone

On 21 Aug 2015, at 12:06, "Michael Schellenberger Costa" <notifications@github.com<mailto:notifications@github.com>> wrote:


Ah yeah that one is one me,
that was on my last commit, to simplify cutting. I though the original code included NaNs for every omitted data point. Obviously I was wrong I can change that easily;

miscco:
Oh strange, I will fix it anyway

okomarov:
Jut git revert those commits.
http://git-scm.com/docs/git-revert
Sent from my iPhone

On 21 Aug 2015, at 12:16, "Michael Schellenberger Costa" <notifications@github.com<mailto:notifications@github.com>> wrote:


Oh strange, I will fix it anyway

miscco:
Hm reverting intertwined commits is hard. I should learn to properly start branches everytime i start a new idea.

egeerardyn:
@miscco Maybe http://stackoverflow.com/questions/20229739/git-revert-several-specific-commits helps? I think especially the tip regarding the order might be helpful: use
```shell
git revert C B A
```
when you want to revert the commits A,B,C (with C the newest commit, A the oldest).

okomarov:
As also @egeerardyn pointed out 

> Adding 3D support and other stuff is NOT in the scope of this PR and so we should put that off until we get the suite passing again.

I consider this PR up to ef6a28f64464a340e01c6fdb0eb1ee10375ac02d only. Adding up commits on top of that is only making it harder to assess where the NaN problem is coming from and hence to merge the speedup feature. 

All additional reworkings/cleanups/features can come afterwards.

miscco:
Actually my bigger problem is github, as I havent figured out how to reset to an earlier commit and push it back up.

okomarov:
Unfortunately, I am not a git expert. I use the platform SmartGit, and do most of the things there. 

In any case, I learnt two things: 1) no matter where you bang your head, the learning curve with git (and tools, hence also SmartGit) is not trivial , 2) Github is an extremely limited interface, and I advice instead to invest some time in either learning the bare git syntax (which I gave up at some point, but a good interactive tutorial is http://gitimmersion.com/) or a tool you are comfortable with (hard to say which one, I started with the simple git-gui + looking for answers on stackoverflow, and moved to smartgit).

egeerardyn:
@okomarov I have another impression with `git`. For most daily stuff, the learning curve does get a lot better after a while (but I'd say it takes about a year of sporadic use to become somewhat fluent). For the more advanced stuff, there remains quite a learning curve and I also still have to look up quite some stuff every now and then.

I agree that GitHub (website and desktop app) is a very limited interface and I like it a lot if one is just looking at the repository (in the worst case deleting a branch or even merging a PR). I also think Github is not written in a way that encourages to change PRs after the fact (it can be done, but mostly by manipulating the `git` repo directly instead of through GitHub).


I wholeheartedly recommend [Pro Git](https://git-scm.com/book/en/v2) for everybody or [Version Control By Example](http://ericsink.com/vcbe/) for people without any experience with version control. Both are free, by the way. For GUIs I can also recommend SmartGit (not free, but works on Linux, Mac and Windows) or SourceTree (free for Mac and Windows): for most tasks, I use the command line, but for very big results (e.g. log of many commits, blames, ...) I prefer to use SmartGit or the git integration in Sublime Text.

@miscco: I will try to have a look at it tonight, to see whether we can cut this branch a bit more cleanly.

miscco:
Yeah, git needs some time, but like LaTeX when you get the hang of it, its amazing. I cant even tell how often it has saved me.  In my case I am familiar with working isolated on my workstation and doing all the shenanigans there, but I barely have to worry about conflicts with servers and other coworkers..

miscco:
Seems that i can reroll github with git push -f. Somhow didnt thought about that.

I will separate the changes on monday, as i do not have access at home. There is one thing i would like to understand till then. Basically the whole thing is a rewrite of crossLine. So later in the code it is used again in the moveToBox function.

There Lambda is required. What i do not understand is, why we always only take lambda(1) for the new alpha?

Anyway i planning on going back to the original structure with isSegmentIntersect as a wrapper around crossLines, that will return alpha that we then turn into a boolean. Therewith we can use the new approach everywhere.
Thoughts?

okomarov:
Speaking for myself, I can only tell that the current structure reflects and builds on existing design, and often/always, it is convenient to follow the "If it ain't broke, don't fix it" law and simply build on top. 

This is to say that, I do not have a detailed picture of what's going on and those cleanfigure features were probably added by different contributors whom also had a partial vision and just ensured that the working legacy structure stayed untouched :smile:. 

If you are willing/daring to improve the codebase, it is gonna cost time to understand the whole design, to re-test everything, to find what went wrong ...etc..etc.. and I think you might be starting to get what I mean. This amount of time is usually unsustainable in comparison to the implementation of the feature WITHIN the legacy design. 

In brief, my take on this refactoring would have been: 
* vectorize crossLines, i.e. use X1(:,1) instead of X1(1), etc..
* check if speedup is satisfactory (recall my comment on caching and that passing a diffed argument instead of diffing inside the function should NOT make a difference)
* seek other speedups, possibly prefering clarity over speed. 

This last point is very important. It is true that the user only sees speed, however, if it takes the **next** developer a whole day (or more) to understand what's going on, next time he needs to look at that code and add a feature/fix a bug, he won't do it :sweat_smile:, and no gain for the user either.

Finally, are you uncomfortable with reverting? Why force push (which has drawbacks)?

miscco:
I did revert, but i have to force push it, or github wont accept the push, as the remote repository has a different path.

okomarov:
You should have not problems if you checked out at 844f279c25e43e8449597bd56628d52e44832c82 and reverted in this order:
```
git revert 844f279c25e43e8449597bd56628d52e44832c82
git revert --mainline 1 cdcd7c6f75d8e8927a4d535a9c34d534e1ddaa5f
git revert 001a36a4709335ab75595f35263a4eac33435fad
```

miscco:
So I have reset back to  	"small cleanups to better preserve the original formulas."
001a36a

Therefore only the relevant changes should be in this PR.

miscco:
So I thought about the inputs for segmentsIntersect()

My problem with segmentsIntersect(X1, X2, X3, X4) is, that there is no nice way of handling the data. The less awkward approach would be to pass X1 = data(1:end-1,:) X2 = data(2:end,:) and then do diffX2X1 = X2-X1, which might be slower and take more memory than needed.

I have to say I dont like it at all. compared to  segmentsIntersect(X1, diffX2X1, X3, X4). I mean anyone who actually looks at the code will understand that diffX2X1 is X2-X1. 

I have added a comment to ease that confusion.

egeerardyn:
I don't think we have to worry that much about speed right now (i.e. premature optimization is the root of all evil). I also, more than once, optimized some code without actually profiling (e.g. `persistent` variables instead of parsing files on disk over and over again) and there my intuition was completely wrong.

Step 1 should be to create working code that is easy to understand. Step 2 is optimization if runtime (memory/speed/...) is unsatisfactory, since optimized code is often more difficult to understand. 

In this case: `(X1, X2, X3, X4)` is easy to understand, `(X1, diffX2X1, X3, X4)` is already quite a bit harder (still no rocket science, though).

miscco:
Yes premature optimization is bad and code complexity too, but given that comment:

  % NOTE: As X2 is only used in the difference (X2-X1), we directly pass
  % that difference instead of X2 individually.

I am unsure, whether it is really more difficult to understand

okomarov:
Given `main1` that does the diff in the subfunction  and `main2` passes the diff to the subfunction:
```
data = randn(1e4,2);

timeit(@()main1(data)) % 0.000129284421538342
timeit(@()main2(data)) % 0.000101186384742638 
```

I do not see room for argument. Clarity of code is **preferred** for the same reason I tried to explain before. If the next developer (could be me and I have short memory), who doesn't know about line intersection formulas, wastes an additional 5min to figure out why the formulas are expressed with a diff argument instead of the 4 points that he finds on e.g. wikipedia, that's 60*5/(0.000129284421538342 - 0.000101186384742638) ~ 10676902 runs with the optimized function to break even. 

The functions are:
```
function out = main1(data)
x1 = data(2:end,1);
x2 = data(1:end-1,2);

out = crossLines(x1,x2,1,2);
out = crossLines(x1,x2,3,4);
out = crossLines(x1,x2,5,6);
out = crossLines(x1,x2,7,8);
end

function out = crossLines(x1,x2,x3,x4)
xd = x1-x2;
out = xd+x3+x4; 
end
```

```
function out = main2(data)
xd = data(2:end,1) - data(1:end-1,2);

out = crossLines(xd,1,2);
out = crossLines(xd,3,4);
out = crossLines(xd,5,6);
out = crossLines(xd,7,8);
end
function out = crossLines(xd,x3,x4)
out = xd+x3+x4; 
end
```

miscco:
But this is an unfair  comparison.

1. The data set is rather small with 1E4. In any case i get an speedup by about 50%. Just with the 1E6 data set it is 5.5ms each iteration, which breaks even at 55k iterations. However, form particular use case, it cuts about of 25s for cleanfigure with each iteration. So about running it 5 times and i have won time

2. It doesnt take you 5 min longer than with X2-X1 in place. If you are new to the code you will have to check the equations in any case and doublecheck the code. When you see a dataDiff instead of X2-X1 finding the comment that tells you dataDiff=X2-X1 is about 30sec.

okomarov:
On average the plots will have much less than 1e4 points, also because the line simplification algo comes first. Yes, you get a 50% speedup that corresponds to ~10 millisecond speedup, i.e. percentages do not mean much without absolute values. 

What is your particular use case? Stress testing should be done with context.

With respect to 2 I disagree. When I look at some code, I decide if I **trust** it or not. I want to trust code because in that case I do not need to understand its internals. However, if I do not trust the code, then I need to check it, and this is when the time waste begins. The diffed argument is a point that raises that kind of mental warning. The method is basically not self-contained.

miscco:
Ok, i am definitely not happy with not passing the diff directly, but there it is. I have preserved all of the original formula, so that one can easily verify the code.

Additionally I made lambda = [lambda1, lambda2] a variable, to use the most part of segmentsIntersect in crossLine, when i a thinking about vectorizing movePointsCloser.

The same checks as usual fail in my MATLAB2015a. However, I can only find a shift by one pixel in Test 8 although not always.

miscco:
For reference:
# MATLAB 8.5 (Linux 3.13.0-62-generic) commit 704ebdbe80 

## Unreliable tests
These do not cause the build to fail.

|   Testcase |                   Name |                       OK |                                                                      Status |
| :--------- | :--------------------- | :----------------------: | :-------------------------------------------------------------------------- |
|  `ACID(1)` |     `multiline_labels` | :heavy_exclamation_mark: | hash 21a208748863224355861184cf9910f7 != (01d4eeae27254d91a0f8316897e23e81) |
|  `ACID(6)` | `sine_with_annotation` | :heavy_exclamation_mark: | hash 995cd2a901446c59f85552d48025ba27 != (7d415c014eec35c249802eb3a3ee519e) |
| `ACID(10)` |       `peaks_contourf` | :heavy_exclamation_mark: | hash 1241fa010c4c53ab967f1e61931b596f != (77bfa3a9f07e49b9061d919551ad0a0f) |
| `ACID(16)` |              `logplot` |       :white_check_mark: |                                                                             |
| `ACID(18)` |           `legendplot` | :heavy_exclamation_mark: | hash d37d19ba36d911a0db60095db48b08ef != (746423d1c0d403421667221a6b0528bf) |
| `ACID(24)` |          `quiver3plot` |       :white_check_mark: |                                                                             |
| `ACID(34)` |                 `bars` |       :white_check_mark: |                                                                             |
| `ACID(38)` |          `subplot2x2b` |       :white_check_mark: |                                                                             |
| `ACID(40)` |        `subplotCustom` |       :white_check_mark: |                                                                             |
| `ACID(42)` |            `bodeplots` | :heavy_exclamation_mark: | hash ca7b2029a08a2a311a94b73339e2128a != (8f6cafc1771cc3a506c8fc46c15ec43e) |
| `ACID(43)` |           `rlocusPlot` |       :white_check_mark: |                                                                             |
| `ACID(48)` |          `zplanePlot2` | :heavy_exclamation_mark: | hash 4c2f1fea92ce3c969475e3f6b40f87cf != (fb3ba1a4db46c9975cd00e6da99ebe48) |
| `ACID(58)` |            `surfPlot2` | :heavy_exclamation_mark: | hash 520bba1c7a536efb15c82dfcc8ce6a23 != (eb8e6d18db994ba4a90e177bd44b5db3) |
| `ACID(94)` | `stackedBarsWithOther` |       :white_check_mark: |                                                                             |
| `ACID(97)` |     `overlappingPlots` |       :white_check_mark: |                                                                             |

## Reliable tests
Only the reliable tests determine the build outcome.
Passing tests are not shown (only failed and skipped tests).

|   Testcase |              Name |                       OK |                                                                      Status |
| :--------- | :---------------- | :----------------------: | :-------------------------------------------------------------------------- |
|  `ACID(8)` |   `peaks_contour` | :heavy_exclamation_mark: | hash 4c0f04124b13d5c12351f4875455a67d != (8885c9a4185aaca594ce5d0139aa578b) |
|  `ACID(9)` |    `contourPenny` | :heavy_exclamation_mark: | hash 1bbca20d0029e2e510eb5d7c46222ee7 != (3ee200ec93371666499d72f790ee9bb3) |
| `ACID(12)` | `double_colorbar` | :heavy_exclamation_mark: | hash 44a0d0b4b35ed634d2f2f568e51f2c8a != (6355484ba42b84205cd6a7d6de76711d) |
| `ACID(14)` |     `double_axes` | :heavy_exclamation_mark: | hash a053ed16097389ed6d6cfe6f353a46ab != (331441239c1bb077aebaf5ce8abe55f1) |
| `ACID(15)` |    `double_axes2` | :heavy_exclamation_mark: | hash ec8ccb7d178fd3e0227fd2cf3b210fa0 != (739edbf88cbf57187bea11ef93ebbf9a) |
| `ACID(23)` |      `quiverplot` | :heavy_exclamation_mark: | hash 6f6723cafaf167b8e13d62f56b5a0aff != (308f4ce197791f2fd5862768e1157ca8) |
| `ACID(62)` |         `spectro` | :heavy_exclamation_mark: | hash e871a0b852a68fe12d78eead39e3de3b != (4426bb06bb44d08d016595f585fb01da) |
| `ACID(71)` |   `parameterSurf` | :heavy_exclamation_mark: | hash e64030978c54cc260847bf78a8c6ea6a != (12522a58d7e3d8dc7f4e38a0970bf047) |
| `ACID(17)` | `colorbarLogplot` |          :grey_question: |                                                                     SKIPPED |

## Test summary
Test results for m2t commit 704ebdbe80 running with MATLAB 8.5 on Linux
 3.13.0-62-generic.

```matlab
suite = @ACID;
alltests = [1:99];
reliable = [2:5 7:9 11:15 17 19:23 25:33 35:37 39 41 44:47 49:57 59:93 95:96 98:99];
unreliable = [1 6 10 16 18 24 34 38 40 42:43 48 58 94 97];
failReliable = [8:9 12 14:15 23 62 71];
passUnreliable = [16 24 34 38 40 43 94 97];
skipped = [17];
```

| Unreliable |    8 |    7 |    0 |    15 |
|   Reliable |   75 |    8 |    1 |    84 |
|      Total |   83 |   15 |    1 |    99 |

Build fails with 8 errors. :heavy_exclamation_mark:

miscco:

miscco:
So following Jenkins, the only test that fails is ACID[23] on Octave. However i do not have it installed. Can someone else check it?

egeerardyn:
Output TeX files are in  https://gist.github.com/egeerardyn/2180dc8899563ce8a9b0

Some points are removed, as you can see from diffing those files:
```diff
411d410
< -2.00078	-2.00096\\
414d412
< -2.00167	-1.80136\\
417d414
< -2.0033	-1.60235\\
420d416
< -2.00602	-1.40369\\
423d418
< -2.01012	-1.20525\\
426d420
< -2.01572	-1.00671\\
429d422
< -2.02253	-0.807621\\
432d424
< -2.02981	-0.607507\\
435d426
< -2.03641	-0.40608\\
438d428
< -2.04105	-0.203417\\
441d430
< -2.04272	   0\\
444d432
< -2.04105	0.203417\\
447d434
< -2.03641	0.40608\\
450d436
< -2.02981	0.607507\\
453d438
< -2.02253	0.807621\\
456d440
< -2.01572	1.00671\\
459d442
< -2.01012	1.20525\\
462d444
< -2.00602	1.40369\\
465d446
< -2.0033	1.60235\\
468d448
< -2.00167	1.80136\\
471d450
< -2.00078	2.00096\\
474d452
< -1.80101	-2.00185\\
534d511
< -1.80101	2.00185\\
537d513
< -1.60146	-2.00325\\
597d572
< -1.60146	2.00325\\
600d574
< -1.40186	-2.00519\\
660d633
< -1.40186	2.00519\\
663d635
< -1.20197	-2.00748\\
723d694
< -1.20197	2.00748\\
726d696
< -1.00159	-2.00968\\
786d755
< -1.00159	2.00968\\
789d757
< -0.800586	-2.0111\\
849d816
< -0.800586	2.0111\\
852d818
< -0.599064	-2.01101\\
912d877
< -0.599064	2.01101\\
915d879
< -0.397384	-2.00897\\
975d938
< -0.397384	2.00897\\
978d940
< -0.196062	-2.00505\\
1038d999
< -0.196062	2.00505\\
```

Before and after figures are below (you guess which is which :stuck_out_tongue: ).


before23
after23

Even diffing the image shows very little changes (i.e. half a pixel on the border or something). So if you change the MD5 hash in the right file, you can make the test pass as far as I'm concerned :+1:

miscco:
It is rather strange, because all the missing points should not have been plotted at all (at least one coordinate is larger than |2|)

So it seems there was a bug there in octave.

okomarov:
@miscco To elaborate on Eagon's sentence, please, change in `\test\suites\ACID.Octave.3.8.0.md5`

    quiverplot : 0cc7aef48dddc2c9075a079eaf580700

to 

    quiverplot : e662f099377f8458f3237bc5237ce0b0

so that we can merge it. Nice job. 

@egeerardyn Can we change the hash format into something more explicit, I never remember which one is the existing one and which is the new one. 

e.g. `new 1241fa010c4c53ab967f1e61931b596f, expected 77bfa3a9f07e49b9061d919551ad0a0f`

We also can rename the column name to Hash (instead of status), and drop the word `hash` in the description.

miscco:
Yeah, I will do that tomorrow. 

I asume I have to rebase now that #744 was merged.

okomarov:
No rush :smile: and also no need to rebase, there are no conflicts to solve. 

PS. It's easier on your end to edit that line than on ours. We would need to merge to an upstream branch (branch residing on the remote repo by `matlab2tikz`), push a commit to that, and merge into develop, drop the branch.

egeerardyn:
@okomarov I will change the report a bit to make clearer which hash is which. However, I won't touch the status column just yet: that's not only hash problems, but also plot failures and m2t crashes that may end up in that column.

PR #738, opened 2015-08-20
Timing the test suite

egeerardyn:
This is my processing script (see #737) for timing the test cases.
This is only the processing, you need to run the test suite yourself (repeatedly, if you want more reliable results).

I haven't fully checked whether there is a burn-in period of the timing results, since each the easies way to get the results, is to run the test suite repeatedly. So possibly, the caches get wiped by the time a test case is run again. If that's the case and a problem, we need to tweak the test framework a bit.

The output looks like this for #737 (after) vs. this branch (before):
screen shot 2015-08-20 at 20 33 05

### DIFF
--- a/test/compareTimings.m
+++ b/test/compareTimings.m
@@ -81,12 +81,19 @@ end
 % ------------------------------------------------------------------------------
 function [h] = plotByTestCase(timing, name)
     hold on;
-    h(1) = plot(timing.before, ...
-                'Color', colors.before, ...
-                'DisplayName', 'Before');
-    h(2) = plot(timing.after, ...
+    if size(timing.before, 2) > 1
+        h{3} = plot(timing.before, '.',...
+                    'Color', colors.before, 'HandleVisibility','off');
+        h{4} = plot(timing.after, '.',...
+                    'Color', colors.after, 'HandleVisibility','off');
+    end
+    h{1} = plot(median(timing.before, 2), '-',...
+                    'Color', colors.before, ...
+                    'DisplayName', 'Before');
+    h{2} = plot(median(timing.after, 2), '-',...
                 'Color', colors.after,...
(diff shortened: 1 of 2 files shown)


egeerardyn:
@matlab2tikzBot : please test this

egeerardyn:
@matlab2tikz/developers Could somebody review this? Otherwise I will merge it, since it's developer-only code anyway.

test/compareTimings.m
```diff
@@ -0,0 +1,252 @@
...
+% statusBefore = cell(N, R);
+% for r = 1:R
+%     statusBefore(:, r) = testHeadless;
+% end
+% % now check out the after commit
+% statusAfter = cell(N, R);
+% for r = 1:R
+%     statusAfter(:, r) = testHeadless;
+% end
+% compareTimings(statusBefore, statusAfter)
```
okomarov: I would reformat slightly the example and indent it to make it more visible, otherwise else is good:

% Example:
%     suite = @ACID;
%     N = numel(suite(0));
%     R = 10;
%     
%     % Record current timings
%     before = cell(N, R);
%     for r = 1:R
%         before(:, r) = testHeadless;
%     end
%     
%     % now check out the after commit
%     after = cell(N, R);
%     for r = 1:R
%         after(:, r) = testHeadless;
%     end
%         
%     compareTimings(before, after)

okomarov:
Apart that minor comment, good to go.

okomarov:
BTW, I like your color palette, I am gonna steal it for my graphs :sweat_smile:

egeerardyn:
@okomarov That's the tango color palette. I do have a matlab function for easy access to those colors. Only downside is that it's not very friendly for colorblind people. 

I'll tackle your comment in the morning, thank you for reviewing.

PR #746, opened 2015-08-24
Octave CI

egeerardyn:
Trial to make CI work again with Octave 3.8 instead of 4.0.


egeerardyn:
This is not going to work like that. `liboctave2` depends on `glibc6` of a different version due to the difference in Ubuntu version of Travis and the packages I was trying to load here.

PR #749, opened 2015-08-25
Jenkins

egeerardyn:
Some Jenkins-related stuff, such as a basic TAP report. Nothing fancy.


test/makeTapReport.m
```diff
@@ -0,0 +1,77 @@
...
+    % returns true if it is a valid (file) stream or stdout/stderr stream
+    bool = isnumeric(val) || ischar(val);
+end
+% ==============================================================================
+function printTAPVersion(stream)
+    % prints the TAP version
+    fprintf(stream,'TAP version 13\n');
+end
+function printTAPPlan(stream, statuses)
+    % prints the TAP version
```
PeterPablo: plan

PeterPablo:
Code looks good to me.

egeerardyn:
Thanks for reviewing. I will update this tonight (my VPN seems to have crashed and Jenkins is not completely satisfied yet).

egeerardyn:
@matlab2tikz/developers Please review this. Jenkins is running well on my spare laptop. This tests our three major environments: Octave 3.8, MATLAB 2014a and MATLAB 2014b.
Normally, the test results should be accessible online.

Next step after this PR, is fixing Octave 4 such that Travis also gets a green light (or temporarily disable Travis).

test/makeTapReport.m
```diff
@@ -0,0 +1,72 @@
...
+% MAKETAPREPORT(status, 'stream', FID, ...) changes the filestream to use
+%  to output the report to. (Default: 1 (stdout)).
+%
+% TAP Specification: https://testanything.org
+%
+% See also: testHeadless, makeTravisReport, makeLatexReport
+
+    %% Parse input arguments
+    SM = StreamMaker;
+    ipp = m2tInputParser;
```
miscco: For consistency you should also do StreamMaker()

test/makeTapReport.m
```diff
@@ -0,0 +1,72 @@
...
+
+    if hasTestFailed(status)
+        result = 'not ok';
+    else
+        result = 'ok';
+    end
+
+    directive = addDirective(status.skip, directive, '# SKIP skipped');
+    directive = addDirective(status.unreliable, directive, '# TODO unreliable');
+
```
miscco: This might be easier to understand if you use a if/elseif there.

egeerardyn: You mean adding the directives inline with something like `if status.skip, ... end` ?

I prefer the `addDirective` function for a few reasons:
 - it's just a ugly string concatenation
 - the logic for both directives is identical (i.e. duplicate code if not factored into a function)
 - the cyclomatic complexity of this function remains a bit lower (either solution (2 vs 4) is on the low-end, though)

I do agree that maybe I can make the intent of the `addDirective` function more clear. What if I'd change the name to `addDirectiveIf( <condition>, <directive>, <addition> )`? The `If` at the end looks a bit strange, but it covers the intent better. What do you think?

test/makeTapReport.m
egeerardyn: :+1:

test/makeTapReport.m
miscco: So my point was, that currently there are only 2 cases. If you plan on expanding that, a seperate function might be suitable. If it remains only two, then 

if status.skip 
   directive = strtrim([directive ' ' # SKIP skipped]);
elseif status.unreliable
   directive = strtrim([directive ' ' # TODO unreliable]);
end

In the other case i would stronglyprefer a more general approach like

addDirectives(status, directive); 

without the explicit condition in the function arguments.

test/private/StreamMaker.m
```diff
@@ -0,0 +1,83 @@
...
+    closeAfterUse = false;
+    if ischar(streamSpecifier)
+        Stream.name = streamSpecifier;
+        Stream.fid = fopen(Stream.name, varargin{:});
+        closeAfterUse = true;
+
+    elseif isnumeric(streamSpecifier)
+        Stream.fid = streamSpecifier;
+        if streamSpecifier == 1
+            Stream.name = 'stdout';
```
miscco: So i am unsure, if the else case with stream.name = streamSpecifier is possible.

What do you think about a switch statement
switch(streamSpecifier)
case 1:
...
case 2:
...
default:
 error('Stream:InvalidStream', ...
          'Unable to create stream named "%s"!', ...
            streamSpecifier);
end

Therewith all possible cases are handled.

test/suites/ACID.m
```diff
@@ -523,6 +523,7 @@
 function [stat] = logplot()
   stat.description = 'Test logscaled axes.';
   stat.unreliable = isMATLAB('>=', [8,4]); %FIXME: #590
+  stat.unreliable = stat.unreliable || isMATLAB('<=', [8,3]); %FIXME: #749 (Jenkins)
```
miscco: Does that mean this is always unreliable?
[>=8,4] || [<=8,3] 
In any case i would directly put that in one line

test/makeTapReport.m
egeerardyn: In TAP, there is currently only support for 2 directives:
 - SKIP (equivalent to our "skipped")
 - TODO (more general than our "unreliable", but the intent is quite similar)

Both are not mutually exclusive, if I recall correctly (so a test could be both TODO and SKIP, e.g. in our case this could happen when running an unreliable test on a machine that doesn't have all the right toolboxes).

I think I will implement your last suggestion: refactor that lot into a function with a bit of duplication.

test/suites/ACID.m
egeerardyn: Yes, indeed and aesthetically, I completely agree with your remark.

However, I kept those as separate lines, because that allows to distinguish the different causes of unreliability. That should allow us to trace those problems more directly.

miscco: How about:
stat.unreliable = isMATLAB('>=', [8,4])...#FIXME: 590
                       || isMATLAB('<=', [8,3]);   #FIXME: 749

test/private/StreamMaker.m
egeerardyn: Hmm, you are right, there is something very fishy going on in there.

The cases that I see are:
 - `fid==1`
 - `fid==2`
 - `ismember(fid, fopen('all'))`

and everything else should be an error.

test/suites/ACID.m
egeerardyn: Nice :+1:

egeerardyn:
@miscco Thanks for reviewing!

I will merge this into `develop`, that way we can start using Jenkins for other PRs as well.

PR #751, opened 2015-08-28
Test fix octave4

okomarov:
check depedencies with `apt-cache depends octave`


okomarov:

okomarov:
@egeerardyn How do I test on Travis?

egeerardyn:
Travis works automatically and otherwise you have to login on Travis and go to the build page, there is a rebuild button that looks like :leftwards_arrow_with_hook:, I think

okomarov:
This is as far as I get: our ACID fails on line 192, at `plot(m)`.

The call that is probably causing the problem is:
```
'axes' is a function from the file /usr/local/share/octave/4.0.0/m/plot/util/axes.m line 67
    if (__is_handle_visible__ (htmp))
```

where `__is_handle_visible__`  is builtin, hence compilation of the PPA release under some OS, is buggy for (Ubuntu/Linaro 4.6.3-1ubuntu5) 4.6.3, travis' version.

egeerardyn:
I don't think we need this PR anymore. Closing. Thank you a lot for your investigation in Octave/Travis/OpenBLAS history! :+1:

PR #752, opened 2015-08-30
Make new/expected hash clearer in reportt

egeerardyn:
Requested by @okomarov, I kept a bit of the verbosity, but I'm not feeling strongly in favor/against one thing or another.


okomarov:
Thanks.

PR #753, opened 2015-08-30
Pr716 rebased

egeerardyn:
This is a rebased version of #716 to allow further testing.


egeerardyn:
Hashes don't seem changed (which is a bit odd, but okay). Graphically, everything looks reasonable in R2014a and R2014b, so this is ready for merging.

PR #750, opened 2015-08-26
Vectorize movePointsCloser

miscco:
This is an upgrade to pull request #737. 

In addition to the changes there, this vectorizes the movePointsCloser function. The new function is now considerably simpler and we can keep most of the previous structure of the code.

As the last changes were complicated enough i decided to split it into 2 separate pull requests

There is one possible functional change. The old code calculated the xNew point and checked if it would still cross one of the axes. However, per construction this has to be true, otherwise xNew would have not been updated.

The usual tests for MATLAB2015a fail:

# MATLAB 8.5 (Linux 3.13.0-62-generic) commit 9f193c4675 

## Unreliable tests
These do not cause the build to fail.

|   Testcase |                   Name |                       OK |                                                                      Status |
| :--------- | :--------------------- | :----------------------: | :-------------------------------------------------------------------------- |
|  `ACID(1)` |     `multiline_labels` | :heavy_exclamation_mark: | hash 21a208748863224355861184cf9910f7 != (01d4eeae27254d91a0f8316897e23e81) |
|  `ACID(6)` | `sine_with_annotation` | :heavy_exclamation_mark: | hash 995cd2a901446c59f85552d48025ba27 != (7d415c014eec35c249802eb3a3ee519e) |
| `ACID(10)` |       `peaks_contourf` | :heavy_exclamation_mark: | hash 1241fa010c4c53ab967f1e61931b596f != (77bfa3a9f07e49b9061d919551ad0a0f) |
| `ACID(16)` |              `logplot` |       :white_check_mark: |                                                                             |
| `ACID(18)` |           `legendplot` | :heavy_exclamation_mark: | hash d37d19ba36d911a0db60095db48b08ef != (746423d1c0d403421667221a6b0528bf) |
| `ACID(24)` |          `quiver3plot` |       :white_check_mark: |                                                                             |
| `ACID(34)` |                 `bars` |       :white_check_mark: |                                                                             |
| `ACID(38)` |          `subplot2x2b` |       :white_check_mark: |                                                                             |
| `ACID(40)` |        `subplotCustom` |       :white_check_mark: |                                                                             |
| `ACID(42)` |            `bodeplots` | :heavy_exclamation_mark: | hash ca7b2029a08a2a311a94b73339e2128a != (8f6cafc1771cc3a506c8fc46c15ec43e) |
| `ACID(43)` |           `rlocusPlot` |       :white_check_mark: |                                                                             |
[...]

PR #695, opened 2015-07-13
Fix octave 4

PeterPablo:
### Warning: might be rebased without prior notice

**Do not merge!**

Octave is not able to run the testsuite without changes. This PR is trying to get things into shape.

* [x] `getEnvironment()` would not properly detect Octave 4
* [x] `fileOpenForWrite()` is now deferring from passing the third argument as empty
* [x] `generateColorDefinitions()` doesn't choke on names anymore
* [x] add comments to touched functions
* [x] get rid of `warning: implicit conversion from numeric to char` (near line `newstr(2,1:nElem-1) = {delimiter};`)
* [x] fix check for installed toolboxes in `ACID.m`
* [ ] get rid of `warning: implicit conversion from numeric to char` for `ACID(74)`
* [ ] `ACID(87)`: plot fails -- should be fixed by #669
* [x] `testGraphical` results in `Segmentation fault` :see_no_evil: 

### New octave
* supports some annotations --> ACID(88), ACID(89), ACID(90)
* contains penny dataset --> ACID(9)
* ...

PR #756, opened 2015-09-03
Vectorize move points closer rebased

miscco:
This is a rebased version of PR #750, that avoids a conflict with the Octave hash table.

### DIFF
--- a/src/cleanfigure.m
+++ b/src/cleanfigure.m
@@ -678,14 +678,15 @@ function movePointsCloser(meta, handle)
         dataInsert_right = [cell(1); dataInsert_right];
       end
 
-      % Add the new data for right points first, then the possible NaN and then
-      % the new left point
-      dataCell = cellfun(@(x, y) vertcat(x, y), dataCell, dataInsert_right, 'UniformOutput', false);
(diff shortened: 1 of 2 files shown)


egeerardyn:
Octave passes here :+1:

egeerardyn:
Since all comments from #750 were addressed, I think we can merge this.

PeterPablo:
Nice! Thank you @miscco for your effort in improving `cleanfigure()`! Hopefully you will stick around and keep on submitting pull requests :smiley:

miscco:
Thx, i thought about expanding the code to 3D but it kind of feels like an unbearable amount of pain ;)

okomarov:
@ Hahaha, I feel your pain :). Thanks for your contributions, really appreciated. 

I do have a suggestion thogh in the cleanfigure domain: per @egeerardyn [suggestion](https://github.com/matlab2tikz/matlab2tikz/pull/695#issuecomment-134517983) we would need a "small-value" rounder (maybe a simpler version first, without alpha). It would help with Latex memory limits and the cross-version compatibility of the testing suite.

egeerardyn:
@miscco Indeed, thank you a lot for your hard work!

@okomarov Good suggestion. I'm going to open a ticket such that we can keep track of that one.

PR #758, opened 2015-09-03
This is a hacky try to set a fixed data precision

miscco:
I am currently unsure, whether this is a correct approach and this is actually more of a request for comments.

### DIFF
--- a/src/cleanfigure.m
+++ b/src/cleanfigure.m
@@ -118,12 +118,14 @@ function indent = recursiveCleanup(meta, h, targetResolution, indent)
       %display(sprintf([repmat(' ',1,indent), '  handle this']))
 
       if strcmp(type, 'line')
+          setDataPrecision(meta, h);
           simplifyLine(meta, h, targetResolution);
           pruneOutsideBox(meta, h);
           % Move some points closer to the box to avoid TeX:DimensionTooLarge
           % errors. This may involve inserting extra points.
           movePointsCloser(meta, h);
       elseif strcmpi(type, 'stair')
+          setDataPrecision(meta, h);
           pruneOutsideBox(meta, h);
       elseif strcmp(type, 'text')
           % Ensure units of type 'data' (default) and restore the setting later
@@ -170,6 +172,63 @@ function indent = recursiveCleanup(meta, h, targetResolution, indent)
               set(h, 'Visible', 'off');
           end
       end
+  end
+
+  return;
+end
+% =========================================================================
+function setDataPrecision(meta, handle)
+  % Make sure, data is within presicion bounds
+  
+  % Retrieve data
+  xData = get(handle, 'XData');
+  yData = get(handle, 'YData');
+  
+  % Obtain zData, if available
+  if isprop(handle, 'ZData')
+    zData = get(handle, 'ZData');
+  else
+    zData = [];
+  end
+
+  if isempty(zData)
+    data = [xData(:), yData(:)];
+  else
+    data = [xData(:), yData(:), zData(:)];
+  end
+
+  if isempty(data)
+      return;
+  end
+  
+  % Get the data range for every axis  
+  xLim = get(meta.gca, 'XLim');
+  yLim = get(meta.gca, 'YLim');
+  
+  if ~isempty(zData)
+      zLim = get(meta.gca, 'ZLim');
+  end
+    
+  if isempty(zData)
+    range = [diff(xLim), diff(yLim)];
+  else     
+    range = [diff(xLim), diff(yLim), diff(zLim)]; 
+  end
+  
+  % Get the maximal precision for the data range
+  eps_range = eps(range);
+  
+  % Scale the data wrt the range
+  data  = bsxfun(@rdivide, data, eps_range);
+  
+  % Round to precision and scale back
+  data  = bsxfun(@times, round(data), eps_range);
+  
+  % Override with the new data
+  set(handle, 'XData', data(:, 1));
+  set(handle, 'YData', data(:, 2));
+  if ~isempty(zData)
+    set(handle, 'ZData', data(:, 3));
   end
 
   return;
--- a/src/cleanfigure.m
+++ b/src/cleanfigure.m
@@ -118,15 +118,15 @@ function indent = recursiveCleanup(meta, h, targetResolution, indent)
       %display(sprintf([repmat(' ',1,indent), '  handle this']))
 
       if strcmp(type, 'line')
-          setDataPrecision(meta, h);
           simplifyLine(meta, h, targetResolution);
           pruneOutsideBox(meta, h);
           % Move some points closer to the box to avoid TeX:DimensionTooLarge
           % errors. This may involve inserting extra points.
           movePointsCloser(meta, h);
-      elseif strcmpi(type, 'stair')
           setDataPrecision(meta, h);
+      elseif strcmpi(type, 'stair')
           pruneOutsideBox(meta, h);
+          setDataPrecision(meta, h);
       elseif strcmp(type, 'text')
           % Ensure units of type 'data' (default) and restore the setting later
           units_original = get(h, 'Units');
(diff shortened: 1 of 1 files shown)


egeerardyn:
That's what PRs are for :smile:.

I think it might produce cleaner code to write this function generically (i.e. independent of XData/YData/ZData) and then call it repeatedly instead of building data objects containing X/Y/Z data combined, e.g. something like 

```
function processData(h, keyword, processing)
    if isprop(h, keyword)
        data = get(h, keyword);
        data = processing(data);
        set(h, keyword, data);
    end
end

...
processData(h, 'XData', @limitPrecision);
processData(h, 'YData', @limitPrecision);
processData(h, 'ZData', @limitPrecision);
...

function data = limitPrecision(data)
    ...
end
```

The main point is that X/Y/Z should get rounded separately since they may have different scales. I think the code should reflect that to avoid confusion.

PeterPablo:
@egeerardyn -- exactly what I thought.
@miscco thank you for the initiative to tackle this. If this is in, it should become easier to merge the octave 4 PR.

okomarov:
I agree with Eagon and his naming convention but I would call it after every other routine.

egeerardyn:
@okomarov I'm unsure whether this should limiting the precision should be the first or last step in the pipeline.

What are your reasons for preferring it at the end?

PeterPablo:
A reason why it might make sense to do this last is the following: If a point is first set to exactly "0" then it might get moved afterwards by the line simplification scheme. This is probably undesired behavior.

egeerardyn:
@PeterPablo Am I mistaken that the simplification algorithm only removes points? Then it shouldn't matter. If it does move the points around, indeed, it would be better to have it at the end.

Now that I think of it: it's exactly the reason why having it at the end will work the best: limiting the precision cannot be undone anymore if it's the last step.
On the other hand, some simplifications might work better if we reduce the precision beforehand, since we are more likely to encounter duplicate values that were otherwise hidden as slightly different values.

PeterPablo:
You are of course correct. The current implementation (opheim simplification) one deletes superfluous points. As this possibly could change in future, I still would set the data precision last. Feel free to change my opinion by presenting a MWE that would produce better results the other way round :-)

okomarov:
The current PR does not limit precision, but rounds values below a threshold to 0. So, maybe `limitPrecision` is a bit ambiguous. 

Also, the simplification routine, already performs pixelation, before running opheim's also. Hence, rounding some values to zero on the whole dataset might just be an unnecessary performance cost. Also, it might sounds as if we are doing things twice, but the pixelation is *basically* filling each screen pixel with at most one data point. If the data point that fills a pixel has a 0.000000000001 coordinate, it stays like that.

To me rounding it to zero, is more of a representation convenience, which should be harmless at the end. 

We also might need to consider log-scale...

egeerardyn:
@okomarov It does limit the precision by
```matlab
data  = bsxfun(@rdivide, data, eps_range);
% Round to precision and scale back
data  = bsxfun(@times, round(data), eps_range);
```

Its idea is equivalent to `round(data/scale)*scale` (except for dimensions and so on). I don't think it makes sense to treat `0` any different from other numbers: `N + e` where `|e| < scale` should also be mapped to `N`, whether `N` is zero, one or pi shouldn't matter.
For integer that will make the notation nicer.

There is indeed some pixelation happening in the simplification algorithm.
I wouldn't worry about the performance cost of this precision limitation:
 - premature optimization ... (you know the rest)
 - this code is essentially vectorized: the major lifting is done at a low level which is executed by C/FORTRAN code instead of interpreted MATLAB. All branches (essentially `isprop(...)`) happen at very high level and don't get called more than once (per HG object).

That makes it fundamentally different from the simplification where the algorithm is more intricate and the major work is performed in MATLAB (i.e. JIT-compiled/interpreted MATLAB code) that branches a lot and many pieces of code that get interpreted more than once per HG object.

You are completely right about logarithmic scales. We certainly have to test it properly. I don't currently think we should account for that transformation in the quantization process, but maybe the quantization causes unwanted behavior when `data = 0` (or rather, gets rounded to 0).

okomarov:
@egeerardyn I got the precision rounding (and crossed out wrong comments) probably when you were writing the explanation. I think it's a good explanation and can be implemented in the help of the function. 

Also correct on performance, not impacted at all. 

Still, I logically prefer this type of rounding to happen at the end. The reason is that, all subsequent modifications become a composite function of the precision limitiation, and I am not sure this is free of edge cases. Putting it at the end, does not compose the data manipulation functions and while still does not guarantee to be free of edge cases, it is simpler to identify where the case might be coming from.

miscco:
@egeerardyn Yeah, that is a very nice approach.

@egeerardyn Nr2. I believe scaling with regard to diff(xlim) is the correct approach. Otherwise large values as well as small values would be troublesome

I was also unsure about whether to put it in front or back. I believ actually both are correct. Fixing the precision ensures, that the following operations result in the same value.

Fixing last should ensure, that there are no differences between values. In principle there should only be one hash file, not N for every build system.

miscco:
Btw, can we get rid of the bsxfun by 
```matlab
eps_range = eps(range)*eye(length(range));
data  = data / eps_range;
data  = round(data)* eps_range;
````

**Comment by okomarov** [comment]
@miscco  Now, I see your point about `diff(xlims)`.

This example clarifies (extreme case):
```
x = [1,2]
y = [100.000000000001, 100.000000000002]
plot(x,y)
ylim([100,100.000000000003])
```

**Comment by egeerardyn** [comment]
@okomarov @miscco I still don't see why you prefer `diff`. Especially in @okomarov's example, that's overly conservative.

I made a little script to check what's happening. 
I calculate the precision based on the data, so implicitly this assumes that the data covers the whole axis limits. Not that this should matter really.

What you see are different datasets (black) and different strategies (lines) to pick an epsilon to be used for the whole dataset. You also see the actual epsilon per point (blue circles). Basically, everything below this blue line is nonsense: such values cannot be represented as an IEEE double (or actually: if you disturb the data by that amount, it will not budge).

```matlab
strategies = {'max',  @(limits) max(abs(limits)); ...
              'min',  @(limits) min(abs(limits)); ...
              'diff', @(limits) max(limits)-min(limits); ...
              'mean', @(limits) mean([max(limits) min(limits)])};

N = 100;
datasets = {linspace(100, 100.000001, N)
            logspace(-5,5,N)};

nDatasets = numel(datasets);
nStrategies = size(strategies,1);

figure;
for iD = 1:nDatasets
    data = datasets{iD};
    
    subplot(1, nDatasets, iD); hold on;
    set(gca, 'YScale','log')
    plot(1:N, data, 'k.:', 'DisplayName', 'Data');
    plot(1:N, eps(data), 'bo--', 'DisplayName', 'eps(data)');
    
    for jS = 1:nStrategies
        strategy = strategies(jS,:);
        epsVal =   eps(feval(strategy{2}, data));
        description = sprintf('%s ( eps  = 10 ^%f)', strategy{1}, log10(epsVal));
        disp(description)
        plot([1 N], epsVal*ones(2,1), '-', 'DisplayName', description);
    end
    legend('show')    
end
```

The produced figure:

eps

And output:
```
```

okomarov:
With my example, I was thinking about the (ill) case where a user wants to show values that are all quite big but change just by a small amount, i.e. when those values change by an amount smaller than `eps(value)`. In this case, if you wanted to plot how the values change, the `abs(value)` would actually change location of the points in a visible manner. Obviously, this simply cannot happen, as @egeerardyn just pointed out.

Also, the ylims axis has some checks in place, so you cannot zoom into territory smaller than eps:
```
x = [1,2];
y = [100.000000000001, 100.000000000002];
h = plot(x,y)
ylim(y)
```

*

miscco:
Hm, i tried your version with the generic function and it introduced a lot of incorrect values for some plots. Which incidentally let brought the handling of Inf values to my attention #761 ;)

miscco:
So i am unsure, whether the generic approach is really valid.

Many properties are specific per axis, e.g. XScale and YScale. You cannot make a generic function, that handles XData *and* considers XScale, as you would have to hardcode XScale. 

On the other hand, when passing the property XScale as an argument, one looses the generic approach, as a routine, that does not care about XScale but something else needs a special treatment.

egeerardyn:
@miscco You can follow a similar approach as what we do in `matlab2tikz` and construct the corresponding keywords when needed.

```matlab
processData(h, 'X', @limitPrecision);
processData(h, 'Y', @limitPrecision);
processData(h, 'Z', @limitPrecision);

function processData(h, axis, processing)
    dataKeyword = [upper(axis) 'Data];
    scaleKeyword = [upper(axis) 'Scale'];
    limitsKeyword = [upper(axis) 'Lim'];
    ...
    if isprop(h, dataKeyword)
        data = get(h, dataKeyword);
        scale = get(h, scaleKeyword);
        limits = get(h, limitsKeyword);
        data = processing(data, scale, limits);
        set(h, keyword, data);
    end
end
```

To avoid the problem you mention, i.e. the many (varying) parameters that need to be passed, in the same framework, we could also structure this like below:

```matlab
processData(h, @limitPrecision, 'XData', 'XScale', 'XLim');
processData(h, @limitPrecision, 'YData', 'YScale', 'YLim');
processData(h, @limitPrecision, 'ZData', 'ZScale', 'ZLim');

...
function processData(h, func, varargin)
    keywords = varargin;
    if isprop(h, keywords{1})
        arguments = cellfun(@(kw) get(h, kw), keywords, 'UniformOutput', false);
        data = func(arguments{:});
        set(h, keywords{1}, data);
    end
end

...
function newData = limitPrecision(data, scale, lim)
   ...
end
```

That way, `limitPrecision` is completely unaware of the implementation details of HG objects and which axis it's working over.

I'm quite surprised that the implementation would produce different values depending on whether you process all data at the same time (but still independently per axis) as you do or explicitly decouple the code by axis.

miscco:
Narf found the bug, i didnt check for zData, now it runs properly. However I am still unsure, how to handle this. Sadly i checked the simple Bode plot example
```matlab
bode(tf([1],[1 1]));
```

Where we already produce some strange things. And i am unsure how to fix it.

egeerardyn:
@miscco Can you push your latest changes, that makes it easier to discuss everything :+1:

miscco:
The only changes that i have compared to develop is PR #761

okomarov:
I would use all lowercase properties, for [Octave compatibility](https://www.gnu.org/software/octave/doc/interpreter/Axes-Properties.html) and like the 

```
processData(h, axisLabel, processing)
...
end
```

**Note:** `axis()` is a builtin function already. No harm here in using it as an argument, but safer with `axisLabel`

egeerardyn:
@okomarov I agree about `axis()` (function) vs `axis` (argument). I think this is named `axisSpecifier` in `matlab2tikz`, since `axisLabel` could be confused with `XLabel` etc. Maybe we should use that instead, then?

I don't like the lowercase properties since those are not very readable (and Octave should follow MATLABs behavior, not the other way around).
Octave does not make a problem from `XData` and so on (maybe in versions before 3.6, which we no longer support). 
But we certainly add
```matlab
if isOctave()
    keywords = lower(keywords);
end
```
to be on the safe side, here.

okomarov:
@egeerardyn :+1: to both.

miscco:
Hm, i am currently toying around with it, but the problem is, that for example XScale is not a property of the handle so one has to keep them somehow separated. I have to think about it a bit when i have some time.

egeerardyn:
@okomarov I'm unsure whether a by-axis approach is more desirable. The problem with that is, that we would still need to visit the tree structure of the graphics at that stage, while I think it would be advantageous to abstract those details away since it's possible the data structure is embedded quite deeply in the tree (depending on how the graph is created and what kind of objects are used).

IMHO, the simplifications should only care about the `data` (maybe how they are shown, e.g. markers/lines), and how the data is viewed (axis limits/scale/direction). This will also make the code easier to test (independent of the HG framework, so it's much simpler and faster and can immediately test the algorithms instead of wrestling with the HG framework in all its different versions).

I thought about that problem yesterday: one possible approach would be to let `processData` also `get()` values from the `Parent` of `h` (and their parents and so on) if a property is not available in `h`. But this would hide quite some properties (e.g. we couldn't make code specific to deal with `axis.Color` since `line.Color` also exists). :-1:

@okomarov Your suggestions of just passing the extra arguments directly (instead of "preprocessing" those with `get`) is a very good one. We probably need to look into structuring how we get those values, but that seems feasible to me :+1:

okomarov:
To be honest, I re-thought about it, and it feels like we are trying to think way too much ahead (and thus falling into the common programmer's mistake that is building a spaceship for a pizza delivery :)) . 

I think the first implementation (1d270f1f8ffd7890def1562bff60d1ba5ac16385), is easy to read and does the job. So, I'd actually go ahead and merge that, and postpone a generalization for the next time that a new data manipulation task arises. This has the benefit that with two potentially different data manipulation tasks, we can determine what (and if) part can be generalized and how.

Independent on what road we pick, we still need to insert a `multiplier` to fine tune the rounding. Ideally it should be an argument and not hardcoded.

egeerardyn:
@okomarov Probably you are right (i.e. [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) principle).

Obligatory video below, since you brought it up:
[IMAGE ALT TEXT HERE](https://youtu.be/7TYJyCCO8Dc?t=18s)

okomarov:
I loved that movie, how happy was he when they finally needed the spaceship :laughing:!

miscco:
So I will most probably have some spare time in October. I will go back to the initial version of this to just get some simplification done.

miscco:
So given the new restructured form of cleanfigure, this becomes rather simple. So inf Favor of #791 this can be closed.

PR #759, opened 2015-09-04
Fix octave4 rebased

egeerardyn:
**This is a rebase of #695.**

## TODO (shamelessly copied)

Octave is not able to run the testsuite without changes. This PR is trying to get things into shape.

* [x] `getEnvironment()` would not properly detect Octave 4
* [x] `fileOpenForWrite()` is now deferring from passing the third argument as empty
* [x] `generateColorDefinitions()` doesn't choke on names anymore
* [x] add comments to touched functions
* [x] get rid of `warning: implicit conversion from numeric to char` (near line `newstr(2,1:nElem-1) = {delimiter};`)
* [x] fix check for installed toolboxes in `ACID.m`
* [x] get rid of `warning: implicit conversion from numeric to char` for `ACID(74)`
* [x] `testGraphical` results in `Segmentation fault` :see_no_evil: 

## Plots failing
* [x] `ACID(87)` : plot fails: `cback2` is not a supported `HeadStyle` for `arrow`
* [x] `ACID(87)` : plot fails: `color` is not supported for `ellipse`
* [x] `ACID(90)` : plot fails: unit `characters` not supported

## Tests fixed from failing
* [x] 16
* [x] 32
* [x] 35
* [x] 61

## Tests marked unreliable
* [x] 29
* [x] 30
* [x] 31 
* [x] 33

### New octave
* supports some annotations --> ACID(88), ACID(89), ACID(90)
* contains penny dataset --> ACID(9)
* ...

PR #761, opened 2015-09-07
Move pruneOutsideBox before simplifyLine

miscco:
If simplifyLine comes before pruneOutsideBox, it will tamper inf values.

This simple plot exemplifies the problem:
```
plot(1:5,[1,2,inf,4,5]); 
cleanfigure;
```
Instead first call pruneOutsideBox, that replaces the inf value with a NaN point

### DIFF
--- a/src/cleanfigure.m
+++ b/src/cleanfigure.m
@@ -610,16 +610,13 @@ function movePointsCloser(meta, handle)
   
   % Only try to replace points if there are some to replace
   if (~isempty(replaceIndices))
-      % Check whether the points are finite.
-      DataIsFinite = all(isfinite(data(replaceIndices,:)), 2);
-
       % Get the indices of those points, that are the first point in a
       % segment
-      id_first  = replaceIndices(replaceIndices < size(data, 1) & DataIsFinite);
+      id_first  = replaceIndices(replaceIndices < size(data, 1));
 
       % Get the indices of those points, that are the second point in a
       % segment
-      id_second = replaceIndices(replaceIndices > 1  & DataIsFinite);
+      id_second = replaceIndices(replaceIndices > 1);
 
       % Define the vectors of data points for the segments X1--X2
       X1_first  = data(id_first,    :);
--- a/src/cleanfigure.m
+++ b/src/cleanfigure.m
@@ -118,8 +118,8 @@ function indent = recursiveCleanup(meta, h, targetResolution, indent)
       %display(sprintf([repmat(' ',1,indent), '  handle this']))
 
       if strcmp(type, 'line')
-          simplifyLine(meta, h, targetResolution);
           pruneOutsideBox(meta, h);
+          simplifyLine(meta, h, targetResolution);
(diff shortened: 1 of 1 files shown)


egeerardyn:
Nice find! :+1: Code looks good, but I want to wait until #759 is merged to also test this on Octave 4.0 and see what happens there (I guess we don't cover that case with our test suite, though).

okomarov:
Visually, the bug in `simplifyLine()` :
![untitled](https://cloud.githubusercontent.com/assets/3731173/9716752/c058cc10-5564-11e5-936e-caa7b27b89e7.

The fix is logical, however, we should make it explicit that `pruneOutsideBox` should always come first because we are introducing dependence on the execution order.

egeerardyn:
@okomarov : :+1: I fully agree.

miscco:
So here i am unsure, what i should change? Would a comment in the line of 

```matlab
% Always remove invisible points before simplifying the line
```

miscco:
On a more serious note. What do you think about removing Inf completely beforehand? In principle every Inf data point get converted to a NaN anyway. Therefore, we can transform them beforehand and remove most the isfinite checks.

egeerardyn:
Just a comment would do, indeed.

Preferably where the order is set, so near the fragments like
```
%TODO: e.g. your comment here somewhere (you can make it stand out, because it is something quite subtle but important
pruneOutsideBox(meta, h);
simplifyLine(meta, h, targetResolution);
movePointsCloser(meta, h);
```

BTW, notice how Travis now :heart: your code :+1: 

I don't think removing the `Inf`s before gains us a lot right now (but to be honest, I don't have `cleanfigure`'s code in my mental memory, so this is just from looking at your diffs and seeing where those checks happen). Let's first merge in this PR.

But if you think we can simplify the code significantly, we can certainly do what you suggest; but we can do so in a new PR.

miscco:
I will finish this tomorrow. Thinking about it i believe, that simplifyLine should come last. As it also depends on movePoints

okomarov:
@miscco This is one way to think about it. The other would be to make it self-contained and remove the infs in `simplifyLine` instead of moving the sub-routines. 

Both approaches have trade-offs. Moving `simplifyLine` last, penalizes performance in a stress test case (unlikely in everyday use) and slows execution time by 8%, hence no biggie. Also, introduces dependence (which can be more serious, but cleanfigure is simple enough still). The pro is that it can allow better modularity of the code: as you pointed out, no need to check for `inf`s evrywhere.

miscco:
Can you elaborate how it penalizes performance? The only thing pruneoutsidebox does, is removing points, so in principle it should improve performance?

okomarov:
If you have 1e6 points, it will operate on that amount instead of the reduded set produced by `simplifyLine` (recall my sine example).

miscco:
As i have falsely pushed this on my develop branch, i will reopen a new PR with a seperate branch to avoid merging problems

PR #763, opened 2015-09-08
try to fix locale to something else than `C`

PeterPablo:
current ACID(1) fails on Travis due to:
```
warning: ft_render: failed to decode string `one ° ∞ three' with locale `C'
```

Setting the locale to an existing language should solve this.


egeerardyn:
Just an idea, but maybe you can track down the issue by placing an `echo on` in the affected test case (you can also temporarily disable all other tests and revert those things afterwards) to help you speed up testing.

egeerardyn:
It seems that this is not working out (Octave doesn't seem to care about our locale) and this PR is already quite old and in non-mergeable state.
Let's open up an issue for this and re-evaluate this when we have some time and desire to tackle this.

PR #764, opened 2015-09-08
Extract function `guitypes`

egeerardyn:
This removes the only FIXME from `figure2dot`.

Note: only executed mentally, so wait for Jenkins and Travis to be sure.

### DIFF
--- a/src/figure2dot.m
+++ b/src/figure2dot.m
@@ -106,8 +106,7 @@ end
 function bool = shouldSkip(h)
     %  returns TRUE for objects that can be skipped
     objType = get(h, 'Type');
-    bool = ismember(lower(objType), {'uimenu', 'uitoolbar', 'uicontextmenu'});
-    %FIXME: maybe integrate this in matlab2tikz?
+    bool = ismember(lower(objType), guitypes);
 end
 % ==============================================================================
 function label = addHGProperty(label, h, propName, default)
--- a/src/figure2dot.m
+++ b/src/figure2dot.m
@@ -106,7 +106,7 @@ end
 function bool = shouldSkip(h)
     %  returns TRUE for objects that can be skipped
     objType = get(h, 'Type');
-    bool = ismember(lower(objType), guitypes);
+    bool = ismember(lower(objType), guitypes());
 end
 % ==============================================================================
 function label = addHGProperty(label, h, propName, default)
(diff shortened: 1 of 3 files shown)


src/private/guitypes.m
```diff
@@ -0,0 +1,13 @@
+function types = guitypes()
+% GUITYPES returns a cell array of MATLAB/Octave GUI object types
+%
+% Syntax
+%   types = guitypes()
+%
+% These types are not ignored by matlab2tikz and figure2dot.
```
PeterPablo: `not`?

src/matlab2tikz.m
```diff
@@ -681,9 +681,8 @@ function addCustomCode(fid, before, code, after)
             case 'histogram'
                 [m2t, str] = drawHistogram(m2t, child);
 
-            case {'uitoolbar', 'uimenu', 'uicontextmenu', 'uitoggletool',...
-                    'uitogglesplittool', 'uipushtool', 'hgjavacomponent'}
-                % don't to anything for these handles and its children
+            case guitypes
```
PeterPablo: `guitypes()` to emphasize that this is a function?

src/figure2dot.m
```diff
@@ -106,8 +106,7 @@ function plot_children(fh, h, parent_node)
 function bool = shouldSkip(h)
     %  returns TRUE for objects that can be skipped
     objType = get(h, 'Type');
-    bool = ismember(lower(objType), {'uimenu', 'uitoolbar', 'uicontextmenu'});
-    %FIXME: maybe integrate this in matlab2tikz?
+    bool = ismember(lower(objType), guitypes);
```
PeterPablo: `guitypes()` to emphasize that this is a function?

src/matlab2tikz.m

src/figure2dot.m
egeerardyn: Please see the other one :stuck_out_tongue:

src/private/guitypes.m
PeterPablo: What about this comment? The way I read the code, those types *are* ignored by m2t and f2d.

egeerardyn: Hmm, you are right `not` should NOT be there.

PeterPablo: yeah... I should have written ~~`not`~~.

egeerardyn: No, I overlooked this comment altogether. Thanks for making me aware!

PR #765, opened 2015-09-09
Test script

egeerardyn:
It bothered me that the Jenkins script and Travis script are a bit different, while there should be need for this.
Also: in Jenkins I had quite a few commands (see here) in the runner, this can now be tracked in git.

Currently, in Jenkins I have the following script:
```bash
if [ -f ./runtests.sh ]; then
    # Make sure to remove switches integrated in runner
    ./runtests.sh ${Runner% *}
else
    echo !! Running compatibility script !!
    export CONTINUOUS_INTEGRATION=true
    cd test
    $Runner "runMatlab2TikzTests()"
fi
```

to facilitate PRs while this one is not merged in.
Afterwards, that should simplify to just `./runtests.sh $Runner` (and some tweaking of the Build Matrix).

### DIFF
--- a/runtests.sh
+++ b/runtests.sh
@@ -8,7 +8,7 @@
 #     ./runtests.sh RUNNER SWITCHES
 #
 # Arguments:
-#   - RUNNER: the (path of) the binary you want to use to execute the tests
+#   - RUNNER: (path of) the binary you want to use to execute the tests
 #         default value: "octave"
 #   - SWITCHES: switches you want to pass to the executable
 #         default value: * "-nodesktop -r" if runner contains "matlab"
--- a/runtests.sh
+++ b/runtests.sh
@@ -1,0 +1,72 @@
+#!/usr/bin/env bash
+#
+# Test script runner for MATLAB2TIKZ continuous integration
+#
+# You can influence the execution by passing one or two parameters
+# to this function, as
+#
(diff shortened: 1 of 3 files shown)


runtests.sh
```diff
@@ -0,0 +1,72 @@
...
+#
+# Test script runner for MATLAB2TIKZ continuous integration
+#
+# You can influence the execution by passing one or two parameters
+# to this function, as
+#
+#     ./runtests.sh RUNNER SWITCHES
+#
+# Arguments:
+#   - RUNNER: the (path of) the binary you want to use to execute the tests
```
PeterPablo: remove first ~~the ~

PeterPablo:
Looks good :+1:
Minor comment concerning the the argument description. Feel free to merge.

runtests.sh
egeerardyn: :+1:

egeerardyn:
Thanks for the review!

okomarov:
@egeerardyn I think this edit gives false positive? 
This build has an error on a reliable test but yet it passes: https://travis-ci.org/matlab2tikz/matlab2tikz/builds/79546274

egeerardyn:
You are right. I will upload a fix in the morning based on http://stackoverflow.com/questions/2870992/automatic-exit-from-bash-shell-script-on-error

PR #771, opened 2015-09-10
runtests exit on error

egeerardyn:
This should fix the test runner problems noticed by @okomarov in #765

### DIFF
--- a/runtests.sh
+++ b/runtests.sh
@@ -21,8 +21,16 @@
 #  - http://stackoverflow.com/questions/229551/string-contains-in-bash
 #  - http://stackoverflow.com/questions/2870992/automatic-exit-from-bash-shell-script-on-error
 #  - http://www.davidpashley.com/articles/writing-robust-shell-scripts/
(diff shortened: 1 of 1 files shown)


egeerardyn:
This is not a complete fix. Error codes are not propagated with this approach). I'm looking into it.

egeerardyn:
The last few commits should fix everything.

There were a few things happening:
 - the environment variables `CONTINUOUS_INTEGRATION` and `CI` were not passed to MATLAB/Octave, so it didn't set the exit code
 - bash by default does not pass along error codes. I have written a short function that explicitly rethrows the exit code after our test runner.

My testing approach on this:
 - tweak the `runMatlab2TikzTests` script such that only the first 5 functions are executed (just for speed of iteration)
 - `./runtests.sh` from your shell and look that the result and error code pass (my prompt displays `✘` after an error, but you can look at `$?` as well)
 - delete the hashes file for the runner you use (i.e. ALL hashes will fail)
 - `./runtests.sh` from your shell and look that the result and error code fail

egeerardyn:
I'm merging this, since this is necessary for all other PRs to be tested without false positives.

Issue #399, opened 2014-11-03
[ACID(20)]moreLegends has overlapping legend entries


PeterPablo:
ACID(21) moreLegends fails to properly position the legends ("Sine" overlapps "Cosine"). Screenshot from MATLAB 2014b. 2014a gives identical results.
![image](https://cloud.githubusercontent.com/assets/4340267/4883734/f048d548-6364-11e4-885b-9380bfe340ec.png)

PeterPablo:
Still present, but now this is `ACID(20)`.

okomarov:
I did some investigation time ago, and in brief it is NOT possible with current pgfplots to implement it. 

When we parse the first axis, we won't find the cosine line at all! We would basically need to add the line from the second axis to the first, make it invisible, but still include it in the legend and exclude l;egend generation for second axis....

PeterPablo:
Legendentry `cosine` is still missing with 68cc381 on HG2 :-(

okomarov:
Yes, expected. Transition towards a proper implementation of legends with plotyy. Anyways, cosine label was not visible.

PeterPablo:
True. I almost closed here, since you indicated this as fixed in #760, though wanted to verify this myself before closing.

okomarov:
fixed in #773

PR #773, opened 2015-09-17
Legend plotyy

### LINKED ISSUE #140: matlab2tikz: plotyy overlapping legends
I experienced a problem when I created a figure with plotyy, markers and different
 legends. In matlab, all looks quit fine. After compiling the latex pdf, the
 legends overlap. 

**EDIT[Oleg]**  Edited out the long example. Wrote below a MWE.

okomarov:
Implements legends for `plotyy()` and fixes #140, #399 (in Matlab for now)

You can check on Matlab ACID(20) or other plotyy examples with legend :)

TODO: 
- [x] Fix errors
- [x] Octave!
- [x] do not use legend string in `\label{%s}` but some hash of the handle?


src/matlab2tikz.m
```diff
@@ -987,9 +1022,8 @@ case guitypes()
         % Take only the first child from a pure hggroup (e.g. bodeplots)
         for ii = 1:numel(entries)
             entry     = entries(ii);
-            isHggroup = any(strcmpi(get(entry,'Type'),{'hggroup', 'matlab.graphics.primitive.Group'})) & ...
-                            strcmpi(class(handle(entry)),'hggroup');
-            if isHggroup
+            isHggroupClass = strcmpi(class(handle(entry)),'hggroup');
+            if isHggroupClass
```
okomarov: Simplifies check for Hggroup class

src/matlab2tikz.m
```diff
@@ -1000,12 +1034,51 @@ case guitypes()
...
+
+end
+% ==============================================================================
 function string = getLegendString(m2t, h)
 % Retrieve the legend string for the given handle 
 
-entries = m2t.axesContainers{end}.LegendEntries;
-idx     = ismember(entries, h);
-string  = getOrDefault(entries(idx), 'displayname', '');
+string  = getOrDefault(h, 'displayname', '');
```
okomarov: No need to compare against legend entries since I need to call it on `PlotyyReferences` too. If handle does not have legend string, it will be empty.

okomarov:
@matlab2tikz/developers test this PR

okomarov:
@matlab2tikzBot : test this PR

src/matlab2tikz.m
```diff
@@ -716,6 +716,38 @@ case guitypes()
...
+        c            = prettyPrint(m2t, legendString, interpreter);
+        refs         = [refs, sprintf('\\addlegendimage{/pgfplots/refstyle=plotyyref:%s}\\addlegendentry{%s};\n',...
+                        legendString{1}, join(m2t, c, '\\'))];
+    end
+    % Add references BEFORE next plot to preserve color order
+    str = [refs, str];
+    
+    % Clear plotyy references
+    m2t.axesContainers{end}.PlotyyReferences = [];
+end
```
PeterPablo: Can there be a `else` case that we should detect?

src/matlab2tikz.m
```diff
@@ -895,7 +928,7 @@ case guitypes()
 % Get legend handle associated with current axis
 
 legendhandle = [];
-env = getEnvironment;
+env = getEnvironment();
```
egeerardyn: Wouldn't it make more sense to just `switch getEnvironment()` below? The values are cached anyway, so performance shouldn't be an issue.

src/matlab2tikz.m
okomarov:
```
else
    % Do nothing: it's gonna be a legend entry. Not a labelled reference nor a referenced entry from the main axis.
end
```

I can add that to make it clearer

src/matlab2tikz.m
egeerardyn: :+1:. However, you might want to add a short note to state that `handle` is not supported on Octave (I know this is in a MATLAB-only branch and so it shouldn't be a problem).

test/suites/ACID.m
```diff
@@ -581,7 +581,7 @@
   legend boxoff
 end
 % =========================================================================
-function [stat] = moreLegends()
+function [stat] = plotyyLegends()
```
egeerardyn: :+1:

src/matlab2tikz.m
```diff
@@ -1000,12 +1032,97 @@ case guitypes()
...
+end
+end
+% ==============================================================================
+function bool = isAxisPlotyy(h)
+% Check if handle is a plotyy axis
+
+switch getEnvironment()
+    case 'Octave'
+        try 
+            get(h, '__plotyy_axes__');
```
PeterPablo: When you perform `get(h)`, is `__plotyy_axes__` a field name or is it a *hidden* property? If the first case, we could get rid of the `try ... catch` statement.

test/suites/ACID.MATLAB.8.3.md5
```diff
@@ -48,7 +48,6 @@ markerSizes : 703bd6d9f3a26af0062adb66493f864f
 markerSizes2 : d6a12826350462efeb1f2194833cb6cc
 meshPlot : 219102d4595a4dad5b6d65ab610fea9e
 mixedBarLine : 0eb277cfaab5c17cc6738ede28886050
-moreLegends : fce1b3851870a7a2c02b6bfbf6def080
```
PeterPablo: Why is this hash removed?

PeterPablo:
I added some intermediate feedback. Thanks!

src/matlab2tikz.m
okomarov: :+1: I probably copy-pasted from somewhere :)

src/matlab2tikz.m
okomarov: hidden property, cannot test with `isfield()`, returns 0 always.

test/suites/ACID.MATLAB.8.3.md5
okomarov: renamed to `plotyyLegends`

src/matlab2tikz.m
PeterPablo: :+1: even better

src/matlab2tikz.m

src/matlab2tikz.m
PeterPablo: :+1: you might want to add this as a comment, so that nobody spends time on trying to optimize this

src/matlab2tikz.m

src/matlab2tikz.m
okomarov: I reuse `env` on line 953

egeerardyn:
@okomarov : I think "but some hash of the handle?" in your OP would be a source of randomness, which our hash-based testing could struggle with. An (ugly) but deterministic alternative would be to have a global counter for that. The other thing I can think of, is to make a string that represents where the referenced object lives in the HG object tree (i.e. figure), so e.g. `c1c3` could map to the third child of the first child of gcf.

okomarov:
I think I'll just sprintf() the absolute value of the handle with a few decimals.

egeerardyn:
@okomarov The problem with that is that we don't know the algorithm that MATLAB/Octave use to determine the handle number. As such, the LaTeX output might differ depending on:
 - which version is used
 - whether there are already figures open (and how many)
 - some other factors?

The easiest way is to introduce a global counter in `m2t` (like we do for PNG images and data files). It's ugly, but it should work as long as we don't change plotting orders.

src/matlab2tikz.m
```diff
@@ -716,6 +716,40 @@ case guitypes()
     end
 end
 % ==============================================================================
+function [m2t, str] = addPlotyyReference(m2t, str, h)
+% Create labelled references to legend entries of the main plotyy axis
+
+% This ensures we are either on the main or secondary axis
+if ~isAxisPlotyy(m2t.currentHandles.gca), return, end
```
miscco: Could you put that on separate lines. Otherwise one cannot put a break point on return.

egeerardyn: I agree.

src/matlab2tikz.m
```diff
@@ -1000,12 +1049,116 @@ case guitypes()
...
+
+    % Recover referenced legend entries of the main axis
+    m2t.axesContainers{end}.PlotyyReferences = legendEntries(~idx);
+end
+end
+% ==============================================================================
+function bool = isAxisMain(h)
+% Check if it is the main axis e.g. in a plotyy plot
+
+if ~isAxisPlotyy(h)
```
egeerardyn: Just a minor thing: I think it makes more sense to do:
```matlab
bool = true;
if isAxisPlotyy(h)
    ...
end
```

than (the current)
```matlab
if ~isAxisPlotyy(h)
    bool = true;
    return
end

...
```

as the latter makes the control flow more obscure IMHO (for exceptional cases (e.g. invisible lines, empty plots, ...), early `return`s can sometimes be advantageous, but I don't think that is the case here).

egeerardyn:
:+1: thank you for looking into this hiatus we had. Feel free to merge whenever you like.

PR #776, opened 2015-09-21
Move simplifyLine after pruneOutsideBox and movePointsCloser.

miscco:
This is a clean version of #761 

@okomarov  simplifyLine does operate on a different set of data points than pruneOutsideBox or movePointsCloser. The former acts on isInBox and the latter two on ~isInBox. Therefore they should be easily replacable

### DIFF
--- a/src/cleanfigure.m
+++ b/src/cleanfigure.m
@@ -118,11 +118,13 @@ function indent = recursiveCleanup(meta, h, targetResolution, indent)
       %display(sprintf([repmat(' ',1,indent), '  handle this']))
 
       if strcmp(type, 'line')
-          simplifyLine(meta, h, targetResolution);
           pruneOutsideBox(meta, h);
           % Move some points closer to the box to avoid TeX:DimensionTooLarge
           % errors. This may involve inserting extra points.
           movePointsCloser(meta, h);
+          % NOTE: Always remove invisible points before simplifying the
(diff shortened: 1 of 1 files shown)


okomarov:
Yes, but you still need to check who is in and who is out and you will still be indexing into a much larger array. As I was saying previously, this represents a negligible cost of 8% of overall runtime, which in absolute terms was about 3seconds. 

BTW, this is good to go.

src/cleanfigure.m
```diff
@@ -118,11 +118,13 @@ function cleanfigure(varargin)
...
 
       if strcmp(type, 'line')
-          simplifyLine(meta, h, targetResolution);
           pruneOutsideBox(meta, h);
           % Move some points closer to the box to avoid TeX:DimensionTooLarge
           % errors. This may involve inserting extra points.
           movePointsCloser(meta, h);
+          % NOTE: Always remove invisible points before simplifying the
+          % line. Otherwise it will generate additional line segments
+          simplifyLine(meta, h, targetResolution);
```
okomarov: This comment should go above `pruneOutsideBox()`.

miscco:
So i "cleaned" (TM) it up a little bit. I submit a new PR.

miscco:
Argh, and i closed the wrong one. It is included in PR #779 but i will still repoen it for the sake of getting it merged.

PR #778, opened 2015-09-23
This tries to simplify the line of 3D plots, 

miscco:
The algorithm utilizes the transformation matrix obtained by viewmtx from the camera angel to project the 3D plot onto the 2D image plane and then feeds the resulting 2D coordinates into the simplifyLine/pixelate function.

As the algorithm would break if one dimension contains only NaNs it also includes #761

### DIFF
--- a/src/cleanfigure.m
+++ b/src/cleanfigure.m
@@ -443,13 +443,22 @@ function simplifyLine(meta, handle, targetResolution)
     if hasMarkers
         % Pixelate data at the zoom multiplier
         mask   = pixelate(data(:,1), data(:,2), xToPix, yToPix);
+        % If some of the data was logarithmic get the original data
+        if isXlog
+            xData = get(handle, 'XData');
+        end
+        if isYlog
+            yData = get(handle, 'YData');
+        end
+        if isZlog
+            zData = get(handle, 'ZData');
+        end
+
+        % Update the data
+        set(handle, 'XData', xData(mask));
+        set(handle, 'YData', yData(mask));
         if is3D
-            set(handle, 'XData', xData(mask));
-            set(handle, 'YData', yData(mask));
             set(handle, 'ZData', zData(mask));
-        else
-            set(handle, 'XData', data(mask, 1));
-            set(handle, 'YData', data(mask, 2));
         end
         return
     end
@@ -489,19 +498,26 @@ function simplifyLine(meta, handle, targetResolution)
         end
     end
 
-    % Merge the cells back together
-    data     = cat(1, dataSegments{:});
+    % Merge the mask back together
     mask     = cat(1, mask{:});
 
-    % Cut the data
-    if is3D
-        set(handle, 'XData', xData(mask));
-        set(handle, 'YData', yData(mask));
-        set(handle, 'ZData', zData(mask));
-    else
-        set(handle, 'XData', data(mask, 1));
-        set(handle, 'YData', data(mask, 2));
+    % If some of the data was logarithmic get the original data
+	if isXlog
+        xData = get(handle, 'XData');
 	end
+	if isYlog
+        yData = get(handle, 'YData');
+	end
+	if isZlog
+        zData = get(handle, 'ZData');
+	end
+    
+    % Update the data
+    set(handle, 'XData', xData(mask));
+    set(handle, 'YData', yData(mask));
+    if is3D
+        set(handle, 'ZData', zData(mask));
+    end
 end
 % =========================================================================
 function mask = pixelate(x, y, xToPix, yToPix)
--- a/src/cleanfigure.m
+++ b/src/cleanfigure.m
@@ -511,12 +511,12 @@ function mask = pixelate(data, xToPix, yToPix)
 
     % Convert data to pixel units, magnify and mark only the first
     % point that occupies a given position
-    mask = [true, diff(round(data(1, :) * xToPix * mult))~=0];
-    mask = [true, diff(round(data(2, :) * yToPix * mult))~=0] | mask;
+    mask = [true; diff(round(data(:, 1) * xToPix * mult))~=0];
+    mask = [true; diff(round(data(:, 2) * yToPix * mult))~=0] | mask;
 
     % Keep end points or it might truncate whole pixels
-    inan         = any(data, 1);
-    df           = diff([false, inan, false]);
+    inan         = any(isnan(data), 2);
+    df           = diff([false; inan; false]);
     istart       = df == 1;
     pend         = find(df == -1)-1;
     mask(istart) = true;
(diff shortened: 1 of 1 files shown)


miscco:
So the following tests fail:

MALTAB2014a: failReliable = [16 36:37 70 74];
MATLAB2014b: failReliable = [16 47 70 74];
Octave: failReliable = [16 22:24 28 32:33 36:37 61 68 70 74 77 79];

I am currently short on time due to an conference next week. I dont know how far I will get with testing.

src/cleanfigure.m
```diff
@@ -347,104 +349,161 @@ function simplifyLine(meta, handle, targetResolution)
...
     end
-    if numel(xData) <= 2
-        return;
+
+    % Check whether this is a 3D plot
+    if numel(zData) >1
+        is3D = true;
+    else
+        is3D = false;
     end
```
okomarov: We should not rely on data to test 3D of the axis. Sometimes, there is ZData but it's all ones, yet the view is 2D.

I thought we had `isAxis3D` in the `test/private` but it is only in `src\private`. Please, copy that function into `test/private` and use that.

src/cleanfigure.m
```diff
@@ -347,104 +349,161 @@ function simplifyLine(meta, handle, targetResolution)
...
+        set(handle, 'YData', yData(mask));
+        set(handle, 'ZData', zData(mask));
+    else
+        set(handle, 'XData', data(mask, 1));
+        set(handle, 'YData', data(mask, 2));
+	end
 end
 % =========================================================================
-function mask = pixelate(x, y, xToPix, yToPix)
+function mask = pixelate(data, xToPix, yToPix)
```
okomarov: Totally against this change in syntax, less clear.

src/cleanfigure.m
```diff
@@ -452,19 +511,19 @@ function simplifyLine(meta, handle, targetResolution)
...
+    inan         = any(isnan(data), 2);
+    df           = diff([false; inan; false]);
     istart       = df == 1;
     pend         = find(df == -1)-1;
     mask(istart) = true;
     mask(pend)   = true;
 end
 % =========================================================================
-function mask = opheimSimplify(x,y,tol)
+function mask = opheimSimplify(data, tol)
```
okomarov: Totally against this change in syntax, unnecessary and less clear.

src/cleanfigure.m
miscco: As isAxis3D is only a 2 liner and i am actually unsure whether it is fully correct, I changed it to check for the elevation. If the elevation is 90 degree, then we have a 2D plot.

src/cleanfigure.m
```diff
@@ -347,104 +349,162 @@ function simplifyLine(meta, handle, targetResolution)
...
-        vxData = log10(xData);
+        xData = log10(xData);
     end
 
     isYlog = strcmp(get(meta.gca, 'YScale'), 'log');
-    vyData = yData;
     if isYlog
-        vyData = log10(yData);
+        yData = log10(yData);
     end
```
okomarov: We need to differentiate between visual data and the raw unscaled one:  

https://github.com/matlab2tikz/matlab2tikz/commit/bf03aafd4cb6e5b75e909256732014ae8253fd15#commitcomment-10332439

src/cleanfigure.m
miscco: I assume you would prefer pixelate(data(:,1), data(:,2), xToPix, yToPix)?

src/cleanfigure.m
```diff
@@ -347,104 +349,162 @@ function simplifyLine(meta, handle, targetResolution)
...
-        % Actual data that inherits the simplifications
-        x = xData(pstart(ii):pend(ii));
-        y = yData(pstart(ii):pend(ii));
-
-        % Line simplification
-        if numel(vx) > 2
-            mask = opheimSimplify(vx,vy,tol);
-            x    = x(mask);
-            y    = y(mask);
-        end
```
okomarov: This version is **so** much easier to read and it's not like the vectorization avoids the loop.

miscco:
So as @okomarov is against the change in pixelate and opheimSimplify arguments, let me elaborate the reasoning behind it.

The problem is, that in 2D we can use our data. However in 3D we utilize a temporary. Therefore, we can either write 2 completely separate function one for 3D one for 2D or take the solution with one 2D data array, that contains either the original data or the 2D projection.

i am completely indifferent if we pass data(:,1), data(:,2) or data

src/cleanfigure.m
miscco: Hm actually it took me quite some time to understand what this was acutally doing. However i believe it comes down to whether one has written the code or not. 

One reason why I would change it, is that it is more consistent with the way we handle the split in movePointsCloser.

okomarov:
@miscco I don't know why you are rushing, but I prefer: `pixelate(vxData, vyData, xToPix, yToPix);` with the note about preserving `vxData` and `vyData`. 

As I referred to the comment by the original author of that implementation, it's important to apply the simplification on the transformed data, but the mask on the raw one. In an earlier commit, I also fell in the same trap and simplified away vxData, but had to revert the **regression**. 

In general, we do not need to collapse everything into `data` (making the code harder to read), but should work with `vxData` and `vyData`, even more so since the projection of the 3D coordinates is on a 2D plane. Hence, I do not see the need to edit the whole code around when the 3D implementation was so elegant and minimal in the issue you raised.

miscco:
Yeah,

I completely missed the implication of the logarithmic scaling. However, I believe it is unnecessary to carry the original data around. My shot at this would be to retain the original data after the simplification if the scale was logarithmic and then update accordingly.

miscco:
So after thinking about it, i believe you are absolutely right. In the end I can do all i want without putting everything in one data structure. 

I was just rushing, as i needed that one figure to compile. Now it needs just 30 sec in contrast to getting silently killed after half an hour.

So the easiest way would be to just start fresh and cherrypick the needed changes. Otherwise i could revert all those changes by hand, most likely missing something.

PeterPablo:
Please go ahead and create a new clean pull request. We might at some point reconsider of not "allowing" rebase/squashing/force pushing, as is done in other projects (e.g. Kodi, ownCloud). I don't want to start the discussion here, but rather encourage you for this time to just close here and continue in a fresh PR. Thanks! I am looking forward to the outcome!

PeterPablo:
To answer your question: We check all tests, whose hashes changed. As you noted, most should not change (2D) -- if they do, we need to inspect why and then decide if this is good or bad. Those that should change (3D) and do change also need to be evaluated by compiling the tex output, firstly to see if the code compiles and to verify if everything looks as expected. So this is basically regression testing and  testing the newly imlemented feature. If everything is as anticipated the hashes updated for each affected environment by either using `saveHashtable()` (or similar -- don't know from the top of my head) or editing the hash files manually.

miscco:
Closed in favor of #779

PR #779, opened 2015-09-24
Cleanfigure project 3D data

miscco:
So in my defence, I got carried away a little ;)

The logic should be rather straightforward. There are 2 new functions:
```matlab
function [xData, yData] = getVisibleData(meta, handle)
function [xLim, yLim]   = getVisibleLimits(meta)
```
The first checks whether the the data is 3D and returns either the projection of the 3D data onto the image plane or the original data.
The second does the same with the limits.

Now we can utilize this in both pruneOutsideBox and simplifyLine, and handle 3D plots as well.

NOTE that this does not easily work with movePointsCloser. In pruneOutsideBox and simplifyLine we only REMOVE points or NaN them. For movePointsCloser we would need the back transformation. At work we now have MATLAB2015b so I cant really test it that well.

P.S: This also requires PR #761 otherwise the code crashes when one of the dimensions is filled only with NaNs

### DIFF
--- a/src/cleanfigure.m
+++ b/src/cleanfigure.m
@@ -505,8 +505,8 @@ function simplifyLine(meta, handle, targetResolution)
             lengthSegments = [inan(1); diff(inan); length(xData)-inan(end)];
 
             % Put the NaNs into separate segments
-            lengthSegments = [lengthSegments-1, ones(size(lengthSegments))];
-            lengthSegments = reshape(lengthSegments(1:end), size(lengthSegments));
+            lengthSegments = [lengthSegments-1, ones(size(lengthSegments))]';
+            lengthSegments = reshape(lengthSegments, numel(lengthSegments), 1);
 
             % Account for the last segment that never ends with a NaN
             lengthSegments(end-1) = lengthSegments(end-1)+1;
@@ -516,16 +516,16 @@ function simplifyLine(meta, handle, targetResolution)
         % Split data into segments
         xDataSegments = mat2cell(xData(:), lengthSegments, 1);
         yDataSegments = mat2cell(yData(:), lengthSegments, 1);
-        mask          = cell(length(lengthSegments), 1);
+        mask          = cell(size(lengthSegments));
 
         % Simplify the respective segments
         for ii = 1:length(lengthSegments)
             % Simplify only if there are more than 2 points
-            if length(xDataSegments{ii}) > 2 && length(yDataSegments{ii}) > 2
+            if numel(xDataSegments{ii}) > 2 && numel(yDataSegments{ii}) > 2
                 mask{ii} = opheimSimplify(xDataSegments{ii}, yDataSegments{ii}, tol);
             else
-                % Always print NaNs
-                mask{ii} = true;
+                % Always print NaNs and 2 point segments
+                mask{ii} = true(size(xDataSegments{ii}));
             end
         end
 
--- a/src/cleanfigure.m
+++ b/src/cleanfigure.m
@@ -181,22 +181,15 @@ end
 function pruneOutsideBox(meta, handle)
   % Some sections of the line may sit outside of the visible box.
 
-  % Check whether this is a 3D plot
-  % If the elevation is not 90° it is a 3D plot
-  [az, el] = view(meta.gca);
-  if el == 90
-    is3D = false;
-  else
-    is3D = true;
-  end
-
-  % Get the data. If it is 3D project it into the image plane
+  % Get the data. If it is 3D, project it into the image plane
   [xData, yData] = getVisibleData(meta, handle);
+
+  % Only do something if there are multiple data points
   if numel(xData) <= 2 || numel(yData) <= 2
       return;
   end
   
-  % Get the limits
+  % Get the (projected) limits
   [xLim, yLim] = getVisibleLimits(meta);
 
   % Merge data vectors
@@ -224,6 +217,9 @@ function pruneOutsideBox(meta, handle)
       % Plot points which are next to an edge which is in the box.
       shouldPlot = shouldPlot | [false; segvis] | [segvis; false];
   end
+
+  % Check whether this is a 3D plot
+  is3D = isAxis3D(meta.gca);
 
   % If some of the data was logarithmic or projected get the original data
   isXlog = strcmp(get(meta.gca, 'XScale'), 'log');
@@ -351,13 +347,7 @@ end
 % =========================================================================
 function [xData, yData] = getVisibleData(meta, handle)
     % Check whether this is a 3D plot
-    % If the elevation is not 90° it is a 3D plot
-    [az, el] = view(meta.gca);
-    if el == 90
-        is3D = false;
-    else
-        is3D = true;
-    end
+    is3D = isAxis3D(meta.gca);
 
     % Extract the data from the current line handle.
     xData = get(handle, 'XData');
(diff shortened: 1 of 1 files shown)


miscco:
So good news,

with the latest commit only ACID(24) fails for octave, although spectacularly.

The other two tests in question are 70 and 76 where I cannot spot a visual difference. I will have a look on 24 in octave.

miscco:
So this is really strange, the failling test quiverplot3D should never even execute The problem is, that at least in my octave, it isnot recognized as a quiverplot.

miscco:
So I found the ordering of the limits to be the problem. Now there is only one tests that fails reliably and that is ACID(70) where the simplifyLine actually helps. 

There is one spurious failure in ACID(76) for matlab2014b that i cant figure out. It isn't even a 3D plot

src/cleanfigure.m
```diff
@@ -322,6 +345,152 @@ function pruneOutsideBox(meta, handle)
         0.0 < lambda(:, 2) & lambda(:, 2) < 1.0;
 end
 % =========================================================================
+function [xData, yData] = getVisibleData(meta, handle)
```
okomarov: The naming should be `Visual` or something else, since the function applies transformations, i.e. log scaling and projections. `Visible` refers to a dicotomic state (on/off) and is also a very common property, hence the source of potential confusion. Moreover, the confusion is especially apparent next to the function `segmentVisible()` which actually checks if datapoint are visible or not.

src/cleanfigure.m
```diff
@@ -322,6 +345,152 @@ function pruneOutsideBox(meta, handle)
...
+    % Check whether this is a 3D plot
```
okomarov: First comment line should be an H1 line.

miscco: H1?

src/cleanfigure.m
```diff
@@ -322,6 +345,152 @@ function pruneOutsideBox(meta, handle)
...
+    end
+
+    % If this is a 3D plot, project the limits into the image plane
+    if is3D
+        % Get the projection angle
+        [az, el] = view(meta.gca);
+
+        % Projection matrix from view
+        % C        = viewmtx(az, el);
+        % NOTE: This is a subset of the MATLAB viewmtx function, as octave
```
egeerardyn: Please do **NOT** copy-paste code from neither MATLAB nor Octave. What I personally tend to do is:
1. describe what function of MATLAB/Octave you consulted and write that in a comment
2. try to understand the part of code you want to emulate,
3. write the logic/formulas on a piece of paper,
4. implement only the cases you are interested without looking at the original code.

Also, I have the impression you have quite some duplicate code in there:
 - similar fragments are present elsewhere
 - conversion from degree to radian

If you immediately convert to radians, that may also help to avoid overlap between our codebase and MATLAB/Octave.

With regard to `rem(rem(..., X),X)`, I think those statements could be simplified a bit (1 call should be enough), also I tend to prefer `mod` instead of `rem` (though I think the behavior for negative numbers is a bit different). You also might want to test whether those conversions are really necessary: you are constructing the transformation matrix from those values, but I guess `sin` and `cos` will already make sure that the right cyclicality is obeyed.

src/cleanfigure.m
miscco: Changed locally

src/cleanfigure.m
egeerardyn: [H1 is the MATLAB term for a first line of function help](http://nl.mathworks.com/help/matlab/matlab_prog/add-help-for-your-program.html?refresh=true). 

For public functions, this should be a short description (=actual H1 line) of what the function does, e.g. 
```matlab
% GETVISIBLEDATA retrieves the data in the visible coordinate frame
```
Such a H1 line makes sure that `lookfor ...` can actually provide useful functions related to `...` (have `...` in their H1 line).

For private functions (such as this one), IMHO we don't have to adhere to H1 comments, as those are not searched by `lookfor`: just a regular comment will do:
```matlab
% retrieves data of a line in the visible coordinate frame

% other comments (note the empty line in-between!)
```

That way, if one calls `help cleanfigure>getVisibibleData`, you get some info on the function, and not a random comment`.

miscco: Fixed locally

src/cleanfigure.m
miscco: Yeah, i only took parts of the function. However, it will be some time until I can have a look on that due to travelling.

As far as i see this is just the product of the 3D rotational matrices around x and z axis
(See https://en.wikipedia.org/wiki/Rotation_matrix)

egeerardyn:
@miscco 
> Also the computation of the transform matrix is standard math
I agree. As long as you don't copy-paste the code to compute that, that should be permissible to use.

> %   Copyright 1984-2007 The MathWorks, Inc.
That indeed implies that for that code file, the standard MATLAB license applies. I don't know the license exactly, but I doubt we are allowed to use ANY part of it. To ensure we don't get into a fight by TMW's lawyers, I don't want to include any code that comes verbatim from their codebase. The safest route is even not to have it in our git history at all.

Computations are fine as long as we implement everything ourselves. Providing quotations of their code is also fine when it's
 1. just a relevant/short fragment (e.g. interpretation of azimuth/elevation but NOT a substantial part of the `view.m` file),
 2. not executed (i.e. a comment in our code), and
 2. properly referenced.

Under those conditions, I think it should be within the fair-use permissions of copyright (but I am not a lawyer). Anyway, for me that's the moral boundary of using whatever material provided under non-permissive licenses (or for which licensing is unclear).

Please see my other comment for more info as well.

okomarov:
#### I have two main comments/suggestions and a few minor ones:
- license wise, use `viewmtx()` if Matlab or one of the hidden properties if Octave (see my other comment for a link to the hidden properties)
- Code re-design:
     I quite like the wrapping of `getVisualData()` and with a proper line of comment it fits well into the codebase. 

  I also see quite a few re-factorings and changes in the implementation which I usually support in two cases, if they significantly improve **performance** or the **memory footprint** (consistently with the saying "If it ain't broke, don't fix it"), or if they drastically improve readability of the code. 

  The re-factorings push vectorization a bit further (but the previous approach was already vectorized), but suffer in clarity. The vectorization, combined with a change in data handling, make the whole approach more **obscure** IMO. 

    Before, we query the raw data, make a copy and get the visual data (after log transforms), get the simplification mask off the **visual** data, apply simplifications to **raw** data. 

   Now, we query the visual data, get the mask, if the data was logscaled, apply the simplification to raw data that we query at the end (what about 3D?). 

  It does not sound too bad, but the readability and complexity of the code is lower. For someone who hasn;t worked on that, it is not immediate that the visual data is actually different from raw (and how). If we wrap the transformations into `getVisualData`, it is even more confusing when we read *again* the handling of the log scaling and we retrieve raw data to apply the mask. 

  I think we should query raw and visual data with `getVisualData()` from the beginning, so that the difference between the two is at least apparent in the implementation. Notice, that unless you can document a worse memory footprint, I do not see a good reason to swap the querying of the raw data to the end. Recall that Matlab already uses in-place optimizations, thus a copy is made only if the copied variable is altered. 

#### Minor things:

- We can use one function to get Visual data and limits
- indentation not everywhere consistent
- some of the assignments have weird space paddings (and I do like alignment of `=` as in http://www.mathworks.com/matlabcentral/fileexchange/50554-okomarov-justify)
- some comments can be improved

Since we are not in a rush, I would rather see the consensus on the first points and eventually tackle the minor things.

egeerardyn:
I agree with @okomarov: if we can get the view information by querying either `viewmtx` or some other way, we should: fewer lines of code means less maintenance work on our side.

miscco:
Closed in favor of the rebased version #790

Issue #691, opened 2015-07-07
Fix remaining quiver issues


PeterPablo:
Status as of merged #683:

R2014b:
![image](https://cloud.githubusercontent.com/assets/4340267/8543188/4d637272-249b-11e5-8ba4-077c59eca1ec.png)
![image](https://cloud.githubusercontent.com/assets/4340267/8543196/5de914bc-249b-11e5-97c3-6a07d25fcd36.png)

R2014a:
Same as R2014b

octave:
Fine, though uses still manual arrow construction. Can this be migrated to native quiver?

egeerardyn:
I've assigned this to the 1.1 release since this shouldn't be left open for too long since the output also looks off for a reasonable use case as shown above.

PeterPablo:
@jam4375 do you have an idea what the specific issue is? As far as I debugged, I am certain that the scaling factors are not correct. Would you like to work on this?

jam4375:
I believe that part of the problem is fixed by adding `point meta min=0` after `point meta={sqrt((\thisrow{u})^2+(\thisrow{v})^2)},`. For `ACID(22)` both vectors are the same size which causes problems with `\pgfplotspointmetatransformed` which maps `point meta` to `[0, 1000]` but since there is only one vector length, the transformed value is 0.

Now adding a scaling factor such as:

`every arrow/.append style={-{Straight Barb[scale length={10*max(0.01,\pgfplotspointmetatransformed/1000)},scale width={10*0.5*max(0.01,\pgfplotspointmetatransformed/1000)}]}}}`

can help fine tune the arrowhead size. In the example above `10` is the additional scaling factor.

PeterPablo:
Thanks for your input! Do you see a way of automatically determining the scaling factor (`10` in your example)? Why is it necessary to limit it to a maximum of `0.01`?

PeterPablo:
```diff
diff --git a/src/matlab2tikz.m b/src/matlab2tikz.m
index dad82ac..5b4af63 100644
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -3705,6 +3705,7 @@ function [m2t, str] = drawQuiverGroup(m2t, h)
         arrowLength = '{sqrt((\thisrow{u})^2+(\thisrow{v})^2)}';
     end
     plotOpts = opts_add(plotOpts, 'point meta', arrowLength);
+    plotOpts = opts_add(plotOpts, 'point meta min', '0');
 
[...]

PR #781, opened 2015-10-01
Native quivers improved

egeerardyn:
This is a partial fix for #691:
 - Octave now converts to native quiver plots in pgfplots
 - I've used @PeterPablo's default scaling, I did experiment a bit with different values, but for our test suites, these values seem OK
 - new parameter `'arrowHeadSizeFactor'` to allow tweaking this by hand

The actual scaling problem is again absolute (physical dimensions) vs. relative (axis units) measurements. I think by the current approach, we can already allow useable results.

In attachment: graphical results of the quiver plots:

[octave.pdf](https://github.com/matlab2tikz/matlab2tikz/files/6268/octave.pdf)
[hg1.pdf](https://github.com/matlab2tikz/matlab2tikz/files/6262/hg1.pdf)
[hg2.pdf](https://github.com/matlab2tikz/matlab2tikz/files/6264/hg2.pdf)

I've discovered some new problems as well and some small things that need fixing but are non-critical IMHO (e.g. handling of `MaxHeadSize`). For those I will create issue reports.


test/examples/example_quivers.m
```diff
@@ -0,0 +1,70 @@
...
+%
+% To know the dimensions of the arrow head, MATLAB defines the quantities
+%  alpha = beta = 0.33 that determine the coordinates of C and D as given below.
+
+clc; 
+clear variables;
+close all;
+
+%% Parameters
+syms x y z u v w alpha beta epsilon real
```
okomarov: Requires symbolic toolbox. Would be nice not having this dependency even if just an example

egeerardyn: It's not practical to get rid of this dependency (i.e. I used this script to determine the angle of the arrow heads, which is something one can do for specific values of `x,y,z,u,v,w` without that toolbox).
I will write some fall-back code such that the script also runs without the toolbox at the cost of producing less generic values.

okomarov:
@egeerardyn There is problem with the current Octave PPA, was compiled without OSmesa, i.e. off-screen printing is bugged and I cannot get the testGraphical in Octave to finally merge this. 

I looked at Matlab graphical output and it was fine.

egeerardyn:
@okomarov Thanks for checking! In Octave, can't you test with the figures visible instead or is it more fundamental than that? (To be honest, my Octave is also not really reliable on OS X).

If that doesn't help, I can always try to send you the Arch package for Octave 3.8 or 4.0 if you are using a 64 bit x86 CPU (and supposing that your distribution is as up-to-date as Arch Linux).

okomarov:
Unfortunately the new sizing of the arrowheads does not play well in Octave

BEFORE
captu1


AFTER
capturecmp

egeerardyn:
Good find. Is that Octave 3.8 or 4.0 (or both)?

okomarov:
4.0

Sent from my iPhone

On 26 Oct 2015, at 20:36, "Egon Geerardyn" <notifications@github.com<mailto:notifications@github.com>> wrote:


Good find. Is that Octave 3.8 or 4.0 (or both)?

egeerardyn:
I can also reproduce that on Octave 3.8, I just didn't cross-check those (since my Octave has been having troubles with producing EPS/PDF output files for a while, upgrading to the latest OS X didn't improve it in any way).

PeterPablo:
I think on MAC I used to change the export to a different file format that behaved "better". Alternatively the export might work better, if you switch to visible figures.

okomarov:
I can share a 3.5 gb virtual box image (with lubuntu) with octave installed that works ok for me.

egeerardyn:
@okomarov Thanks for the offer, but I think I will first check on my Jenkins server if graphical output works there.

src/matlab2tikz.m
```diff
@@ -4005,17 +4010,37 @@ function warnFacetedInterp(m2t)
...
+        % MATLAB documentation. However, in practice, there seems to be a SQRT
+        % involved somewhere (e.g. if u.^2 + v.^2 == 2, all MHS values >
+        % 1/sqrt(2) are capped to 1/sqrt(2)).
+
         arrowHeadOpts = opts_new();
-        arrowHeadOpts = opts_add(arrowHeadOpts, 'scale length', ...
-                                 '{max(0.01,\pgfplotspointmetatransformed/1000)}');
-        arrowHeadOpts = opts_add(arrowHeadOpts, 'scale width', ...
-                                 '{0.5*max(0.01,\pgfplotspointmetatransformed/1000)}');
+        arrowHeadOpts = opts_add(arrowHeadOpts, 'angle''', '18.263');
```
miscco: That and the above line should also be below the comment

src/matlab2tikz.m
```diff
@@ -4005,17 +4010,37 @@ function warnFacetedInterp(m2t)
...
         %TODO: scale the arrows more rigorously to match MATLAB behavior
-        %There is a "MaxHeadSize" property (at least in R2014b) that plays a
-        %role in determining the quiver size.
+        %There is a "MaxHeadSize" property that plays a role.
+        % Currently, this is quite hard to do, since the size of the arrows
+        % is defined in pgfplots in absolute units (here we specify that those
+        % should be scaled up/down according to the data), while the data itself
+        % is in axis coordinates (or some scaled variant). I.e. we need the
+        % physical dimensions of the axis to compute the correct scaling!
+        userInfo(m2t, ['Please change the "arrowHeadSizeFactor" option', ...
```
miscco: Could you put that code after the comments?

src/matlab2tikz.m
```diff
@@ -4005,17 +4010,37 @@ function warnFacetedInterp(m2t)
...
+                       ' if the size of the arrows is incorrect.']);
+        arrowHeadSizeFactor = sprintf(m2t.ff, abs(m2t.cmdOpts.Results.arrowHeadSizeFactor));
+
+        % NOTE: `set(h, 'MaxHeadSize')` is bugged in HG1 (not in HG2 or Octave)
+        % according to http://www.mathworks.com/matlabcentral/answers/96754
```
miscco: That note is separate to the explanation of MaxHeadSize, Can you put it at the end? So we have

Comment/Explanation
Note/Warning
Code

src/matlab2tikz.m
egeerardyn: What code are you referring to? As I see it, the code `userInfo(...)` is already after the comments `% There is a ...`.

src/matlab2tikz.m
```diff
@@ -4005,17 +4010,37 @@ function warnFacetedInterp(m2t)
...
+
+        % MaxHeadSize is said to be relative to the length of the quiver in the
+        % MATLAB documentation. However, in practice, there seems to be a SQRT
+        % involved somewhere (e.g. if u.^2 + v.^2 == 2, all MHS values >
+        % 1/sqrt(2) are capped to 1/sqrt(2)).
```
miscco: I might be wrong, but isnt that just the Pythagoras? 
a^2+b^2 = c^2

So MaxHeadSize is give as a fraction of the arrow length (default 0.2) so
```matlab
MaxHeadSize = 0.2 * sqrt(uData.^2 + vData.^2);
```

src/matlab2tikz.m
egeerardyn: If you refer to the "18.263" value, I completely agree: the comment about that should come first :+1:.

src/matlab2tikz.m
miscco: As far as i read it the comments include the note and the following explanation of MaxHeadSize calculation

src/matlab2tikz.m
```diff
@@ -98,6 +98,9 @@ function matlab2tikz(varargin)
 %   interpret tick labels as TeX. MATLAB(R) doesn't do that by default.
 %   (default: false)
 %
+%   MATLAB2TIKZ('arrowHeadSizeFactor', FLOAT, ...) allows to resize the arrow heads
+%   in quiver plots by rescaling the arrow heads by a positive scalar. (default: 10)
```
miscco: So i think this is the critial point. In principal you mix two options

1. Should the arrowHeadSize be changed

2. To what value should it be set.

So i would tend to either add a boolean value, that determines whether we should override the default value, or set the default to 0/-1 and only change the default size for arrowHeadSizeFactor>0

src/matlab2tikz.m
```diff
@@ -2724,7 +2728,8 @@ function assertRegularAxes(handle)
         cl = 'specgraph.errorbarseries';
 
     % quiver plots
-        % TODO: maybe check for existence of `udata` and absence of `ldata`
+    elseif hasProperties(h, {'udata','vdata'}, {'ldata'})
```
miscco: I dont really get this line, as ldata is part of 3D quiverplots, why check for its absence

src/matlab2tikz.m
miscco: Yes i meant that line and the 

```
arrowHeadOpts = opts_new();
```

src/matlab2tikz.m
```diff
@@ -4005,17 +4010,37 @@ function warnFacetedInterp(m2t)
...
-        arrowHeadOpts = opts_add(arrowHeadOpts, 'scale length', ...
-                                 '{max(0.01,\pgfplotspointmetatransformed/1000)}');
-        arrowHeadOpts = opts_add(arrowHeadOpts, 'scale width', ...
-                                 '{0.5*max(0.01,\pgfplotspointmetatransformed/1000)}');
+        arrowHeadOpts = opts_add(arrowHeadOpts, 'angle''', '18.263');
+        % In MATLAB (HG1), the arrow head is constructed to have an angle of
+        % approximately 18.263 degrees in 2D as can be derived from the
+        % |quiver| function. One of the example files covers this derivation.
+        arrowHeadOpts = opts_add(arrowHeadOpts, 'scale', ...
+              ['{' arrowHeadSizeFactor '*\pgfplotspointmetatransformed/1000}']);
```
miscco: Can you add an comment/explanation for the magic /1000 here?

src/matlab2tikz.m
egeerardyn: Makes sense :+1:

src/matlab2tikz.m
egeerardyn: The `sqrt` is certainly related to Pythagoras (2-norm, Euclidean distance, length, ... of the arrow).

However, what the comment meant to show is that, the `MaxHeadSize`(MHS) is not relative to the actual length of an arrow, but rather that if you have an arrow with length 2, that you see two behaviors:
 * `MaxHeadSize < 1/sqrt(2)`: changing MHS changes the size of the arrow, and
 * `MaxHeadSize >= 1/sqrt(2)`: changing MHS has no effect on the figure.

It's the easiest to see if you make yourself a little quiver plot and try it out in MATLAB (>= 2014b, due to the bugs in <=2014a) by interactively changing the MHS.

I.e. `MHS > 1/sqrt(2)` is capped at `1/sqrt(2)` it seems, but I didn't investigate further how all of that fits together to replicate the same behavior in TikZ code, but instead I wrote it down in that comment such that I (or somebody else) can pick this up later.

src/matlab2tikz.m
egeerardyn: Fair enough :+1:

src/matlab2tikz.m
egeerardyn: I disagree with you. This does not mix two options: if the user provides his own value, we should use that. Otherwise, we use our default. It may seem a bit counter-intuitive with the default value "10" instead of a "1" or "0" or something, but that allows for a more straightforward passing of the data (otherwise, it's another `if value <= 0, hsz = 10, end` or `hsz = hsz * 10` somewhere in the code: it always introduces the magic value "10" somewhere). My choice here is to just publicly announce that factor to the user (since some users may wonder why we write `10*\pgfplotspointmetatransformed/1000`, so I think it makes sense to just show it).

The same happens for "MaxChunkSize", there we also use a magic constant that is based on some quick and dirty calculations.

In any case: I think this parameter should not survive the 2.0 release, by then we should have an automatic way to determine it and get rid of the parameter altogether.

src/matlab2tikz.m
egeerardyn: Are you sure? Do note that this function is Octave-only! Normally, LData is used to store the lower errorbar, which we don't have on quivers (UData is either used for the upper errorbar or the x-direction of the arrow).

In Octave, it seems that each of the HG objects only have the relevant properties present, unlike in HG1 in MATLAB where all properties are present (HG2 is a bit more strict, but I don't know whether quivers actually do or do not have the LData in HG2).

miscco: A sorry, it was UData, VDAta and WData

http://de.mathworks.com/help/matlab/ref/quiverseries-properties.html

src/matlab2tikz.m
miscco: Ah ok,

seems that they believe, an arrow should have a longer tail than the edges. Seems kind of reasonable.

src/matlab2tikz.m
miscco: True enough, could you then drop the *Factor* in the variable name, as it seems like an absolute size.

src/matlab2tikz.m
miscco: I believe, this it the line causing the regression, as it sets a maximal size of the Arrow, regardless of its original size.

I think you should use getAndRescaleQuivers as a reference for the length of the arrow and scale it with MaxArrowSize, which might go down to 0.2 to match MATLABs behavior.

src/matlab2tikz.m
egeerardyn: It's still a factor (i.e. it scales the length), but fair enough to change it since that monstrosity is way too long anyway (and factor says nothing important to the user).

miscco:
As another point, i have troubles understanding the getAndRescaleQuivers function, which is closely related to this PR

Why is the length of the quivers scaled wrt the axis range, is this due to pgfplots?

Can XData have a size not equal [n, 1]?, the any(numel(x)==1) seems dubious

What is the relation between the range of the data to the number of arrows?

Wouldnt it be just sufficient to simply scale u, v, w with getOrDefault(h,'AutoScaleFactor', 0.9)?

src/matlab2tikz.m
egeerardyn: The `\pgfplotspointmetatransformed` contains the length of each arrow, so it is scaled with the arrow length due to
```matlab
    if is3D
        quiverOpts = opts_add(quiverOpts, 'w', '\thisrow{w}');
        arrowLength = '{sqrt((\thisrow{u})^2+(\thisrow{v})^2+(\thisrow{w})^2)}';
    else
        arrowLength = '{sqrt((\thisrow{u})^2+(\thisrow{v})^2)}';
    end

    plotOpts = opts_add(plotOpts, 'point meta', arrowLength);
```
so everything does scale with the arrow size.

egeerardyn:
@miscco: to understand the `getAndRescaleQuivers` functions, you have to look at the `quiver` and `quiver3` function in MATLAB as indicated in the comments of that file. We just re-implemented their scaling (hence the "weird" code). If there is a functional difference between our implementation and theirs, that might be a bug on our side.

The reason behind all that rescaling is that the raw `{X|Y|Z}Data` and `{U|V|W}Data` are in data units. However, the quivers are typically drawn such that each quiver fits within a single grid rectangle (or cube in 3D) defined by the data, hence the rescaling.

src/matlab2tikz.m
```diff
@@ -4005,22 +4010,54 @@ function warnFacetedInterp(m2t)
...
+        userInfo(m2t, ['Please change the "arrowHeadSize" option', ...
+                       ' if the size of the arrows is incorrect.']);
+        arrowHeadSize = sprintf(m2t.ff, abs(m2t.cmdOpts.Results.arrowHeadSize));
+
+        % Write out the actual scaling for TikZ.
+        % `\pgfplotspointsmetatransformed` is in the range [0, 1000], so
+        % divide by this span (as is done in the pgfplots manual) to normalize
+        % the arrow head size.
+        arrowHeadOpts = opts_add(arrowHeadOpts, 'scale', ...
+              ['{' arrowHeadSize '*\pgfplotspointmetatransformed/1000}']);
```
miscco: So out of curriosity. If \pgfplotspointmetatransformed ranges from [0, 1000] and we scale it to [0, 1]. Wouldnt that mean, that a arrowHeadSize ~= 10 / 1000 = 0.01 which is 20 times smaller than the 0.2 in matlab?

Could you check the difference when setting this to 200?

miscco:
Yeah, the code seems much more clearer now with the comments etc arranged. Sadly the quiver.m fucntion in my MATLAB2015b had no information about their scaling.

```matlab
function hh = quiver(varargin)
%QUIVER Quiver plot.
%   QUIVER(X,Y,U,V) plots velocity vectors as arrows with components (u,v)
%   at the points (x,y).  The matrices X,Y,U,V must all be the same size
%   and contain corresponding position and velocity components (X and Y
%   can also be vectors to specify a uniform grid).  QUIVER automatically
%   scales the arrows to fit within the grid.
%
%   QUIVER(U,V) plots velocity vectors at equally spaced points in
%   the x-y plane.
%
%   QUIVER(U,V,S) or QUIVER(X,Y,U,V,S) automatically scales the
%   arrows to fit within the grid and then stretches them by S.  Use
%   S=0 to plot the arrows without the automatic scaling.
%
%   QUIVER(...,LINESPEC) uses the plot linestyle specified for
%   the velocity vectors.  Any marker in LINESPEC is drawn at the base
%   instead of an arrow on the tip.  Use a marker of '.' to specify
%   no marker at all.  See PLOT for other possibilities.
%
%   QUIVER(...,'filled') fills any markers specified.
%
%   QUIVER(AX,...) plots into AX instead of GCA.
%
%   H = QUIVER(...) returns a quivergroup handle.
%
%   Example:
%      [x,y] = meshgrid(-2:.2:2,-1:.15:1);
%      z = x .* exp(-x.^2 - y.^2); [px,py] = gradient(z,.2,.15);
%      contour(x,y,z), hold on
%      quiver(x,y,px,py), hold off, axis image
%
%   See also FEATHER, QUIVER3, PLOT.

%   Clay M. Thompson 3-3-94
%   Copyright 1984-2014 The MathWorks, Inc.

[~, cax, args] = parseplotapi(varargin{:},'-mfilename',mfilename);
narginchk(2,inf);
% Parse remaining args
try
    pvpairs = quiverparseargs(args);
catch ME
    throw(ME)
end

if isempty(cax) || ishghandle(cax,'axes')
    cax = newplot(cax);
    % just create a new axes for new
    % until newplot is working so that 
    % cla(ax, 'reset',hsave); doesn't wipe out the default
    % shaped arrays like ColorOrder and LineStyleOrder
    % cax = gca; cla(cax);
    
    parax = cax;
    hold_state = ishold(cax);
else
    parax = cax;
    cax = ancestor(cax,'Axes');
    hold_state = true;
end
[ls,c] = nextstyle(cax);

h = matlab.graphics.chart.primitive.Quiver;
set(h,'Parent',parax,'Color_I',c,'LineStyle_I',ls,pvpairs{:});

% call CreateFcn explicitly immediately following object
% creation from the M point of view.
% this was the last line of @Quiver/Quiver.m
% There is no obvious place to have the CreateFcn 
% automatically executed on the MCOS side, so we call it here
h.CreateFcn;

if ~hold_state, box(cax,'on'); end

if nargout>0, hh = h; end
```

egeerardyn:
Indeed, in R2015b, the interesting parts seem to be p-coded (see the `matlab.graphics.primitive.Quiver` stuff), what a pity.

src/matlab2tikz.m
egeerardyn: The units are different in both as far as I understand. If you fetch this PR into your own repository, i.e.
```bash
git remote add egeerardyn git@github.com:egeerardyn/matlab2tikz.git
git fetch egeerardyn
git checkout egeerardyn/nativeQuivers-improved
```
you can try for yourself. My experience with that factor is that it works linearly, i.e. if you change the `10` to `200`, the arrow heads will be huge, not more realistic.

src/matlab2tikz.m
```diff
@@ -4005,22 +4010,54 @@ function warnFacetedInterp(m2t)
...
+        % is defined in pgfplots in absolute units (here we specify that those
+        % should be scaled up/down according to the data), while the data itself
+        % is in axis coordinates (or some scaled variant). I.e. we need the
+        % physical dimensions of the axis to compute the correct scaling!
+        % 
+        % There is a "MaxHeadSize" property that plays a role.
+        % MaxHeadSize is said to be relative to the length of the quiver in the
+        % MATLAB documentation. However, in practice, there seems to be a SQRT
+        % involved somewhere (e.g. if u.^2 + v.^2 == 2, all MHS values >
+        % 1/sqrt(2) are capped to 1/sqrt(2)).
```
miscco: I believe, that the cap you see is there to prevent the length off the head to exeed that of the shaft

egeerardyn: It's a bit long ago for me, you may be very right on the intent, but the way is a bit weird: note that it seems to be capped to 1/sqrt(L) where L is the length of the quiver. Anyhow, in this case I would say that the intent is one thing, re-implementing the behavior is another thing (since I'm not convinced on the relationship).

egeerardyn:
Test output (for convenience) for the failing cases:
## Test summary
Test results for m2t commit e7a3936de0 running with MATLAB 8.4 on Linux 4.0.7-2-ARCH.

```matlab
suite = @ACID;
alltests = [1:99];
reliable = [2:5 7:9 11:17 19:23 25:33 35:37 39 41 44:47 50:57 59:93 95:96 98:99];
unreliable = [1 6 10 18 24 34 38 40 42:43 48:49 58 94 97];
failReliable = [15];
passUnreliable = [49];
skipped = [17];
```

| Unreliable |    1 |   14 |    0 |    15 |
|   Reliable |   82 |    1 |    1 |    84 |
|      Total |   83 |   15 |    1 |    99 |

Build fails with 1 errors. :heavy_exclamation_mark:

But I have the impression I saw the same test being unreliable on my end.

miscco:
Hm i dont really undertsand how ACID 15 can fail, if it doesnt even touch quivers.

miscco:
From my side, the PR definitely improves the code and helps with fleshing out this rather special feature.

The regression in Octave is not that critical, given one can always use a larger HeadSize.

Maybe you can change the default to 20?

miscco:
I have the feeling, that the merge of develop kind of screwed something up. At least in the commit, there are some strange things like incorrect whitespaces etc. Can you check that.

egeerardyn:
Well, I merged `develop` back into this PR, so this PR also has the latest changes of develop. Moreover, I fixed the hashes in a dirty way: by re-running the tests on the merged code. That's also the reason for the whitespace shenanigans: the whitespace fixing was in develop but not yet in this PR. I will fix it at the end (because whitespace is a pain to merge most of the time).

For the octave regression, I can just make the tinker factor different in Octave and MATLAB in our testing suite. It's dirty, but it should do the trick for the little while that we have this factor.

miscco:
To alleviate the pain of merging, would it help to run whitespace format before you merge develop into the branch?

egeerardyn:
I just forced-pushed the commit before the merge and will redo the merge. Because you are right: most of this branch was removed and there were mostly faulty things happening.

Reverting the merge is not a viable option (see `man git-revert`, Git Pro book and [referenced docs](https://github.com/git/git/blob/master/Documentation/howto/revert-a-faulty-merge.txt)).

miscco:
Could it be, that you have to rebase onto develop, as the changes there are more recent than those in this PR and take precedence?

okomarov:
For the Octave case, we had it working well with the previous settings, so it shouldn't be hard to fall back to whatever we were using there.

egeerardyn:
Hopefully, this keeps Travis and Jenkins happy. I think the merge succeeded this time. But intermittently, I f*'ed it up as well. So I think I need to use another merge tool (I stepped through the changes with DeltaWalker, but I guess I should go back to either SmartSynchronize or Kdiff3).

egeerardyn:
I have adapted the scaling factor for that test in Octave. It now looks like:
screen shot 2015-12-14 at 23 02 26

on Octave 3.8.2.

Do note the red and blue arrows in the gnuplot render. This is by design: `ACID(22)` now plots 2 quivers on top of each other with different `MaxHeadSize`, but that property is currently not handled at all.

Anyhow, this moves the merging out of the way already (whitespace changes are a pest). In fact, the problem is when the base commit (i.e. where develop and  PR start to differ) does not have the whitespace changes, this makes 3-way merging quite an ordeal. The easiest way I found was to direct the 3-way merge tool I used to ignore whitespace and after merging use the `formatWhitespace` tool. But this workflow did introduce some parts of the code that were `=`-aligned to become unaligned. Anyhow: I think that's a small price to pay.

egeerardyn:
Okay, I'm fed up with this for the time being.

However, I am taking the liberty to merge this right already:
 - The Octave test cases look okay now,
 - Our support for quiver plots has improved since master (arrow heads resize now, before it were heads of a constant small size).

So as far as I'm concerned, the overall feature is an improvement over our last version and hence releasable.

PR #783, opened 2015-10-02
Octave hashtable

egeerardyn:
This is a follow-up on #720 to address the comments of @okomarov.

We really need this in, since `saveHashtable` in Octave was broken (I only tested the previous PR with MATLAB, my bad!).


PeterPablo:
Code looks good, though didn't runtime test. Sorry...
:+1:

egeerardyn:
Thanks for reviewing. I did the runtime testing here (Octave 3.8 and MATLAB 2014 (don't remember A or B, but since it doesn't interface with HG, I think it is unlikely to matter).

PR #787, opened 2015-10-09
Cleanfigure restructure

miscco:
So when I was writing the 3D projection part in #779, I realized, that I had to do a lot of restructuring. 

To keep it "cleaner" I decided to pull it out and make a new PR. So there are some new functions:

```matlab
[xData, yData] = getVisualData(meta, handle)
[xLim, yLim] = getVisualLimits(meta)

replaceData(handle, id_replace, dataReplace)
insertData(handle, id_insert, dataInsert)
removeData(meta, handle, id_remove)

removeNaNs(meta, handle)
```

I believe their names are rather self explanatory. To get them working i had to restructure our simplification functions slightly. In the end I decided, that every function should generate lists of indices e.g. id_replace for all the data that should be replaced and removed and so on.

This unifies the data handling and should help us to maintain the stuff easier.

I have also renamed some of the output when it was a mask rather out

### DIFF
--- a/src/cleanfigure.m
+++ b/src/cleanfigure.m
@@ -181,23 +181,19 @@ end
 function pruneOutsideBox(meta, handle)
   % Some sections of the line may sit outside of the visible box.
   % Cut those off.
-
-  xData = get(handle, 'XData');
-  yData = get(handle, 'YData');
-
-  % Obtain zData, if available
-  if isprop(handle, 'ZData')
-    zData = get(handle, 'ZData');
-  else
-    zData = [];
+  
+  % TODO: 3d simplificattion of frontal 2d projection
+  if isAxis3D(meta.gca)
+      return;
   end
 
-  if isempty(zData)
-    data = [xData(:), yData(:)];
-  else
-    data = [xData(:), yData(:), zData(:)];
-  end
+  % Extract the visual data from the current line handle.
+  [xData, yData] = getVisualData(meta, handle);
+  
+  % Merge the data into one matrix
+  data = [xData(:), yData(:)];
 
+  % Dont do anythingif the data is empty
   if isempty(data)
       return;
   end
@@ -206,8 +202,9 @@ function pruneOutsideBox(meta, handle)
   %hasMarkers = ~strcmp(marker,'none');
   hasLines = true;
   hasMarkers = true;
-  xLim = get(meta.gca, 'XLim');
-  yLim = get(meta.gca, 'YLim');
+  
+  % Extract the visual limits from the current line handle.
+  [xLim, yLim]   = getVisualLimits(meta);
 
   tol = 1.0e-10;
   relaxedXLim = xLim + [-tol, tol];
@@ -216,8 +213,7 @@ function pruneOutsideBox(meta, handle)
   numPoints = size(data, 1);
 
   % Get which points are inside a (slightly larger) box.
-  dataIsInBox = isInBox(data(:,1:2), ...
-                        relaxedXLim, relaxedYLim);
+  dataIsInBox = isInBox(data, relaxedXLim, relaxedYLim);
 
   % By default, don't plot any points.
   shouldPlot = false(numPoints, 1);
@@ -226,8 +222,7 @@ function pruneOutsideBox(meta, handle)
   end
   if hasLines
       % Check if the connecting line is in the box.
-      segvis = segmentVisible(data(:,1:2), ...
-                              dataIsInBox, xLim, yLim);
+      segvis = segmentVisible(data, dataIsInBox, xLim, yLim);
       % Plot points which are next to an edge which is in the box.
       shouldPlot = shouldPlot | [false; segvis] | [segvis; false];
   end
@@ -270,10 +265,6 @@ function pruneOutsideBox(meta, handle)
   % Override with the new data.
   set(handle, 'XData', data(:, 1));
   set(handle, 'YData', data(:, 2));
-  if ~isempty(zData)
-    set(handle, 'ZData', data(:, 3));
-  end
-
   return;
 end
 % =========================================================================
@@ -282,31 +273,30 @@ function movePointsCloser(meta, handle)
   % to the boundary of that box and make sure that lines in the visible
   % box are preserved. This typically involves replacing one point by
   % two new ones and a NaN.
-
-  % Extract the data from the current line handle.
-  xData = get(handle, 'XData');
-  yData = get(handle, 'YData');
-  zData = get(handle, 'ZData');
-
-  if ~isempty(zData) && any(zData(1)~=zData)
-    % Don't do funny stuff with varying zData.
-    return;
-  end
   
-  data = [xData(:), yData(:)];
+  % TODO: 3d simplificattion of frontal 2d projection
+  if isAxis3D(meta.gca)
+      return;
+  end
 
-  xlim = get(meta.gca, 'XLim');
-  ylim = get(meta.gca, 'YLim');
+  % Extract the visual data from the current line handle.
+  [xData, yData] = getVisualData(meta, handle);
+  
+  % Extract the visual limits from the current line handle.
+  [xLim, yLim]   = getVisualLimits(meta);
 
   % Calculate the extension of the extended box  
-  xWidth = xlim(2) - xlim(1);
-  yWidth = ylim(2) - ylim(1);
+  xWidth = xLim(2) - xLim(1);
+  yWidth = yLim(2) - yLim(1);
   
   % Don't choose the larger box too large to make sure that the values inside
   % it can still be treated by TeX.
   extendFactor = 0.1;
   largeXLim = xlim + extendFactor * [-xWidth, xWidth];
   largeYLim = ylim + extendFactor * [-yWidth, yWidth];
+  
+  % Put the data into one matrix
+  data = [xData, yData];
 
   % Get which points are in an extended box (the limits of which
   % don't exceed TeX's memory).
@@ -320,17 +310,17 @@ function movePointsCloser(meta, handle)
   replaceIndices = find(~dataIsInLargeBox);  
   
   % Only try to replace points if there are some to replace
-  if (~isempty(replaceIndices))
+  if ~isempty(replaceIndices)
       % Check whether the points are finite.
-      DataIsFinite = all(isfinite(data(replaceIndices,:)), 2);
+      dataIsFinite = all(isfinite(data(replaceIndices,:)), 2);
 
       % Get the indices of those points, that are the first point in a
       % segment
-      id_first  = replaceIndices(replaceIndices < size(data, 1) & DataIsFinite);
+      id_first  = replaceIndices(replaceIndices < size(data, 1) & dataIsFinite);
 
       % Get the indices of those points, that are the second point in a
       % segment
-      id_second = replaceIndices(replaceIndices > 1  & DataIsFinite);
+      id_second = replaceIndices(replaceIndices > 1  & dataIsFinite);
 
       % Define the vectors of data points for the segments X1--X2
       X1_first  = data(id_first,    :);
@@ -353,7 +343,7 @@ function movePointsCloser(meta, handle)
       % If a point is part of two segments, that cross the border, we need to
       % insert a NaN to prevent an additional line segment
       [trash, trash, id_conflict] = intersect(id_first (~isInfinite_first), ...
-                                      id_second(~isInfinite_second));
+                                      id_second(~isInfinite_second)); %#ok<ASGLU>
 
       % Cut the data into length(replaceIndices)+1 segments.
       % Calculate the length of the segments
@@ -410,12 +400,6 @@ function movePointsCloser(meta, handle)
   % Set the new (masked) data.
   set(handle, 'XData', data(:, 1));
   set(handle, 'YData', data(:, 2));
-  if ~isempty(zData)
-    % As per precondition, all zData entries are equal.
-    zDataNew = zData(1) * ones(size(data, 1), 1);
-    set(handle, 'zData', zDataNew);
-  end
-
   return;
 end
 % =========================================================================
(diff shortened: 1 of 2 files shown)


src/cleanfigure.m
```diff
@@ -251,78 +250,161 @@ function pruneOutsideBox(meta, handle)
...
+      % cell at the beginning/end
+      if(id_replace(end) == size(data, 1))
+        dataInsert_first  = [dataInsert_first; cell(1)];
+      end
+      if(id_replace(1) == 1)
+        dataInsert_second = [cell(1); dataInsert_second];
+      end
+
+      % Put the cells together, right points first, then the possible NaN
+      % and then the left points
```
miscco: I am kind of unhappy about this. Does somebody have a better idea how to merge the 3 cell arrays?

src/cleanfigure.m
```diff
@@ -251,78 +250,161 @@ function pruneOutsideBox(meta, handle)
...
+
+  % Insert the data
+  insertData(handle, id_replace, dataInsert);
   
-  % Check whether lambda is in bound
-  out = 0.0 < lambda(:, 1) & lambda(:, 1) < 1.0 &...
-        0.0 < lambda(:, 2) & lambda(:, 2) < 1.0;
+  % Get the indices of the to be removed points respecting the now inserted
+  % data points
+  id_remove = id_replace + cumsum(cellfun(@(x) size(x,1), [cell(1);dataInsert(1:end-2)]));
```
miscco: Same here, I don't like the cellfun shenanigans. some easier way to track things? One could track them individually i guess.

src/cleanfigure.m
```diff
@@ -251,78 +250,161 @@ function pruneOutsideBox(meta, handle)
...
+
+      % Create a cell array for the moved points
+      dataInsert_first  = mat2cell(newData_first,  ones(size(id_first)),  2);
+      dataInsert_second = mat2cell(newData_second, ones(size(id_second)), 2);
+
+      % Add an empty cell at the end of the last segment
+      dataInsert_first  = [dataInsert_first;  cell(1)];
+      dataInsert_second = [dataInsert_second; cell(1)];
+
+      % If the first or the last point would have been replaced add an empty
```
egeerardyn: Why?

src/cleanfigure.m
```diff
@@ -251,78 +250,161 @@ function pruneOutsideBox(meta, handle)
...
+      % conflict sites
+      dataInsert_NaN              = cell(length(length_segments),1);
+      dataInsert_NaN(id_conflict) = mat2cell(NaN(length(id_conflict), 2),...
+                                             ones(size(id_conflict)), 2);
+
+      % Create a cell array for the moved points
+      dataInsert_first  = mat2cell(newData_first,  ones(size(id_first)),  2);
+      dataInsert_second = mat2cell(newData_second, ones(size(id_second)), 2);
+
+      % Add an empty cell at the end of the last segment
```
egeerardyn: Why?

The comments should explain *why* you do something (or why it has to be done in that particular way), not *what* you do. The latter is what the code should convey.

src/cleanfigure.m
```diff
@@ -251,78 +250,161 @@ function pruneOutsideBox(meta, handle)
...
+  insertData(handle, id_replace, dataInsert);
   
-  % Check whether lambda is in bound
-  out = 0.0 < lambda(:, 1) & lambda(:, 1) < 1.0 &...
-        0.0 < lambda(:, 2) & lambda(:, 2) < 1.0;
+  % Get the indices of the to be removed points respecting the now inserted
+  % data points
+  id_remove = id_replace + cumsum(cellfun(@(x) size(x,1), [cell(1);dataInsert(1:end-2)]));
+  
+  % Remove the data point to be replaced. This has to happen fist, as it is
```
egeerardyn: fist -> first

src/cleanfigure.m
```diff
@@ -251,78 +250,161 @@ function pruneOutsideBox(meta, handle)
...
+  % not clear, hom many points are to be inserted
+  % Remove the data outside the box
+  removeData(meta, handle, id_remove);
+
+  % Remove possible NaN duplications
+  removeNaNs(meta, handle);
+  return;
```
egeerardyn: No need to call `return` here

src/cleanfigure.m
miscco: I am kind of confused, that is the old code, that is just moved around.

src/cleanfigure.m
miscco: Same as above, that is the "old" code

src/cleanfigure.m
```diff
@@ -743,89 +752,198 @@ function movePointsCloser(meta, handle)
...
+  dataCell = [dataCell';
+              dataInsert'];
+
+  % Merge the cells back together
+  data     = cat(1, dataCell{:});
+
+  % Set the new (masked) data.
+  set(handle, 'XData', data(:,1));
+  set(handle, 'YData', data(:,2));
+  return;
```
egeerardyn: No need for `return`

src/cleanfigure.m
miscco: Ah I forgot to update the comment. Fixed locally

src/cleanfigure.m
```diff
@@ -743,89 +752,198 @@ function movePointsCloser(meta, handle)
...
-      % As we have transposed rhs we can write this as:
-      % rhs * Rotate * (X4-X3) => [nx2] * [2x2] * [2x1] = [nx1]
-      lambda(id_detA, 1) = (rhs(id_detA, :) * Rotate * (X4-X3))./detA(id_detA);
+  % Set the new data.
+  set(handle, 'XData', xData);
+  set(handle, 'YData', yData);
+  if meta.is3D
+  	set(handle, 'zData', zData);
+  end
+  return;
```
egeerardyn: Remove `return`

src/cleanfigure.m
```diff
@@ -743,89 +752,198 @@ function movePointsCloser(meta, handle)
...
+  % Remove the NaNs
+  data(id_remove,:) = [];
+
+  % Set the new data.
+  set(handle, 'XData', data(:,1));
+  set(handle, 'YData', data(:,2));
+  if meta.is3D
+  	set(handle, 'zData', data(:,3));
   end
+  return;
```
egeerardyn: Remove `return`

src/cleanfigure.m
```diff
@@ -96,7 +96,8 @@ function cleanfigure(varargin)
 
   % Update the current axes.
   if strcmp(get(h, 'Type'), 'axes')
-      meta.gca = h;
+      meta.gca  = h;
+      meta.is3D = isAxis3D(h);
```
egeerardyn: Is there a significant overhead for repeatedly calling the `isAxis3D` function? If not, I would opt to NOT cache its output as you do here.

miscco: Nothing except personal laziness

src/cleanfigure.m
egeerardyn: @miscco I only noticed it was old code after a while. Sorry.

src/cleanfigure.m
miscco: Yeah, i saw it in some function and just plastered it everywhere. Removed locally.

miscco:
So I don't know your preference, if I push my fixes now, then your comments get hidden. I would prefer to wait a bit till I push them, so anyone else can have a look at them.

src/cleanfigure.m
miscco: @egeerardyn Don't worry, it doesn't hurt to have better comments. I actually had to think about that line for a minute.

src/cleanfigure.m
egeerardyn: I don't think the `cellfun` is unclear here. What you could do to improve readability, is to add some intermediate variables, e.g.
```matlab
numberOfPointsToInsert = cellfun(@(x) size(x,1), dataInsert(1:end-2));
id_remove = id_replace + cumsum([1, numberOfPointsToInsert]);
```
Note: I didn't actually run this snippet.

src/cleanfigure.m
egeerardyn: I don't think it's really dirty, but @okomarov tends to be a genius when it comes to avoiding `cellfun` (personally, I rather like to use `cellfun` and so on).

Anyhow, the only slightly misleading thing I see, are the variable names `x,y,z`; maybe it's better to use `a,b,c`? But it's nothing major.

Is there a reason why you can't just concatenate the arrays in the normal way as was the case before?

src/cleanfigure.m
```diff
@@ -206,8 +203,9 @@ function pruneOutsideBox(meta, handle)
   %hasMarkers = ~strcmp(marker,'none');
   hasLines = true;
```
miscco: BTW, i always stumble upon those lines. Can't we just remove those lines and the checks altogether?

egeerardyn:
Thank you for your hard work on this. I will try to have a more in-depth look at your code this evening or tomorrow. What certainly still needs to be done is a quick look at why those tests are failing. Normally, you should be able to download the compiled files from the Jenkins server.

Feel free to push your changes to this PR. Once you address my comments, I think it makes the most sense to no longer see them (that way, we can keep track of what still has to be done).

src/cleanfigure.m
miscco: Hm in the old code it was easier as I directly transformed that into a simple matrix. Here it is more difficult due to the new functions.

I could just merge the arrays and pass them to dataInsert, however it would be more difficult to get the number of inserted data points

src/cleanfigure.m
egeerardyn: To be honest, I have no clue why those lines are commented out. This seems like somebody went debugging in there and forgot to switch them back. We should `git blame` those to see what happened.

miscco:
I believe i might have found the culprit:
```
\addplot [color=green,solid,mark=asterisk,mark options={solid},forget plot]
  table[row sep=crcr]{%
 nan	 nan\\
};
```

Previously pure nan sets were deleted. I have no idea why it didnt hapen here

src/cleanfigure.m
```diff
@@ -743,89 +752,198 @@ function movePointsCloser(meta, handle)
...
+  yData = get(handle, 'YData');
+  if meta.is3D
+  	zData = get(handle, 'ZData');
+    data  = [xData(:), yData(:), zData(:)];
+  else      
+    data  = [xData(:), yData(:)];
+  end
+
+  % Remove consecutive NaNs
+  id_nan   = any(isnan(data),2);
```
egeerardyn: Nitpicking mode: I think those `=`s can do with a space before and some of the `,` could stand a space after.

miscco:
So the only tests that fail now are the two quiverplots in Octave. However, i am lost whats wrong with them. At least in my Octave4 I cannot find a differrence.

miscco:
Soo, looking at the output i find the following differences:

Test24: 
```
diff Test_24_orig Test_24_new 
64a65
> nan   nan     nan\\
110a112
> nan   nan     nan\\
```

The NaNs are at the end of the data files and should come from me being more strict with 3D data.

Test23:
```
diff Test_23_orig Test_23_new 
21,22c21,22
< height=3.945in,
< at={(0.837in,0.532in)},
---
> height=3.935in,
> at={(0.837in,0.531in)},
2527d2526
< nan   nan\\
3164a3164
> 1.79899415212222      -1.99814575077236\\
3243a3244
> 1.79899415212222      1.99814575077236\\
3246a3248
> 1.99921747738733      -1.99903647765436\\
3249a3252
> 1.99832675050533      -1.79863688077976\\
3320a3324
> 1.99832675050533      1.79863688077976\\
3323a3328
> 1.99921747738733      1.99903647765436\\
```

The first change is the height/width, where i have no clue.

The second change is a missing NaN data point. I am completely lost, why it is missing. I have veryfied that it is not removed by removeNaNs().

Also the additional datapoints seem kind of strange. I dont really get why they are inserted.

miscco:
After thinking a bot more about it I belive, this is again a precision error in octave. All the points that are additionally there are atthe very edge of the image. That leads to the increased size oft he figure aswell. Where they come from, i have no clue however. It might be that pruneOutsideBox gives us a different behavior?

src/cleanfigure.m
```diff
@@ -743,89 +750,200 @@ function movePointsCloser(meta, handle)
...
+  % Set the new data.
+  set(handle, 'XData', xData);
+  set(handle, 'YData', yData);
+  if isAxis3D(meta.gca)
+  	set(handle, 'zData', zData);
+  end
+end
+% =========================================================================
+function removeNaNs(meta, handle)
+  % Removes superflous NaNs in the data
```
okomarov: Should elaborate on how it differs from `data(isnan(data)) = [];`. If I remember correctly the function collapses streaks of NaNs to a single NaN and removes NaNs at the ends of a line.

src/cleanfigure.m
```diff
@@ -342,109 +422,124 @@ function simplifyLine(meta, handle, targetResolution)
...
-    end
-    xData = [linesx{1:end-1}];
-    yData = [linesy{1:end-1}];
+  mask = data(:,1) > xLim(1) & data(:,1) < xLim(2) ...
+       & data(:,2) > yLim(1) & data(:,2) < yLim(2);
+end
+% =========================================================================
+function mask = segmentVisible(data, dataIsInBox, xLim, yLim)
+    % Given a bounding box {x,y}Lim, determine whether the line between all 
+    % pairs of subsequent data points crosses the box.
```
okomarov: A bit cryptic, I would suggest a more generic H1 line, the details are left inside the code

okomarov:
I like the idea of having an API to deal with data insert/replace/remove. However, it is a bit dangerous the way the positions are passed around, in the sense that each data operation alters the positions recorded in `id_*`. I am not sure what's the best practice but maybe we could make it more explicit by actually passing(returning) the data to(from) the function so that the get and set calls are visible from the parent function.

src/cleanfigure.m
miscco: That again the old comment ;)

egeerardyn:
@miscco The different hashes are fine. I cannot visually see a difference.

On my Mac (Octave 3.8.2), I produced the following report:

# Octave 3.8.2 (Mac OS 10.11) commit 158a43632e

## Reliable tests
Only the reliable tests determine the build outcome.
Passing tests are not shown (only failed and skipped tests).

|   Testcase |          Name |                       OK |                                                                                   Status |
| :--------- | :------------ | :----------------------: | :--------------------------------------------------------------------------------------- |
| `ACID(23)` |  `quiverplot` | :heavy_exclamation_mark: | new hash 6e2d74dd2d73bfa916c571bf704e360b != expected (0cc7aef48dddc2c9075a079eaf580700) |
| `ACID(24)` | `quiver3plot` | :heavy_exclamation_mark: | new hash 8d74c782c57b27be844585d37328d3cf != expected (1e2266cae37d6565e96e1f0466540d6c) |

## Test summary
Test results for m2t commit 158a43632e running with Octave 3.8.2 on Mac OS 10.11.

```matlab
suite = @ACID;
alltests = [23:24];
reliable = [23:24];
unreliable = [];
failReliable = [23:24];
passUnreliable = [];
skipped = [];
```

| Unreliable |    0 |    0 |    0 |     0 |
|   Reliable |    0 |    2 |    0 |     2 |
|      Total |    0 |    2 |    0 |     2 |

Build fails with 2 errors. :heavy_exclamation_mark:

That seems to match with Linux, so feel free to overwrite the hashes :+1:.

src/cleanfigure.m
miscco: No this would remove *all* NaNs. However, we want only those at the beginning, the end and those that are doubled

src/cleanfigure.m
```diff
@@ -96,7 +96,7 @@ function cleanfigure(varargin)
```
miscco: BTW how come, that we have this check against quiverplots? It seems fairly ineffective.

miscco:
@okomarov I explicitely put the data setter into those functions. I believe that is actually safer, as it is clear which functions tamper with the id*s. Maybe i should add a comment at the beginning of insertData and remove* that they might turn previous indices invalid?

miscco:
As a second note, those function render the 3D case rather easy.

src/cleanfigure.m
```diff
@@ -743,89 +750,200 @@ function movePointsCloser(meta, handle)
...
-  %
-  % returns the lambda for which they intersect (and Inf if they are parallel).
-  % Technically, one needs to solve the 2x2 equation system
-  %
-  %   x1 + lambda1 (x2-x1)  =  x3 + lambda2 (x4-x3)
-  %   y1 + lambda1 (y2-y1)  =  y3 + lambda2 (y4-y3)
-  %
-  % for lambda1 and lambda2.
+function replaceData(handle, id_replace, dataReplace)
+  % Replaces data at id_replace with dataReplace
```
miscco: As a matter of fact, we only ever replace data points by NaNs. So one might change this to
replaceData(handle, id_replace), or even more oscure
replaceDataWithNaN(handle, id_replace)

The current function is more flexible and can potentially be used sometimes later when the need arises.

src/cleanfigure.m
egeerardyn: You mean the cleanfigure around every test case? Indeed that's quite odd, but it has helped to find bugs in cleanfigure before.

miscco: Ah it cut the lines, here is what i meant:
```matlab
   if strcmp(get(h, 'Type'), 'specgraph.quivergroup')
   %if strcmp(class(handle(h)), 'specgraph.quivergroup')
       return;
   end
```

src/cleanfigure.m
egeerardyn: I traced it back to ce899382ce6c0edfc69bc415cf0b769dce0f575f which is almost 2 years old. The commit message suggests it was copied in from the main matlab2tikz code (back when `cleanfigure` still was a part of it), but I can't find the corresponding part in that file right now.

For this case, I'd say: just put a TODO above those lines:
```matlab
%TODO: test again with lines uncommented (see #787 and ce899382)
```

miscco: I opened an Issue for that. I would say, that this is unrelated to the restructure here, and i would prefer to get it landed rather soon, as it is a prerequisite for 3D handling.

egeerardyn: :+1:

egeerardyn:
As far as I am concerned, we can merge this. Thank you for your hard work on this!

I do think that @okomarov has good a point in https://github.com/matlab2tikz/matlab2tikz/pull/787#issuecomment-147101001, but this is something we can address in the future when we know better what we want from that internal API.

okomarov:
nice job :+1:

PR #790, opened 2015-10-12
Utilze 2D projection to handle 3D plots

miscco:
This is a rebased version of my Project3D PR #778. Regarding the viewmtx issue I went went with simply using the appropriate 3D projections and not bothering with anything else. This should make it easier to understand.

However, there is one drawback. This PR does not include the back transformation, that would be required for movePointsCloser. This is a topic for itself. my current guess for an implementation would be:
1. Project into 2D
2. Move the points closer to the box.
3. Define the line between the original points and their projection
4. Move the new points along that line so that it is closest to the original one.


miscco:
Hm, so far so good, only the 3D plot ACID(70) differs and quiverplot3D.

However, there is a visual difference in ACID(70). I am going to check why

miscco:
Edit, something is strange, i dont know why but it seems the projection is wrong. I have to dig into that.

src/cleanfigure.m
```diff
@@ -799,30 +845,87 @@ function simplifyLine(meta, handle, targetResolution)
...
+	% Get the projection angle
+    [az, el] = view(meta.gca);
+
+    % Convert from degrees to radians.
+    az = az*pi/180;
+    el = el*pi/180;
+
+    % The transformation into the image plane is done i a two-step process.
+    % First rotate around the z-axis by -az (in radians)
+    % Second rotate around the x-axis by (el -pi/2) radians.
```
egeerardyn: Nice to see a lot of comments, however, I would place the `Second, rotate ...` more towards the bottom. You can even make that into a cell comment (i.e. ``%% Rotate around x-axis over (el - pi/2) radians`)

src/cleanfigure.m
```diff
@@ -761,11 +753,17 @@ function simplifyLine(meta, handle, targetResolution)
...
+  % log_scaling and projection into the image plane)
+
+  % Check whether this is a 3D plot
+  is3D = isAxis3D(meta.gca);
 
   % Extract the data from the current line handle.
   xData = get(handle, 'XData');
   yData = get(handle, 'YData');
+  if is3D
+    zData = get(handle, 'ZData');
```
okomarov: 4 spaces indent, also below

okomarov:
Two minor comments, indentation is off, and `getVisualData` and `getVisualLimits` share 90% of the code.

For the first, you can `ctrl+i` the whole code in the matlab editor. For the second, if you feel like it, you could factor out the common part into a `projectDataToCameraView_`.

okomarov:
If you could follow up your comment in https://github.com/matlab2tikz/matlab2tikz/pull/790#issuecomment-147362338 with a snapshot of the plots that would be great. Sometimes it's enough for us to look at the plot to have an idea what's going on and give useful suggestions.

src/cleanfigure.m
```diff
@@ -799,30 +845,87 @@ function simplifyLine(meta, handle, targetResolution)
   if isYlog
     yLim  = log10(yLim);
   end
+  if is3D
+      isZlog = strcmp(get(meta.gca, 'ZScale'), 'log');
```
egeerardyn: Nitpicking mode: 

I think `ZScale` is always present on the axis, so there is little need to put it inside `if is3D, ..., end`, you could just define it in the same way as for `X` and `Y` (that would make it more obvious to refactor a bit later on).

The same goes for `ZLim`: that one is always defined (but the values, obviously, only make sense when the axis is three-dimensional).

I think that allows to keep the flow of the code a bit more linear and hence easier to follow for future readers. (This comes at the cost that now the workspace always has those Z-variables that don't always make sense).

src/cleanfigure.m
```diff
@@ -251,7 +246,7 @@ function pruneOutsideBox(meta, handle)
       id_remove  = id_remove(~idx);
   end
   % Replace the data points
-  replaceData(handle, id_replace, NaN(length(id_replace), size(data,2)));
```
egeerardyn: Is there a reason why you hoisted the `NaN` creation into the function? I think it makes more sense to have this generic `replaceData` function than to hardcode the function to replace with `NaN`. The new function is a function that you can implement by wrapping the old function with just 1 or 2 lines of new code.

miscco: So the reason is, that we actually only insert NaNs, and i would have to handle the 3D case outside of the helper function as that defines the dimension of the NaN matrix.

The size of Data is always 2 btw as it is the projected data.

egeerardyn: That's fine :smile:. 
*(I though we might have used that in our `pixelate` function as well, but it seems not, and probably it's not any easier to do so).*

src/cleanfigure.m
```diff
@@ -924,19 +1050,46 @@ function removeNaNs(meta, handle)
...
+  bottomRight = [xLim(2), yLim(1)];
+  topRight    = [xLim(2), yLim(2)];
+end
+% ==========================================================================
+function corners = corners3D(xLim, yLim, zLim)
+    % Determine the corners of the 3D axes as defined by xLim, yLim, and
+    % zLim
+
+    % Lower square of the cube
+    lowerBottomLeft  = [xLim(1), yLim(1), zLim(1)];
```
egeerardyn: I wonder if we couldn't write this in function of `corners2D`:

```matlab
function corners = corners3D(xLim, yLim, zLim)
% Determine the corners of the 3D axes as defined by (z/y/z)Lim

    % (X,Y)-coordinates of the corners
    [bottomLeft, topLeft, bottomRight, topRight] = corners2D(xLim, yLim);
    % Z-coordinate for lower/upper face of the cube
    lowerFace = zLim(1);
    upperFace = zLim(2);

    corners = [bottomLeft  lowerFace;
               topLeft     lowerFace;
               bottomRight lowerFace;
               topRight    lowerFace;
               bottomLeft  upperFace;
               topLeft     upperFace;
               bottomRight upperFace;
               topRight    upperFace];
end

Comment (miscco) [inline src/cleanfigure.m]: Not really. 

The problem is, that corners2D is called within the segmentVisible function. Therefore we would have to pass meta and handle all the way down.

Comment (miscco) [inline src/cleanfigure.m]: Also in a different note, it would considerably change the syntax corners2D is used. The only way around would be to have 8 outputs in corners3D which is kind of painfull.

**Comment by miscco** [comment]
So i think i found my error. The problem was the orientation of the azimuth. It should be passed as -az, which slightly changed the projection matrix and let my question my basic algebra skills ;)

**Comment by miscco** [comment]
Argh, So albeit it passes all tests i found that my own use cases now do not work any more. I really need to figure out why. 

In short, I have multiple segments plotted and some of them vanish. So i believe there might be some NaN problems.

**Comment by okomarov** [comment]
It would be good if you included your test case in the ACID.

**Comment by miscco** [comment]
Yeah, i will, as soon as i have found out, what is going wrong here.

File: src/cleanfigure.m (hunk 5 above)
Comment (egeerardyn) [inline src/cleanfigure.m]: @miscco I meant, call `corners2D` from within `corners3D` (as in my example above). That wouldn't change the syntax anywhere else, as far as I see, since we don't need `meta` or `handle` in these functions.

File: src/cleanfigure.m [hunk 6]
```diff
@@ -864,25 +934,28 @@ function removeData(meta, handle, id_remove)
       return
   end
 
+  % Check whether this is a 3D plot
+  is3D = isAxis3D(meta.gca);
+
   % Extract the data from the current line handle.
   xData = get(handle, 'XData');
```
egeerardyn: I'm wondering whether what is the optimal way to express this.
Currently:
```matlab
xData = get(h, 'XData');
yData = get(h, 'YData');
if is3D
  zData = get(h, 'ZData');
end

xData(idx) = [];
yData(idx) = [];
if is3d
  zData(idx) = [];
end

set(h, 'XData', xData);
set(h, 'YData', yData);
if is3d
  set(h, 'ZData', zData);
end
```

versus

```matlab
removePoints(handle, 'XData', idx);
removePoints(handle, 'YData', idx);
if is3d
  removePoints(handle, 'ZData', idx);
end

function removePoints(h, propertyName, idx)
    % deletes the (X,Y,Z)Data at certain indices
    data = get(h, propertyName);
    data(idx) = [];
    set(h, propertyName, data);
end
```

The latter is shorter (and less duplication of code), but I wonder whether it is readable enough. What do you think?

egeerardyn:
@miscco With respect to adding tests to `ACID`, you can also add new test suites specific for these kinds of features (see the other functions in the `suites` folder). If you are just testing small things one at a time, those are often a better fit than the `ACID` suite (but we currently don't test for regressions on them, which can be both a good thing or a bad thing).

miscco:
Ah i found one of the bugs ;), i passed corners3d(yLim, yLim, zLim)

miscco:
Ah and i found the second more subtle bug. It is in simplifyLine. I dont know how, but I failed to account for the correct number of data points in the previous segments. Especially as lineStart-1 give you exactly that, so there is no need to keep check of that.

okomarov:
@miscco Alternatively, fyi, you could save the logical index in a cell array and concatenate at the end. Then pass the logical array to removeData. 

The advantage of this approach is that if the final concatenated array does not match the length of the data, it is easier to spot the bug. And you also avoid a `find()` call, which deteriorates performance.

src/cleanfigure.m
miscco: Hm, so it is 17 vs 12 lines of code, so we would have won 5 lines. As much as i love encapsulation, i believe, that this is a bit too much.

src/cleanfigure.m
miscco: So i looked a bit at the two options and i have to say, that the original one seems preferable.

The point is, that we do not gain anything from that and it is easier to see, where those corner points actually come from.

miscco:
@okomarov This is true, however I believe, it is easier to read this way. Also for large plots, this could become rather costly in terms of working memory.

miscco:
So regarding additional ACID tests. I believe it would be useful to have different versions of parameterCurve3d.

The first one should have different axis limits, as the bug with wrong corners was otherwise not found.
```matlab
function [stat] = parameterCurve3d_A()
  stat.description = 'Parameter curve in 3D with different axis ranges';
  stat.issues = 790;

t = linspace(0,6*pi, 1e6);
 plot3(t, 100 * sin(t), cos(t));
end
```

The second one should segmentize the data into lets say 4 chunks of data and plot them.  
```matlab
function [stat] = parameterCurve3d_A()
  stat.description = 'Parameter curve in 3D with different axis ranges';
  stat.issues = 790;
range = [1e5, 2e5, 3e5, 4e5]; 
t = linspace(0,6*pi, 1e6);
t = mat2cell(t, 1, range);
plot3(t{1}, 100 * sin(t{1}), cos(t{1}),...
t{2}, 100 * sin(t{2}), cos(t{2}),...
t{3}, 100 * sin(t{3}), cos(t{3}),...
t{4}, 100 * sin(t{4}), cos(t{4}));
end
```

egeerardyn:
@misscco: I think memory usage is only very rarely an issue in our codebase. Those logicals will only take a byte per data point, while the actual data is 4 bytes per data point per dimension (=2 or 3). Moreover, we eventually convert everything to text, so this means if we use 15 digits + sign + period + delimiter = 17 to 18 bytes per data point per dimension (for ASCII encoding).

E.g. for a huge plot with 10M points in 3D, the data takes up 240MB, the logicals will take up 10MB and if all the points end up in the text that will be around 540MB (I tested in MATLAB, and it appears to be upward of 1GB, but that might be UTF-8 vs ASCII encoding):

```matlab
>> data = randn(3, 10e6);
>> log  = logical(ones(1, 10e6));
>> str  = sprintf('%0.15f',data); % beware: slow!
>> whos
  Name      Size                           Bytes  Class      Attributes

  data      3x10000000               240 000 000  double               
  log       1x10000000                10 000 000  logical              
  str       1x524999877            1 049 999 754  char                 
```

Conclusion: having those logicals is unlikely to be a bottleneck.

miscco:
Maybe i am just lost and in dire need of another coffee, but something just doesnt make sense.
So the code does kind of works:
```matlab
figure(1)
h = ezplot3('cos(t)', 'sin(t)', 'sin(5*t)', [-pi pi]);

data = get(h,{'XData','YData','Zdata'});
data = [cat(1,data{:})', ones(numel(data{1}),1)];

% Projection matrix on screen
[az,el] = view();
A = viewmtx(az,el);

data_transformed = A*data';
figure(2)
plot(data_transformed(1,:), data_transformed(2,:))  
```


But if I put the axis into different ranges then strange things happen: 
```matlab
figure(1)
h = ezplot3('t', 'sin(t)', '20*cos(t)', [0 10*pi]);

data = get(h,{'XData','YData','Zdata'});
data = [cat(1,data{:})', ones(numel(data{1}),1)];

% Projection matrix on screen
[az,el] = view();
A = viewmtx(az,el);

data_transformed = A*data';
figure(2)
plot(data_transformed(1,:), data_transformed(2,:)); 
```

Currently i am puzzled.

egeerardyn:
@miscco 
I am barely familiar with all those graphics calculations, but I have the impression that the `CameraPosition`, `CameraViewAngle`, ... also should play a role somehow.

When I read the `doc viewmtx` page, I have the impression that `view()` and `viewmtx(az, el)` should produce the same output, but they do not. In any case, neither transformation matrix produces the expected output.

okomarov:
http://stackoverflow.com/questions/33089948/project-3d-plot-to-2d-screen-plane

egeerardyn:
I wrote a little function to interactively play with this. If you want to play with the actual code, you will need to adapt the `updateProjection` subfunction.

```matlab
function testViewMtx(varargin)

if nargin == 0
    %varargin = {'cos(t)', 'sin(t)', 'sin(5*t)', [-pi pi]};
    varargin = {'t', 'sin(t)', '20*cos(t)', [0 10*pi]};
end

figure('name','viewmatrix tester');
%% top plot
% this plots the data and allows rotation
hAx(1) = subplot(3,1,1);
hLine(1) = ezplot3(varargin{:});

%% middle plot
% this plots the data and locks the rotation (i.e. the original plot)
hAx(2) = subplot(3,1,2);
colors = get(gca,'ColorOrder');
hLine(2) = ezplot3(varargin{:});

% just make sure the lines are what one expects
for ii = 1:numel(hLine)
    set(hLine(ii),'Color', colors(ii, :));
end

data = get(hLine(1),{'XData','YData','Zdata'});
data = [cat(1,data{:})', ones(numel(data{1}),1)];

hAx(3) = subplot(3,1,3);

% set the handlers to match the behavior described above
storeProp(hAx(2), 'View');
lh(1) = addlistener(hAx(1), 'View', 'PostSet', @updateProjection);
lh(2) = addlistener(hAx(2), 'View', 'PostSet', @restoreProp);
setappdata(gcf, 'Listeners', lh);
updateProjection;

%% handler: project the data in the bottom axis
    function updateProjection(src, event)
        axes(hAx(3)); cla; hold all;
        
        for ax = [1 2]
            [az, el] = view(hAx(ax));
            A = viewmtx(az, el);
            data_transformed  = A*data';
            Xt = data_transformed(1,:);
            Yt = data_transformed(2,:);
            
            plot(hAx(3), Xt, Yt, 'Color', colors(ax,:));
        end
    end
%% handler: restore the old value of the data
    function storeProp(h, propName)
        propVal = get(h, propName);
        setappdata(h, propName, propVal);
    end
    function restoreProp(src, event)
        gcbo = event.AffectedObject;
        propName = event.Source.Name;
        if isappdata(gcbo, propName)
            propValFixed = getappdata(gcbo, propName);
            propVal = get(gcbo, propName);
            if ~isequal(propVal, propValFixed)
                set(gcbo, propName, propValFixed);
            end
        end
    end

end
```

Maybe this helps to track down the problems, as everything updates on-the-fly.

miscco:
I believe, the answer lies in the combination of the following sites:

https://en.wikipedia.org/wiki/3D_projection

and

undocumentedmatlab.com/blog/undocumented-view-transformation-matrix

Especially the following function seems promising:
```matlab
function [p] = local_Data2PixelTransform(ax,vert)
% Transform vertices from data space to pixel space.
 
% Get needed transforms
xform  = get(ax,'x_RenderTransform');
offset = get(ax,'x_RenderOffset');
scale  = get(ax,'x_RenderScale');
 
% Equivalent: nvert = vert/scale - offset;
nvert(:,1) = vert(:,1)./scale(1) - offset(1);
nvert(:,2) = vert(:,2)./scale(2) - offset(2);
nvert(:,3) = vert(:,3)./scale(3) - offset(3);
 
% Equivalent xvert = xform*xvert;
w = xform(4,1) * nvert(:,1) + xform(4,2) * nvert(:,2) + xform(4,3) * nvert(:,3) + xform(4,4);
xvert(:,1) = xform(1,1) * nvert(:,1) + xform(1,2) * nvert(:,2) + xform(1,3) * nvert(:,3) + xform(1,4);
xvert(:,2) = xform(2,1) * nvert(:,1) + xform(2,2) * nvert(:,2) + xform(2,3) * nvert(:,3) + xform(2,4);
 
% w may be 0 for perspective plots 
ind = find(w==0);
w(ind) = 1; % avoid divide by zero warning
xvert(ind,:) = 0; % set pixel to 0
 
p(:,1) = xvert(:,1) ./ w;
p(:,2) = xvert(:,2) ./ w;
```

I could believe, that the failure of the current implementation is the position of the camera, which was supposed to be 5*diag(plot)

okomarov:
@miscco As yair commented, we cannot use that after R2014b. I can't get my head around why a double rotation of the current view onto the x-y plane does not work...it sounds very intuitive and at this point I am starting to think that there is either a bug or a conceptual misunderstanding.

egeerardyn:
Indeed. 

What I also noticed: in my script above, I use `[az,el] = view(gca); A = viewmtx(az,el)` and this works very smoothly. However, `A = view(gca)` produces extremely jumpy animations as if some slight changes in the angles sometimes produce bad matrices.
Another thing I saw: if you open up `inspect` on the 3D axis, you can actually play with some of the other values. It seems the `CameraUpVector` also plays a role and probably there are some other ones as well.

It's also quite unfortunate that in HG2, the relevant files to obtain the transformation matrix are p-coded (not that it seems to be the right transformation matrix, but that would have given a hint). You can, however, debug p-coded files, but it's quite hard to extract useful information.

Worst-case, we can always contact TMW about this. But since Michelle Hirsch responded on Yair's blog, I hope that we see better support/documentation for all of this.

Anyhow, while this is a nice-to-have feature, I have the impression that we would need to put a disproportionate amount of time and work into finding out what happens behind the scenes.

miscco:
Hm i believe, it has to do with the position of the camera, which btw is accessible by.

get(gca, 'CameraPosition'). In the wikipedia article this would by 'c'

I could imagine, that depending on the distance from the origin this might play a role. At least thats the only hint i have, as all the plots that work have a close camera and those that dont have one farter away.

miscco:
So one thing i will try if i have more time is an approach like that:

```matlab
figure(1)
h = ezplot3('t', 'sin(t)', '20*cos(t)', [0 10*pi]);
data = get(h,{'XData','YData','Zdata'});
data = [cat(1,data{:})', ones(numel(data{1}),1)];

% Get the position of the camera
CamPos = get(gca, 'CameraPosition');
CamPos = [CamPos, 0];
% Projection matrix on screen
[az,el] = view();
A = viewmtx(az,el);

data_projected = A*bsxfun(@minus, data, CamPos)';

xDataProjected = data_projected(1,:)./data_projected(3,:);
yDataProjected = data_projected(2,:)./data_projected(3,:);
figure(2)
plot(xDataProjected, yDataProjected)  
```

It is not really correct now, but it is going into the right direction.

okomarov:
Solution :) http://uk.mathworks.com/matlabcentral/answers/248362-screen-2d-projection-of-3d-plot

miscco:
Give that man a cookie! :+1:

egeerardyn:
@okomarov Nice job. :cookie: 
So it's basically a question of scaling the data to unit range before applying the transformation matrix.

The results look very close (well the only difference I could find was the angle of the envelope of those curves doesn't match the one in the 3D plot). But such differences can e.g. be caused by an incorrect aspect ratio when showing both figures.

okomarov:
If you leave the `axis equal` out the plot is very very close...
I still do not understand why the scaling has such a major impact :|

miscco:
I believe, that the aspect ration is accounting for the position of the camera, which was indicated by c_XYZ in the formula in the wikipedia site.

However, It is still not producing good results, as line simplification now screws around. I believe, one has to make sure, that the actual data aspects are preserved.

miscco:
I believe, that this is what in the actual formula on camera projection is given as the c vector. It is the distance of the camera to the data, which has a strong effect onto the transformation.

miscco:
Coming back to the aspect ratios, I believe, that the actual problem is the axis scaling in the original figure. The normal rotation assumes, that x,y, and z coordinate have the same metric (dx=dy=dz). However, the axes are not isometric, which has to be accounted for. and is reflected in the DataAspectRatio.

miscco:
Given that the current code seems to be working pretty well, any more comments?

src/cleanfigure.m
```diff
@@ -915,28 +1000,94 @@ function removeNaNs(meta, handle)
...
+    % sin(x - pi/2) = -cos(x)
+    Rot_x = [ 1         0         0         0
+              0         sin(el)   cos(el)   0
+              0        -cos(el)   sin(el)   0
+              0         0        0          1];
+
+    % Get the data aspect ration
+    AspectRatio = get(meta.gca, 'DataAspectRatio');
+    scale       = 1./AspectRatio;
+    scaleMatrix = [ scale(1) 0        0        0
```
egeerardyn: You can write this more succinctly:
```matlab
scaleMatrix = diag([1./aspectRatio 1]);
```
Also, it is common to begin variable names (with more than 1 letter) with a lowercase letter, i.e. `rot_x`, `rot_z` and `aspectRatio`.

miscco: True, i will write it that way. 

aspectRation was a type, however i wrote Rot_x to better refer to the wikipedia article about rotational matrices, that has them as R_x and R_y. Although i have to say, that i am totally indifferent here ;)

src/cleanfigure.m
```diff
@@ -493,11 +484,8 @@ function simplifyLine(meta, handle, targetResolution)
                 mask      = opheimSimplify(x, y, tol);
                 % Remove all those with mask==0 respecting the number of
                 % data points in the previoous segments
-                id_remove = find(mask==0) + countData;
+                id_remove = find(mask==0) + lineStart(ii) -1;
             end
```
miscco: Ah there is the bugfix... grml

miscco:
It seems there is a merge conflict, as i missed the space behind the -1 in line 487. I can fix that monday.

egeerardyn:
That's probably due to the recent merge. :smile:

miscco:
@matlab2tikzBot please test this

src/cleanfigure.m
```diff
@@ -223,7 +218,7 @@ function pruneOutsideBox(meta, handle)
   % By default, don't plot any points.
   shouldPlot = false(numPoints, 1);
   if hasMarkers
-      shouldPlot = shouldPlot | dataIsInBox;
+      shouldPlot = shouldPlot | dataIsInBox | all(isnan(data),2);
```
okomarov: Shouldnt it be `~all(isnan`?

okomarov:
There are 2 .jpg in the converted folder top delete.

src/cleanfigure.m
```diff
@@ -916,28 +1005,90 @@ function removeNaNs(meta, handle)
...
 % ==========================================================================
-function [bottomLeft, topLeft, bottomRight, topRight] = corners(xLim, yLim)
-    % Determine the corners of the axes as defined by xLim and yLim
-    bottomLeft  = [xLim(1); yLim(1)];
-    topLeft     = [xLim(1); yLim(2)];
-    bottomRight = [xLim(2); yLim(1)];
-    topRight    = [xLim(2); yLim(2)];
+function [bottomLeft, topLeft, bottomRight, topRight] = corners2D(xLim, yLim)
+  % Determine the corners of the axes as defined by xLim and yLim
+  bottomLeft  = [xLim(1), yLim(1)];
```
PeterPablo: Nitpicking: you use four spaces everywhere else, it seems :-)

test/suites/ACID.m
```diff
@@ -1818,10 +1818,10 @@
 function [stat] = parameterCurve3d()
   stat.description = 'Parameter curve in 3D with text boxes in-/outise axis.';
   stat.issues = 378;
-
-  ezplot3('sin(t)','cos(t)','t',[0,6*pi]);
+  stat.issues = 790;
```
PeterPablo: Mlease merge this line with `stat.issues = 378;` above, so it reads `stat.issues = [378, 790]` -- otherwise it overwrites the previous value.

test/suites/ACID.m
```diff
@@ -1818,10 +1818,10 @@
 function [stat] = parameterCurve3d()
   stat.description = 'Parameter curve in 3D with text boxes in-/outise axis.';
```
PeterPablo: `outise` --> `outside`, if you edit here anyhow... Thanks!

src/cleanfigure.m
```diff
@@ -272,7 +267,9 @@ function movePointsCloser(meta, handle)
   % box are preserved. This typically involves replacing one point by
   % two new ones and a NaN.
 
-  % TODO: 3d simplification of frontal 2d projection
+  % TODO: 3D simplification of frontal 2D projection. This requires the
+  % full transformation rather than the projection, as we have to calculate
+  % the inverse trnasfornmation to project back into 3D
```
PeterPablo: `trnasfornmation` -> `transformation`

miscco:
I went berzerk and updated all the indentation.

PeterPablo:
Thanks! The diff without whitespace can be viewed using the GitHub parameter [`?w=1`](https://github.com/matlab2tikz/matlab2tikz/pull/790/files?w=1).

egeerardyn:
@miscco Nice!

okomarov:
@miscco If you could run `testGraphical()`, then `make` the acid.pdf and share it here (as a link) for Octave and Matlab (one version should be fine), I am happy to review (the .pdfs) and finally merge your PR.

miscco:
There it is.

https://www.pdfhost.net/index.php?Action=Download&File=2071d9820ab1436eacf4ed359bb80c27

miscco:
ls

okomarov:
Cannot open file, says it's damaged

miscco:
Tried a different provider

okomarov:
@miscco Looks good. Only thing is `ACID(70)`, not smooth enough and it's actually not cleanfigure's fault but opengl's. Can you please make the plot finer?

miscco:
Hm, i could plot them with higher resolution maybe. I will try tomorrow.

okomarov:
```
t = linspace(0,20*pi,1000);
plot3(t,sin(t),50*cos(t))
```

miscco:
Ah you mean the ezplot3 yeah, i originaly had a version with linspace, but thought i keep with the previous code.

PR #791, opened 2015-10-13
Limit the precision of the data

miscco:
So a big bonus of the previous restructuring of the code is, that the limiting of the precision becomes rather easy.

This is a rebased version of #758


miscco:
Hm surprisingly it doesnt do anything, so it might not be worth to do anything here.

src/cleanfigure.m
```diff
@@ -939,6 +939,30 @@ function removeNaNs(meta, handle)
...
+
+  % Get the maximal precision for the data range
+  eps_range = eps(range);
+
+  % Scale the data wrt the range
+  data  = data / eps_range;
+
+  % Round to precision and scale back
+  data  = round(data) * eps_range;
+end
```
egeerardyn: I would put the precision reduction on a single line, i.e.
```matlab
% the least significant bit is proportional
% to the numerical precision of the largest number
lsb = eps(range) * alpha;

data = round(data/lsb) * lsb;
```
since that expresses best what the intent is. Putting everything across different lines is often nice, but here it's gone a bit overboard.

Note that, `alpha` is supposed to become a parameter to the `cleanfigure` function. Just imagine `plot(x, sin(x))`. The easiest to see the effect is around `0`: `sin(x)` may be small, e.g. `1e-14`, which in practice is a zero as one is likely to see. But `1e-14` will not be put to zero since `eps = 2e-16`. This `alpha` allows to put the bar a bit lower for this.

src/cleanfigure.m
```diff
@@ -777,9 +777,9 @@ function simplifyLine(meta, handle, targetResolution)
     yData = log10(yData);
   end
 
-  % Turn the data into a row vector
-  xData = xData(:);
-  yData = yData(:);
+  % Limit the precision of the data
+  xData = limitPrecision(xData(:));
+  yData = limitPrecision(yData(:));
```
egeerardyn: I think I see why nothing is happening: you are limiting the precision on the transformed data, i.e. the visual data, which you don't write back to the objects (or rather, that is what I expect, that you don't write back this transformed data).

src/cleanfigure.m
```diff
@@ -939,6 +939,30 @@ function removeNaNs(meta, handle)
...
+  % Only do something if the data is not empty
+  if isempty(data) || all(~isfinite(data))
+      return
+  end
+  
+  % Only consider finite values
+  mask = isfinite(data);
+
+  % Get the maximal value of the data
+  range = max(abs(data(mask)));
```
okomarov: Can we rename `range` into something else, e.g. `amax`, since by definition `range = max-min`?

egeerardyn: That's not a bad idea to avoid confusion. Maybe `maxVal` or `maxAbs` or something similar.

Also, I think both lines of code `mask = ...` and this one, can happen on a single line. It's nothing too complex.

Regarding the comments: I think we can do with fewer comments, given that the explain *what* is done (== job of the code) and not *why* it is done (== job of the comments).
If both lines are put on a single line, it makes a tiny bit more sense to put a comment to help reading the code
```matlab
% Determine the largest value that is still finite
largestValue = max(abs(data(isfinite(data))));
```

src/cleanfigure.m
```diff
@@ -939,6 +939,30 @@ function removeNaNs(meta, handle)
     topRight    = [xLim(2); yLim(2)];
 end
 % =========================================================================
+function data = limitPrecision(data)
+  % Limit the precision of the given data
+  
+  % Only do something if the data is not empty
+  if isempty(data) || all(~isfinite(data))
+      return
```
egeerardyn: The alternative would be to check if `range` (in your current code) is empty and return early in that case, since that is the actual problem.

In this code, I thought at first that those cases don't have to be treated in a special way, but they do.

miscco: Actually this captures two different problems. The first is an empty range. The second is that range or maxValue would be NaN if one of the data points is inf or NaN.

src/cleanfigure.m
miscco: Damn, sometimes one is just code blind. :(

src/cleanfigure.m
egeerardyn: Both cases that you describe are indeed problematic, but each of those lead to `data(isfinite(data))` to become empty and hence `range` to be empty and hence `data/[]` to fail.

So while there are two conditions you need at the beginning, there is only a single one that you need on `maxValue`. Moreover, currently it's impossible for `maxValue` to become `NaN` or `Inf` since those non-finite values don't get passed to the call to `max`.

miscco: Hm that is strange, because as i tested it, range got NaN.

egeerardyn: The only case that I saw where `max` returns `NaN` is if you do `max(NaN(1,N))`. In all other cases, `NaN` is omitted (see the help of `max`). But since there should never be a call to max which even has `NaN` or `Inf` to begin with, I don't see a case where that can happen.

miscco:
Yeah, breaking it again.

As far as i can see, the differences in hashes at least for matlab come from different axis properties (see test 31 for an example)

It seems octave is doing a much better job at reducing the number of data points in ACID 70 (although it shouldnt) BTW is it intended that the title is erased when using cleanfigure? I went back a lot of commits, but it was always the same. This also leads to some strange formating in the tikz file.

src/cleanfigure.m
```diff
@@ -500,6 +510,46 @@ function simplifyLine(meta, handle, targetResolution)
...
+  if isempty(data) || all(all(~isfinite(data)))
+      return
+  end
+
+  % Get the maximal value of the data, only considering finite values
+  maxValue = max(abs(data(isfinite(data))));
+
+  % The least significant bit is proportional to the numerical precision
+  % of the largest number. Scale it with a user defined value alpha
+  eps_range = eps(maxValue) * alpha;
```
egeerardyn: You might want to get rid of `range` here as well and use something more verbose:
 - `lsb` (that's what I would use, but it's probably quite specific to electrical engineering jargon (DAQ/ADC-related stuff))
 - `precision`
 - ...
since `range` is not the right term and not used anywhere else anymore.

src/cleanfigure.m
```diff
@@ -61,14 +64,17 @@ function cleanfigure(varargin)
   m2t.cmdOpts = m2t.cmdOpts.addParamValue(m2t.cmdOpts, 'targetResolution', 600, @isValidTargetResolution);
 
   m2t.cmdOpts = m2t.cmdOpts.addParamValue(m2t.cmdOpts, 'minimumPointsDistance', 1.0e-10, @isnumeric);
+  m2t.cmdOpts = m2t.cmdOpts.addParamValue(m2t.cmdOpts, 'scalePrecision', 1, @isnumeric);
```
egeerardyn: Now that I think of it, we probably should also define a value where the whole function is disabled. E.g. `scalePrecision <= 0` or `NaN` could be reasonable things to do.

miscco:
So any more comments on this? Note that I have added a check to it in case of alpha <= .0

egeerardyn:
Code is :+1: for me.
Only thing left to do is run the test suite and make sure it passes. What I suggest is that you run `testGraphical` on your machine and compare the results visually before and after this PR (or `diff` the resulting TikZ files). If they look the same, all should be fine and you can overwrite the hashes with those from Jenkins.

I don't see a reason why HG1 or Octave would produce very different deltas for this PR, but anyway: I will try to have a go at it tomorrow evening on my machine to doublecheck.

miscco:
As i am currently loaded with work i would wait here for the 3D stuff to be merged, so that I can rebase cleanly.

miscco:
So i had some spare time and it seems, that logarithmic plots seem affected. From what i found from testGraphical are the following tests:
ACID(8): Different representation of the numbers 0.03->3*10^-2
acid-8
ACID(16): completely broken
acid-16
ACID(17): Some strange brokenness
ACID(22-24): Arrowheads are gone/extremely small
acid-20
acid-21
acid-22
acid-20
acid-21
acid-22
ACID(32-33): x axis gone???

acid-30
acid-31
ACID(42) didnt compile, dont know why. Had to remove the plots
ACID(48-49): Additional ticks
ACID(56): sphere isnt cut off
acid-54
acid-55
ACID(57) again not cutoff and log scale is gone
ACID(63) white lines in bars are gone
ACID(68-69) Size completely different of the formulas

acid-66
acid-67
ACID(72) troughts filled with black?
ACID(73): Im out of words
acid-71
ACID(78) green field???
acid-76

ACID(86) Additional lines??
acid-84

ACID(87-89) just broken in many ways
acid-85
acid-86
acid-87

ACID(92-93): wrong coloring
acid-91
acid-92
ACID(97): wrong coloring
acid-95

ACID(98-99) wrong layering
acid-96
acid-97
Here the pdf

miscco:
Ok, i have to say, that at least some of the failures stem from MALTAB2015b. Because i get mostly the same stuff when using develop. Actually there, more tex files fail to compile

okomarov:
Many of the things you listed are either missing features or known bugs. Others are artifacts of the pdf export, for instance 78 and 86 are correctly rendered in a Matlab figure but the pdf doesn't keep e.g. the green box. 

From my visual memory, and the figures you posted, 16 and 17 are the tests to look at and 93 (although it might be a regression due to version or another PR).

egeerardyn:
Thank you for looking into this!

There are some of the tests I want to respond to:
 - [x] *ACID(8): Different representation of the numbers 0.03->3*10^-2* This is a feature of `pgfplots` that can be set by the user. Not our concern.
 - [ ] *ACID(16): completely broken* **probably this PR**
 - [ ] *ACID(17): Some strange brokenness* **probably this PR**
 - [x] *ACID(22-24): Arrowheads are gone/extremely small*. Fixed in #781 
 - [ ] *ACID(32-33): x axis gone???* I guess this is caused by `stemseries` not being translated fully.
 - [ ] *ACID(42) didnt compile, dont know why. Had to remove the plots*
 - [ ] *ACID(48-49): Additional ticks* Probably nothing serious, just different settings in pgfplots and MATLAB
 - [x] *ACID(56): sphere isnt cut off* default behavior in HG2, we have a feature request for that somewhere
 - [x] *ACID(57) again not cutoff and log scale is gone* Known bug.
 - [x] *ACID(68-69) Size completely different of the formulas* We don't actually scale those "properly" (question is what would be proper scaling)
 - [x] *ACID(73): Im out of words* There is a TODO for this in the code.
 - [x] *ACID(78) green field???* MATLAB eps export removes the coloring, not our problem.
 - [x] *ACID(86) Additional lines??* This is not our problem. It's MATLAB's EPS export that does not support transparency (but we do).
 - [x] *ACID(87-89) just broken in many ways* We have some feature requests open for this
 - [ ] *ACID(92-93): wrong coloring* Seems like the layering is reversed.
 - [x] *ACID(97): wrong coloring* MATLAB eps export removes the coloring, not our problem.
 - [x] *ACID(98) wrong layering* MATLAB eps export does not really support transparency
 - [ ] *ACID(99) wrong layering*

There are a few figures missing:
 - [ ] *ACID(63) white lines in bars are gone*
 - [ ] *ACID(72) troughts filled with black?*


In general, I think it's more useful to do something like (MATLAB on Unix)
```matlab
cd ./test
addpath(fullfile(pwd('suites'))
!git checkout develop
testGraphical
!cd tex && make -j 8 && cp acid.pdf before.pdf && open before.pdf
!cd ..
!git checkout <thisPR>
testGraphical()
!cd tex && make -j 8 && cp acid.pdf after.pdf && open after.pdf
```

and compare where before and after differ. Everything else should be known problems and is probably not related to the PR anyway.

miscco:
So ACID 18 seems to be slightly broken, which stems from simplifyLine. I will have a lock why it is behaving incorrectly.

ACID(16) however can be fixed by calculating the LSB of the visual data instead and the scaling back.

miscco:
Hm i found the error. There was a bug i introduced in simplifyLine, that i actually thought to have fixed in a commit. Strangely the fix is gone now?!. Maybe i made a rebase f... up?

miscco:
So currently i cant find any differences, that were not there before, so I will rebase it again and you guys can maybe have a look on MATLAB 2014a/b

miscco:
So as I finally have some spare time again, I will rebase this again once 3D projection is merged and look at the outcome.

miscco:
So here is the graphical output of this PR reabsed on the latest develop.

http://workupload.com/file/40UDpb6X

Some strange things seem to happen in ACID(8), especially as cleanfigure does not lead to any visual changes in the figure itself.

egeerardyn:
@miscco What strange things in `ACID(8)`? Do you mean the formatting of the iso-line numbers? That's a responsibility of pgfplots, so that's something that could be improved by setting the proper keys. 

However, currently we don't set them (and I think in many cases it makes a lot more sense to set those globally in a document).

Similar things tend to happen with axis labels (see e.g. `ACID(22)`).

miscco:
Ah, ok it seems that we should just update the reference figures some time in the near future. In ACID(8) there are some differences to the reference, that are also there in develop. (some contours seem different, as the labels in the reference cover them.)

miscco:
So i have updated the hashes, note that there will be merge conflicts with #790 as it changes one of the tests.

egeerardyn:
@miscco The reference figures are what MATLAB/Octave produces if you just use the `print` function (with EPS/PDF output).

I don't see big differences between the shape of the iso-lines on both plots (and those that I see can probably be explained by moving the labels a bit, but as far as I recall, we don't have any control over those locations in both MATLAB and pgfplots).

okomarov:
@miscco There is a `cleanfigure.m~` artifact, due to rebasing/conflict solving (?). When solving the next conflict please do a merge commit rather than a rebase, it's good to save the timing of the comments :)

okomarov:
In reference to [Egon's list](https://github.com/matlab2tikz/matlab2tikz/pull/791#issuecomment-149932406) with the problems:
* ACID(63), white lines are [artifacts of the renderer](http://www.mathworks.com/matlabcentral/answers/99614-why-do-i-get-white-line-artifacts-in-my-image-when-rotating-a-3-d-surf-plot-in-matlab-7-1-r14sp3) during the print() 
* ACID(72) is a pre-existing problem. If not listed among the issue it should be. I vaguely recall it's a HG2 problem only.
* ACID(93) should have an issue
* ACID(99) issue related to contourf()

I do not see problems with the `cleanfigure()` from the .pdf. We just need a visual check on Octave this PR should be good to go. 

I reckon the tuning of the alpha parameter can be done in a separate PR.

src/cleanfigure.m
```diff
@@ -501,6 +512,64 @@ function simplifyLine(meta, handle, targetResolution)
...
+  if isAxis3D(meta.gca)
+      data  = [xData(:), yData(:), zData(:)];
+      isLog = [isXlog, isYlog, isZlog];
+  else
+      data  = [xData(:), yData(:)];
+      isLog = [isXlog, isYlog];
+  end
+
+  % Only do something if the data is not empty
+  if isempty(data) || all(all(~isfinite(data)))
```
okomarov: equivalently `all(isinf(data(:)))`

egeerardyn:
With respect to `ACID(93)`, couldn't this be due to an outdated version of `pgfplots`? I know its manual says that both types together is not really supported, but I am 100% sure I tested that plot before releasing 1.0.0: it does work properly when your pgfplots version is recent enough (I don't remember what exact version though, I suspect I put that version in the code though).

miscco:
Have you checked the pdf i uploaded? There the issue with 93 is solved. It was indeed an outdated pgfplots version (1.09)

Actually i think 98 does the right thing.

@okomarov Yeah, rebasing is a nightmare, i explicetely checked for chagned files, but i must somehow missed them. I will merge in the future, alone to keep me sane. The Whitespaces!!

okomarov:
The issue in ACID(93) persists, looking at the pdf:

untitled

The merging won't spare you the conflicts resolution. With the rebase there was a git setting to carry over the resolution over the many commits (so maybe that's the nightmare you are experiencing).

miscco:
The nightmare is, that all the code changed through the indentation cleanup  i added. so its hard to find the real conflicts with git diff.

Yeah it seems more text is wrong. Doesnt that also apply for develop?

egeerardyn:
@miscco : Maybe try with `git diff --ignore-space-change` or `git diff --color-words` (I rely on the latter a lot of LaTeX documents).

miscco:
So i went ahead with @okomarov suggestion and just merged it rather than a rebase. It should be fine now.

miscco:
Gentle Ping

okomarov:
Can you please run a testGraphical on the tests that had their hashes changed before this PR and after and link to the results?

miscco:
So here are they (MATLAB2015b)

legendplotBoxoff
acid-17

parameterCurve3d
acid-68

quiveroverlap (that one is always brocken
acid-20

sine with markers
acid-2

polarplot
acid-27

roseplot
acid-28



compassplot
acid-29

customLegend 
acid-76

stairsplot
acid-26

zplanePlot1
acid-45

overlappingPlots
acid-95

quiverplot
acid-21

okomarov:
@miscco Are you on win? If yes, please install MinGW, to get the `make`, so that you can automate the production of the graphical output into a .pdf (upload it on e.g. dropbox) and link to it. I am saying that because I understand the pain of posting each single case and also because we need a comparison against current develop on those tests.

If you are wondering why, it's because (except for quiverplot which seems unreliable) we are not using a scaling factor alpha > 1, hence I did not expect tests to fail. Since this PR is quite important for reducing discrepancies across platforms/editions it's good to have a track of what is exactly changing and why.

Also, I would have ran the tests myself to help you out but I am currently working on one hand and a leg :)

miscco:
@okomarov Right now i am on win, but normaly i am under linux. 
The JPGs are taken from the PDF via
```
convert -density 216 acid.pdf acid.jpg
```

I actually just overloocked, that many tests were failing idividually for the different versions, and thought it would be less jpgs

okomarov:
We don't mind looking into full .pdfs, less work for everyone :)

miscco:
So i have merged to the latest develop. As we now start to accumulate additional options in recursiveCleanup, i went ahead and passed m2t.cmdOpts.Results rather than the 3 individual values. That should keep the code more readable and easier to maintain.

miscco:
Here is the pdf:
http://workupload.com/file/VsKJu2Ns

miscco:
So in addition to the acid.pdf, i went ahead and  tried the imagemagick compare magic.
So there is no pixel difference between this PR and develop branch :smile:

egeerardyn:
Code looks nice to me, especially :+1: for just passing the `cmdOpts` struct instead of all these variables.

The only thing I am thinking about, is the default parameter that we set for `scalePrecision`. Currently, that is `1`, so we round everything to a precision of `eps(max(abs(data)))`. That probably works for most graphs just fine, but it is not really a conservative option that we introduce (i.e. it eats some of the precision of our user's data).
Perhaps we should set it to `0` by default (so disable the function) and in our test framework set it to `1`. What do you guys think about this?

miscco:
So generally, i am kind of skeptical, whether this change has any effect outside of our test suite.

egeerardyn:
@miscco It probably has some effect outside our test suite (i.e. points shifting a bit). However, I think you are right that there is should be no visible influence at all.

The only case I see is e.g. when you have a plot like
```matlab
plot([1e20 1e30 1e40 10 20])
ylim([0, 100])
```
where `max(eps(abs(data)))` would be in the order of `1e25`.
However, since we first prune points outside of the box, such a situation can never arise.

Also, since `cleanfigure` is only used by more advanced users (I think), a default value is probably not an issue for them. So any value is also fine by me.

egeerardyn:
@miscco If you think this is done (could you remove the "work-in-progress" label in such cases?), I am fine with merging this in :+1:. Results look fine to me, but I have to admit I don't have the track record here very clearly in my mind.

miscco:
Hm, i think i dont have sufficient rights to do that.

or I just simply dont know how

egeerardyn:
Hmm, on this end you do seem to have the right permissions:
screen shot 2015-12-01 at 19 55 42

In any case, since you have contributed a lot of great code over the last few weeks, I think it's about time I add you as a full-blown developer.

miscco:
Ok now i have the possibility to change the labels. Seems that associate developers dont.

egeerardyn:
You're completely right. Associate developers get read access on the repository, so apparently [cannot](https://help.github.com/articles/permission-levels-for-an-organization-repository/) set the labels.
Sorry for the misunderstanding.

Anyhow, this means you also get to merge PRs of other devs from now on :smile:

PR #792, opened 2015-10-14
Keep the title visible.

miscco:
Currently, cleanfigure turns text outside of the axis limits invisible. This however is always true for the title, which in octave would switch the title invisible. Therefore check whether the handle is the title.


egeerardyn:
:+1: on the code. It's very nice that you fixed this unintended behavior.

You only need to make sure that the hashes are adapted accordingly to make all tests pass, but then it can be merged.

miscco:
So i did update the hashes. However please note, that i have not possibility to check it. The strange thing is still, that for example in octave 4.0 the title was omited but still in the tikz picture. It might be that there is some workaround for that.

egeerardyn:
@miscco The title (and X/Y/Z labels for that matter) are handled in a special way in `m2t` (look for the `getTitleOrLabel_` function in the code, or for `drawTitleOfAxes` in specific for this case).

TL;DR: these always get drawn by placing them in the options of the axes. That makes me wonder, though: how does that work together with this fix? Can you check in the TikZ output files whether the titles are specified redundantly? Probably, this should also be visible in the graphical output as 2 -slightly shifted- copies of the same text.

miscco:
Sadly, I cannot test it, as i only have MATLAB2015b and Octave 4.0 For both however there is change at all. Therefore i would guess, that it does not have any backdrops.

egeerardyn:
Thank you! I have merged it since the results look fine on R2014a, since I presume that you tested on R2015 without a problem and Octave 4.0 vs 3.8 shouldn't be really different, I think.

Issue #687, opened 2015-06-29
Uniformize test suite for R2015


egeerardyn:
In MATLAB R2015a (and probably R2015b as well), the test suite hashes have changed quite a bit (only the non-OK results are shown):

```
The following tests are known to be unreliable. They, however, do not cause the build to fail.

           ACID(  1) multiline_labels         : hash 914722e4c88bbb14f637b98f568e9196 != (f469978c28dab64e4ab9e0435a92b3d6)
           ACID(  6) sine_with_annotation     : hash 967ec66a75def06e5028e6c606e8eb2c != (9da03e407cbb6e787e2e83e1ddd9753a)
           ACID(  9) contourPenny             : hash 2ff4dd71c712f2aac2577a5312c153ec != (33c206795e296e58802dcd582c623d05)
           ACID( 10) peaks_contourf           : hash a167ed71207095c8e98ee1d5553deaab != (ddb1ec21fe11c28effa5fdf41c394e04)
           ACID( 12) double_colorbar          : hash 62e8de88aa986a9c1efa06cb0575f446 != (effce959246ead301919d3cfdc917cc6)
           ACID( 16) logplot                  : hash abc60dd4344bdb611d59da4f4aa0982d != (ca3f85050beb641236ba94a1d9bc4b5f)
           ACID( 17) colorbarLogplot          : SKIPPED
           ACID( 18) legendplot               : hash e2c416d672c1e7dfc0c0d512e2be34b3 != (7054ed0ea5ccca16b74e32b01787a93a)
           ACID( 24) quiver3plot              : hash be662bed35ba46b60e8e37d3ed27b6d7 != (f8dbf572e4cfb51aa660cee0ca2ae485)
           ACID( 34) bars                     : hash e4c8ea6e0d90c330728f4909a382f11f != (d85ca6286f7643b97fcfe57f87d4021b)
           ACID( 38) subplot2x2b              : hash bdf7da9eb8bac34c793117fd99fd5f94 != (c6e82de40e07dfc7d9b429e6bdcbf7f6)
           ACID( 40) subplotCustom            : hash 5d6957a0d3f342c346e8f2eeca1dd58a != (9e60fdf0f636770e831a2875cafb7345)
           ACID( 45) besselImage              : hash 8c7ea17c04f1dee3adfaea8a4b7345d9 != (dc98fd21efedd5794d48024d334c19b8)
[...]

PR #793, opened 2015-10-14
Added a hash file for MATLAB2015b

miscco:
Even if jenkins doesnt check it, it really helps to keep people with MATLAB2015b (me) sane, as i can finally tell whether I changed something or not.


egeerardyn:
Although I sympathize with your inconvenience, I don't want to merge this.

There is already enough to do with 3 environments and those 3 make sense as they have really different graphics systems. If we add R2015 hashes, that's just another burden. The thing is that having those hashing in source control is already an anti-pattern since we mix source and artifacts in the repository (this means that the hashes can be outdated and sometimes/often cannot be trusted). This makes bug-hunting more difficult.

I would propose that you check your R2015a/b results against a build that runs on Jenkins. Where your tests fail, you can just set 
```matlab
status.unreliable = isMATLAB('>=', 8.5); %TODO make test compatible with R2015
```
or change the current line accordingly. 

That way our test coverage is less the ideal in R2015 releases, but Jenkins will help us to enforce R2014 at least and we don't end up maintaining 4 sets of hashes. In the long run, however, we should certainly look at why those differences exist and try to get rid of them as in #687.

PR #794, opened 2015-10-21
Fix hashes in R2014a

egeerardyn:
Apparently, I f*'ed up the hashes during the merge in decd496

This should fix that.

I noticed this during testing.


miscco:
Any idea how that happened?

egeerardyn:
Not really, and I don't have the time to investigate it right now. It's likely the PR that was merged right before it that behaves more platform dependently, but I don't have solid backing for that hunch.

PeterPablo:
If decd496 really broke so many ACID tests, I suggest to revert that commit/PR instead of simply adding five FIXME's that go with unreliable tests. As you know I put a lot of effort into reducing the number of unreliable tests. Throwing this away for a quick fix seems wrong.
Unfortunately I currently don't have the time to investigate this. It makes me curious that you note `MAC vs. linux`, which indicates that this issue is actually older, but was never flagged before. I wonder why though, since I used to run the testsuite on both MAC and linux.

miscco:
Hm, so the one function in matlab2tikz is 
```matlab
function [m2t, opts] = getTitleOrLabel_(m2t, handle, opts, labelKind, tikzKeyword)
% gets a string element from an object
    if ~exist('tikzKeyword', 'var') || isempty(tikzKeyword)
        tikzKeyword = lower(labelKind);
    end
    object = get(handle, labelKind);

    str = get(object, 'String');
    if ~isempty(str)
        interpreter = get(object, 'Interpreter');
        str = prettyPrint(m2t, str, interpreter);
        style = getFontStyle(m2t, object);
        if length(str) > 1 %multiline
            style = opts_add(style, 'align', 'center');
        end
        if ~isempty(style)
            opts = opts_add(opts, [tikzKeyword ' style'], ...
                   sprintf('{%s}', opts_print(m2t, style, ',')));
        end
        str = join(m2t, str, '\\[1ex]');
        opts =  opts_add(opts, tikzKeyword, sprintf('{%s}', str));
    end
end
```
So as far as I can tell, it does not check for visibility. Therefore, the change in the PR should not have any effect at all. Could it be, that i have used 'title' for the property in the PR? It should actually be safe and there were no conflicts in jenkins at all.

PR #797, opened 2015-10-22
Fixed incorrect accounting for previous line segments in simplifyLine

miscco:
This fixes an awkward bug in simplifyLine that I actually thought I had fixed o.O.

In the end the accounting for the elements of the previous line segments was bogus and completely overcomplicated.

### DIFF
--- a/src/cleanfigure.m
+++ b/src/cleanfigure.m
@@ -491,8 +491,8 @@ function simplifyLine(meta, handle, targetResolution)
             if numel(x) > 2
                 mask      = opheimSimplify(x, y, tol);
                 % Remove all those with mask==0 respecting the number of
-                % data points in the previoous segments
-                id_remove = find(mask==0) + lineStart(ii) -1;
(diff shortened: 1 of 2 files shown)


miscco:
What is really strange about this bug, is that the code actually crashed for me and i fixed it. I dont know how I messed this up.

Quiverplot as always says it changed, however i can see no difference. Is there a reason why ACID(16) legendplot is marked unreliable?

egeerardyn:
Legendplot (`ACID(18)`) is unreliable for an unknown reason (see the code of `ACID.m`), but `ACID(16)` is reliable.

When rebasing and stuff, I also often find that it's unclear whether a change has been included or not, especially if you have to go back in to fix some conflicts. Probably you forgot a `git add` or something after you resolved the rebase conflict.

miscco:
Some more Request for this fix?

egeerardyn:
:+1: Thank you for fixing this up!

PR #798, opened 2015-10-23
drawBaseline for barplots and stemseries

okomarov:
Now draws baseline in barplots and stemseries plots:

ACID(32)
capture

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -3847,7 +3847,7 @@ end
 % ==============================================================================
 function [m2t,str] = drawBaseline(m2t,hparent,isVertical)
     str = '';
-    if isOff(get(hparent,'ShowBaseLine'))
+    if isOff(get(hparent,'ShowBaseLine')) || ~isHG2()
         return
     end
     
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -3847,7 +3847,8 @@ end
 % ==============================================================================
 function [m2t,str] = drawBaseline(m2t,hparent,isVertical)
     str = '';
-    if isOff(get(hparent,'ShowBaseLine')) || ~isHG2()
+    baseValue = get(hparent, 'BaseValue');
+    if isOff(get(hparent,'ShowBaseLine')) || ~isHG2() || baseValue ~= 0
         return
     end
     
@@ -3872,11 +3873,11 @@ function [m2t,str] = drawBaseline(m2t,hparent,isVertical)
 
     % Get data
     if isVertical
-        xData = repmat(get(hBaseLine, 'BaseValue'),1,2);
+        xData = repmat(baseValue,1,2);
         yData = get(m2t.currentHandles.gca,'Ylim');
     else
         xData = get(m2t.currentHandles.gca,'Xlim');
-        yData = repmat(get(hBaseLine, 'BaseValue'),1,2);
+        yData = repmat(baseValue,1,2);
     end
     
     [m2t, table, tabOpts] = makeTable(m2t, '', xData, '', yData);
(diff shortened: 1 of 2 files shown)


src/matlab2tikz.m
```diff
@@ -3710,6 +3710,9 @@ function warnFacetedInterp(m2t)
     str = sprintf('\\addplot[%s] plot table[%s] {%s};\n', ...
                  opts_print(m2t, drawOptions, ','), ...
                  opts_print(m2t, tabOpts, ','), table);
+
+    [m2t, baseline] = drawBaseline(m2t,h,isHoriz);
```
egeerardyn: What bothers me a bit is that you pass `isHoriz` to a parameter called `isVertical`. I guess `isHoriz` means the bars are in a horizontal direction, so the base line should be vertical.
I think that should become apparent from the code as well (if my explanation is correct, of course).

src/matlab2tikz.m
```diff
@@ -3842,8 +3845,52 @@ function warnFacetedInterp(m2t)
     [m2t, drawOptions] = setColor(m2t, h, drawOptions, 'fill', faceColor);
 end
 % ==============================================================================
+function [m2t,str] = drawBaseline(m2t,hparent,isVertical)
+    str = '';
```
egeerardyn: Could do with some help text.

src/matlab2tikz.m
```diff
@@ -3842,8 +3845,52 @@ function warnFacetedInterp(m2t)
...
+    baseValue = get(hparent, 'BaseValue');
+    if isOff(get(hparent,'ShowBaseLine')) || ~isHG2() || baseValue ~= 0
```
egeerardyn: Why don't we draw the line when `baseValue == 0`? That should be documented in the code (or we should just draw it).

src/matlab2tikz.m
```diff
@@ -3842,8 +3845,52 @@ function warnFacetedInterp(m2t)
...
+        return
+    end
+    
+    if ~exist('isVertical','var')
```
egeerardyn: Please put this at the top of the function, that makes sure that `isVertical` is available in the whole function.

egeerardyn:
Nice work already!

src/matlab2tikz.m
```diff
@@ -3842,8 +3845,52 @@ function warnFacetedInterp(m2t)
...
+        xData = repmat(baseValue,1,2);
+        yData = get(m2t.currentHandles.gca,'Ylim');
+    else
+        xData = get(m2t.currentHandles.gca,'Xlim');
+        yData = repmat(baseValue,1,2);
+    end
+    
+    [m2t, table, tabOpts] = makeTable(m2t, '', xData, '', yData);
+
+    str = sprintf('%s\\addplot[%s] plot table[%s] {%s};\n', ...
```
egeerardyn: Shouldn't this be `%s\\addplot[%s] table[%s] {%s};\n` ? I know we have the `plot table` somewhere else in the code as well (and it compiles just fine), but in the `pgfplots` manual I can only find `\addplot table`, never `\addplot plot table`, though `plot table` is used as a search term in the manual.

What do you think?

src/matlab2tikz.m
okomarov: Yes, you are correct. I was testing if you were paying attention :laughing:!
I will add comments, that I usually do at the end unless the code is hard to understand.

src/matlab2tikz.m
okomarov: U mean when ~= 0?

src/matlab2tikz.m

src/matlab2tikz.m
okomarov: Will have to look into it, copy-pasted from stemplot if I recall correctly.

src/matlab2tikz.m
egeerardyn: Oh, indeed. But the question remains the same (well, actually replace `==` with `~=`, but you get the point).

src/matlab2tikz.m

egeerardyn:
:+1: Feel free to merge when you're done with this.

PeterPablo:
Code looks good (did not runtime test)! :+1: 
Is HG1 not affected, since the baseline is already dealt with implicitly?

okomarov:
@PeterPablo Yes, since the baseline is a line object and the handleAllChildren() detects it.

Issue #808, opened 2015-11-04
whitespace 


okomarov:
Can we decide for one standard so that I can develop a small tool to manually run before issuing a PR?

I suggest for Matlab's smart indent + remove whitespace on empty lines. Hence, without the 2space indentation in a function.

What are your thoughts?

egeerardyn:
Sure!

My personal preference is:
 - no trailing spaces if possible
 - 4 spaces (not tabs) per level: 2 spaces is often not enough
 - [Indent all functions](http://blogs.mathworks.com/community/2009/05/11/keep-your-code-readable-with-smart-indenting/)

However, I'm happy with any agreement (only spaces vs. tabs I'm completely in the spaces camp).

However, the problem with automatic formatters is that these often ruin properly crafted code, e.g. line continuations:
```matlab
% crafted
var = struct('bla', 'bla', ...
             'foo', 'foo', ...
             'baz', 'baz');
```
versus
```matlab
% smart-ident
var = struct('bla', 'bla', ...
    'foo', 'foo', ...
    'baz', 'baz');
```

Luckily, smart-indent leaves space in the middle intact (although we could do with some guidelines there as well: spaces after each comma).

In any case, if we can formulate some rules, I'm quite convinced there might be some plugins for Jenkins available to check those rules such that in the future we don't pull oddly-formatted code.

egeerardyn:
I will leave this open as a reminder to myself to fix indentation before the 1.1 release of m2t.

PR #809, opened 2015-11-04
Whitespace format

okomarov:
Whitespace formatting as per Egon's comments in https://github.com/matlab2tikz/matlab2tikz/issues/808.

Also, keeps crafted formatting intact. 

PS. I'll add H1 line and help at the end.


src/dev/formatWhitespace.m
```diff
@@ -0,0 +1,66 @@
...
+    % Check for mlint error
+    if ~isempty(tree.mtfind('Kind','ERR'))
+        [~,name,ext] = fileparts(obj.Filename);
+        error('formatWhitespace:mlintError','''%s'' contains syntax errors. Cannot proceed.',[name,ext])
+    end
+
+    % Smart indent except for crafted continuations of line
+    d.smartIndentContents()
+    lines      = textToLines(d.Text);
+    iDots      = ~cellfun('isempty',regexp(lines, '\.\.\.','once'));
```
egeerardyn: Wouldn't this be more readable using just `strfind` (I thought there was also a MATLAB function to convert escape literal strings into regular expressions, but I can't find it back).

src/dev/formatWhitespace.m
```diff
@@ -0,0 +1,66 @@
...
+    import matlab.desktop.editor.*
+    try
+        d        = getActive();
+        oldLines = textToLines(d.Text);
+    catch
+        error('formatWhitespace:noEditorApi','Check that the Editor API is available.')
+    end
+
+    % Parse file into a tree
+    tree = mtree(d.Text);
```
egeerardyn: I knew this kind of trickery existed somewhere in MATLAB. :smile: Is there actually some sort of documentation that you use for this? If so, I would really love to see it.

okomarov: Yeah! It's called trial-and-error hahah (and reading the code...)

I used it in my https://github.com/okomarov/justify/blob/master/justify.m and you can check the use case.

src/dev/formatWhitespace.m
okomarov: strfind all the way :+1:

src/dev/formatWhitespace.m
egeerardyn: yay for trial-and-error (and undocumented MATLAB features)

egeerardyn:
What I don't see in the code is how it is enforced that all functions are indented (i.e. how that setting is normalized across environments: that's a setting in MATLAB, but it is certainly not the default one).

Also, I think one of the things that may be worthwhile to implement as well, would be to specify a file name that needs to be processed (i.e. open it in the editor, process and close) and if none is supplied, just process the active editor.

okomarov:
Function level indentation is enforced in the following way:

- Smart indent, i.e. reset indentation to 0 spaces.
- The last for loop adds 4 spaces to each block within a function. It does so for each function block in a sequential way, hence nested functions (that we do not have)  will have their content indented twice.

egeerardyn:
Here it seems to have an unintended consequence that some levels are indented too much. I will produce an MWE where the problem shows together with my MATLAB settings. (edit will follow)

src/dev/formatWhitespace.m
```diff
@@ -0,0 +1,66 @@
...
+
+    % Find position of function/end pairs
+    tmp   = tree.mtfind('Kind','FUNCTION');
+    stpos = tmp.lineno;
+    enpos = tmp.lastone;
+
+    % Pad content with 4 spaces
+    for f = 1:numel(stpos)
+        pos        = stpos(f)+1:enpos(f)-1;
+        lines(pos) = strcat({'    '}, lines(pos));
```
egeerardyn: Perhaps it would be better to define a variable `INDENT = repmat(' ',1,4);` at the beginning of the file.

miscco:
Very nice, it was really confusing at the beginning, what indentation was intended.

okomarov:
`MWEIndent.m` is indented correctly now.

src/dev/formatWhitespace.m
```diff
@@ -0,0 +1,71 @@
...
+        error('formatWhitespace:noEditorApi','Check that the Editor API is available.')
+    end
+
+    % Get specific document if filename is specified
+    if ~isempty(filename)
+        allDocs    = matlab.desktop.editor.getAll();
+        [~,fnames] = cellfun(@fileparts, {allDocs.Filename},'un',0);
+        [tf, pos]  = ismember(filename,fnames);
+        if ~tf
+            error('formatWhitespace:filenameNotFound','Filename "%s" not found.', filename)
```
egeerardyn: Perhaps replace `tf` by a more descriptive name. Also, if I interpret this correctly, does this mean that this function only works with files that are currently open in the editor?
If so, please just add a `TODO: open file if it isn't open in the editor already`.

egeerardyn:
:+1: Nice work! I think this will improve readability of our code base significantly.

When all current PRs are merged, we should create a PR with just indentation changed. That way, we avoid merge conflicts amongst different PRs.

okomarov:
Ahem...here I already formatted m2t, as in  56ec8c0.

egeerardyn:
Ah, that's why `matlab2tikz.m` was changed without any visible changes on GitHub :+1: I thought it was some weird kind of artefact from testing the procedure.

Anyhow, I'll try to do this regularly.

Issue #393, opened 2014-10-31
Adding string to the tikzpicture environment


t4qjXH8N:
Situation: I manually added an extra axis with images and arrows to a regular dataplot, created with matlab2tikz. This manually added axis has its own grid, which makes it easy to position the images and the arrows. 

Problem: Everytime when I update my dataplot with matlab2tikz, the manually added axis is lost (of course). 

My approach: I created a text file containing the new axis environment only. After reading it into a string with MatLAB, I could not manage to add the string into the tikzpicture environment. Adding strings to the OPTIONS of the tikzpicture environment is, of course, possible but does not solve the problem.

Feature Request: Add an option to add strings to the tikzpicture environment.

Or are there any other ideas?

egeerardyn:
After #670, I think the best option would be to pass some options in `UserData` of the HG object tree.

One of the points I previously marked as a downside, i.e. "it would riddle the code with checks just for handling the UserData.", can actually be used to our advantage, e.g. using an implementation like
```matlab
function [m2t, str] = handleObject(m2t, h, actualHandler)
    assert(isa(actualHandler,'function_handle'));

    UD = get(h, 'UserData');
    UD = sanitizeUserData(UD); % make sure the data is what we expect, remove all the rest
    extraOptions = getExtraOptions(UD);
    [m2t, str] = actualHandler(m2t, h, extraOptions);
    str = [codeBefore(UD), str ,codeAfter(UD)];
end
function [UD] = sanitizeUserData(UD);
    % make sure UD is a struct
    % it contains the right fields (or initialize them to empty defaults)
    % etc. etc.
end
```

That way the checks are concentrated and do not make the code more complex at the current generators. The only thing we have to account for, is that all `actualHandlers` only account for `extraAxisOptions` at the end.

Currently, I don't care anymore about the downside that this requires knowledge of our users of the HG object tree: with great power comes great responsibility.

PeterPablo:
:+1: exactly the approach I would go for. We (only) need to make sure that the user supplied options clash with the ones that are generated by m2t. It probably makes sense to give the user's settings precedence.

egeerardyn:
@PeterPablo : 
Indeed, I would also say that user-provided options should get precedence. Otherwise, the user shouldn't have provided those options in the first place, again "with great power comes great responsibility".

[...]

PR #810, opened 2015-11-06
added extra legend options as argument

eliasm:


egeerardyn:
@matlab2tikzbot: test this PR

okomarov:
@eliasm It would be great if you had a test figure with two subplots with each using a different property :)

egeerardyn:
@eliasm Could you please elaborate a bit why you want to add this?
If I'm not mistaken, you can get exactly the same result by setting
```latex
\pgfplotsset{every axis legend/.append style={...}}
```
either to your LaTeX preamble or by passing something equivalent to the options we already have (e.g. `extraTikzpictureOptions`).

Since the code is clean, I don't mind adding it to the repository, but I hope you understand it will be deprecated when #393 is implemented (which should be in a few months).

src/matlab2tikz.m
```diff
@@ -218,6 +222,7 @@ function matlab2tikz(varargin)
 ipp = ipp.addParamValue(ipp, 'extraCodeAtEnd', {}, @isCellOrChar);
 ipp = ipp.addParamValue(ipp, 'extraAxisOptions', {}, @isCellOrChar);
 ipp = ipp.addParamValue(ipp, 'extraTikzpictureOptions', {}, @isCellOrChar);
+ipp = ipp.addParamValue(ipp, 'extraTikzLegendOptions', {}, @isCellOrChar);
```
miscco: While you at it, can you change the above to extraTikzPictureOptions, to stay consistend. I assume previously tikzpicture was considered one word.

eliasm:
@egeerardyn You are right, the exact same effect can be achieved by:
```
matlab2tikz('plot.tex', 'extraAxisOptions', 'every axis legend/.append style={transpose legend, legend columns=3}');
```
Thanks for looking at it!

Issue #474, opened 2014-12-02
Properly test cleanfigure


egeerardyn:
I recently deleted the test suite for the point reduction algorithm because it was nonfunctional for quite some time and I have the impression that rewriting from scratch would serve us more good than trying the salvage the 50-odd poor lines we had. 

We should re-introduce a function to test `cleanfigure` properly. 
I have some code in a private branch that does part of this: i.e. for my own point reduction algorithm, it plots the before and after together for visual inspection. For manual testing such a thing is good enough, but eventually we might want properly automated unit tests to cover this.

This code is also available in a [gist of mine](https://gist.github.com/egeerardyn/4f45aefc213194ed2185).

PeterPablo:
Pull requests https://github.com/matlab2tikz/matlab2tikz/pull/531 and https://github.com/matlab2tikz/matlab2tikz/pull/533 add ACID tests to partly test `cleanfigure()` (`pruneOutsideBox()` and `movePointsCloser()`). This is certainly far from complete, but at least a start.

egeerardyn:
In the long run, we should actually test `cleanfigure` on its own, i.e. craft
some tests along the lines of:
 - if we have a line with many points, make sure it reduces to the endpoints
 - if we have a line that crosses the borders, make sure it is treated in the right way
 - if we have a line with multiple segments (NaNs in between), make sure that simplification is identical to simplification of each of the segments
 - ...

PR #811, opened 2015-11-09
Fix broken lineSimplify

miscco:
This fixes an embarassing bug, where only the last line segment would be simplified.

Damn, i am feeling really bad, that i screwed it up. Luckily it bit me, when i handled my large data sets

### DIFF
--- a/src/cleanfigure.m
+++ b/src/cleanfigure.m
@@ -472,6 +472,7 @@ function simplifyLine(meta, handle, targetResolution)
         lineStart = find(id_diff == 1);
         lineEnd   = find(id_diff == -1)-1;
         numLines  = numel(lineStart);
+        id_remove = cell(numLines, 1);
 
         % Simplify the line segments
         for ii = 1:numLines
@@ -484,9 +485,12 @@ function simplifyLine(meta, handle, targetResolution)
                 mask      = opheimSimplify(x, y, tol);
                 % Remove all those with mask==0 respecting the number of
                 % data points in the previous segments
-                id_remove = find(mask==0) + lineStart(ii) - 1;
+                id_remove{ii} = find(mask==0) + lineStart(ii) - 1;
             end
         end
+        
+        % Merge the indices of the line segments
+        id_remove = cat(1, id_remove{:});
     end
 
     % Remove the data points
--- a/test/suites/ACID.m
+++ b/test/suites/ACID.m
@@ -209,7 +209,7 @@ function [stat] = plain_cos()
   x = cos(t);
 
   % Explicitely cut the line into segments
-  x([2e5, 5e5, 8e5]) = NaN;
+  x([2e4, 5e4, 8e4]) = NaN;
 
   % Plot the cosine
   plot(t, x);
--- a/test/suites/ACID.m
+++ b/test/suites/ACID.m
@@ -205,10 +205,18 @@ end
 function [stat] = plain_cos()
   stat.description = 'Plain cosine function.';
 
-  fplot( @cos, [0,2*pi] );
-  hold on;
-  
+  t = linspace(0, 2*pi, 1e6);
+  x = cos(t);
+
+  % Explicitely cut the line into segments
+  x([2e5, 5e5, 8e5]) = NaN;
+
+  % Plot the cosine
+  plot(t, x);
+  xlim([0, 2*pi]);
+
   % also add some patches to test their border color reproduction
+  hold on;
   h(1) = fill(pi*[1/4 1/4 1/2 1/2]   ,  [-2 1 1 -2], 'y');
   h(2) = fill(pi*[1/4 1/4 1/2 1/2]+pi, -[-2 1 1 -2], 'y');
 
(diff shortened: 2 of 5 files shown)


miscco:
@matlab2tikzBot test this

egeerardyn:
@miscco Don't feel bad. Actually, if we don't notice such things, that's on all of us. It means we can improve our code reviews, which is where such stupid bugs should be seen (well, also, I think that in the medium-long term we should invest more in unit testing and full test coverage; that way, our testing helps us in this task). So I don't want anybody to feel bad about such stupidities :smile:

miscco:
Yeah, i think the problem is, that the test cases are too small. So for me, luatex simply crashed, because it had to digest a 256MB tikz file. We dont have to go into that extreme, but something along that line might help.

BTW is jenkins and travis broken?

egeerardyn:
Travis runs fine (you can inspect the logs yourself as well). Jenkins also runs fine, but I upgraded my box yesterday, so I think there are a few things that might have been broken during that process. I will check ASAP.

egeerardyn:
Problem was with the upgrade. This tends to fill up my `/tmp` partition which made the machine slow down to a crawl. Also, for Octave 3.8.2, this relies on libhdf5.so.9 which had been removed in favor for libhdf5.so.10 during the upgrade. I installed both versions together now.

You may want to relaunch the tests on Jenkins to ensure everything goes well. Also, on Travis, the problem seems to be a hash mismatch on the `quiverplot` (which I haven't inspected any further).

miscco:
@matlab2tikzBot Please test this.

miscco:
Hm, i have trouble finding the difference, as for me, the test passes in octave 4.0.

miscco:
Here is the difference:

```
3177d3176
< 1.79899415212222      -1.99814575077236\\
3257d3255
< 1.79899415212222      1.99814575077236\\
3261d3258
< 1.99921747738733      -1.99903647765436\\
3265d3261
< 1.99832675050533      -1.79863688077976\\
3337d3332
< 1.99832675050533      1.79863688077976\\
```

src/cleanfigure.m
```diff
@@ -484,9 +485,12 @@ function simplifyLine(meta, handle, targetResolution)
                 mask      = opheimSimplify(x, y, tol);
                 % Remove all those with mask==0 respecting the number of
                 % data points in the previous segments
-                id_remove = find(mask==0) + lineStart(ii) - 1;
+                id_remove{ii} = find(mask==0) + lineStart(ii) - 1;
```
okomarov: At this point, you do not need the `find()` and the `lineStart` shifts, and you can save the index directly. The removeData function should work with the logical index too.

miscco: Hm not sure if this is correct. This is still the individual line segment. So we need to translate the mask to indices and then add the line segments up.

okomarov:
I honestly do not know why I did not see that either...

To avoid this bugs in the future, we might need to adapt a test in the ACID, maybe the simple sine (one of the first ones), to have a few more data points (not too many) and some NaNs in between, so that the line looks still smooth, and does so also after the cleanfigure call.

miscco:
i think the best one would be ACID(17), as there are individual line segments of sine.

miscco:
BTW i dont even know, how often i have changed that quiverplot hash, which always either had those few points in or not.

okomarov:
@egeerardyn Sounds like the quiverplot is unreliable

miscco:
We have NaNs, as the out of plot segments are removed by pruneOutsideBox.

egeerardyn:
Fine by me to mark `quiverplot` unreliable for the time being.

In the long run, we should actually test `cleanfigure` on its own, i.e. craft
some tests along the lines of:
 - if we have a line with many points, make sure it reduces to the endpoints
 - if we have a line that crosses the borders, make sure it is treated in the right way
 - if we have a line with multiple segments (NaNs in between), make sure that simplification is identical to simplification of each of the segments
 - ...

okomarov:
@egeerardyn Can you please add the comments you posted here to the cleanfigure testing suite issue?

egeerardyn:
@okomarov Good point! Done (see #474).

miscco:
Ok i have updated ACID(2) to cut the cosine into 4 segments and updated the hashes.

src/cleanfigure.m
okomarov: You're right, it would need some additional padding for the NaNs that create the splits.

okomarov:
I tested it's all good.
The only thing is the number of points that I would reduce to 100,000 since cleanfigure takes more than 6 seconds to run and we do not want to keep the ACID heavy.

miscco:
Fixed the little test and merged upon current develop to be safe

test/suites/ACID.m
```diff
@@ -205,10 +205,18 @@
...
   stat.description = 'Plain cosine function.';
 
-  fplot( @cos, [0,2*pi] );
-  hold on;
-  
+  t = linspace(0, 2*pi, 1e5);
+  x = cos(t);
+
+  % Explicitely cut the line into segments
+  x([2e5, 5e5, 8e5]) = NaN;
```
okomarov: This now goes way over 1e5.

miscco:
Today is obviously a disguised monday

okomarov:
No worries :smile:, it's just the hash now.

okomarov:
Thanks for this fix :+1:

PR #813, opened 2015-11-10
Also check deviations for plot visibility & drawLine refactoring

### LINKED ISSUE #812: Error bars in bar plot are not displayed
I needed to make a bar plot with error bars and found a nice function called barwitherr on [fileexchange](http://www.mathworks.com/matlabcentral/fileexchange/30639-barwitherr-errors-varargin-). However the error bars are not exported to tikz. I think I found the responsible lines and a manual workaround, but couldn't come up with a complete solution.

In line 134 of the barwitherr.m file, the markers are disabled.
`set(hErrorbar(col), 'marker', 'none')`

Then around line 1400 of matlab2tikz.m, there is a check whether there is anything to plot. As the markers have been disabled and the lines that would connect the error bars have been disabled, the if-condition evaluates to true and the function decides that there is nothing to plot. As a consequence, the error bars are not exported.
```
hasLines   = ~isNone(lineStyle) && lineWidth > 0;
hasMarkers = ~isNone(marker);
if ~hasLines && ~hasMarkers
    return
end
```

As a quick workaround one can set the marker property in the barwitherr function instead of 'none' to '.'. Then matlabt2tikz will write out something like
`\addplot [color=black,only marks,mark=*,mark options={solid},forget plot]`
which gives some fat filled circles as markers. To get rid of the markers, one can set its size to a very small value, e.g.
`\addplot [color=black,only marks,mark=*,mark options={mark size=.25},forget plot]`

Hope this can be resolved easily or someone can make use of the workaround!

egeerardyn:
When a line has deviations, it is always visible.
This should fix #812.

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -1649,7 +1649,7 @@ function [m2t, str] = drawLine(m2t, h, yDeviation)
     % - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
     [data] = getXYZDataFromLine(m2t, h);
 
-    % Also add
+    % Also add errorbars to the data when needed
     if hasDeviations
         data = [data, yDeviation(:,1:2)];
     end
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -1612,22 +1612,23 @@ function [m2t, str] = drawLine(m2t, h, yDeviation)
 % This is an extremely common operation and takes place in most of the
 % not too fancy plots.
     str = '';
+    if ~exist('yDeviation','var')
+        hasDeviations = false;
+    else
+        hasDeviations = true;
+    end
+    %TODO: get rid of yDeviation parameter, fetch actual data instead
 
-    % Check if there is anything to plot
-    lineStyle     = get(h, 'LineStyle');
-    lineWidth     = get(h, 'LineWidth');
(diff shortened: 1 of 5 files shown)


okomarov:
Can we add/edit an ACID test to have something like:

    errorbar(1:3,[1,1,1],'LineStyle','none')

egeerardyn:
@okomarov: perhaps not a bad idea.

src/matlab2tikz.m
```diff
@@ -1613,18 +1613,15 @@ function assertRegularAxes(handle)
 % not too fancy plots.
     str = '';
 
-    if ~isVisible(h)
-        return
-    end
-
```
miscco: Generally i am rather fond of exiting early. I am sure, that it has no runtime effect at all, but is there any particular reason you put it in the end?

egeerardyn: It's hardly put in the end. Personally, I rather have 1 "if" per concept instead of first checking visibility and afterwards do a more thorough check on visibility (by means of markers and so on).

egeerardyn:
I cleaned up the code a bit, since the `drawLine` function looked ugly and did a few too many things.
To address the core of @miscco 's comment (i.e. return early), I extracted that check into another function, so now it reads quite a bit cleaner IMHO.

miscco:
:+1: Very nice

Issue #506, opened 2014-12-31
Tools for regression testing


egeerardyn:
We should provide some scripts to do regression tracking.
Currently, we have some MD5-based regression tests in place on Travis, but that only indicates that something is wrong. It does not help much in tracking down what or where something is wrong.

What would be awesome to have is a more in-depth analysis of those results. A possible workflow would be
1. `git checkout before`
2. Run test suite and produce all LaTeX files, PDF report, ...
3. Move those file to some safe location
4. `git checkout after`
5. Run test suite and produce all LaTeX files, PDF report, ...
6. Compare the files of `before` and `after` (the easiest is probably text-based diffing of the output)

Specifically steps 3 and 6 should be part of such a tool, but encapsulating the whole workflow would be a nice extension. Another wild idea would be integration with `git bisect` (or similar features).

okomarov:
I feel that we really need this badly...I was thinking about the approach. I do not have enough time/knowledge to develop myself but I can share my two cents.

Take a **safe** and **try catch** approach. Try to checkout before, if there are uncommitted changes just abort for now (no fancy stashing or similar).

Also, on my MinGW (Win), the `make -j#` syntax for some reason interprets jobs as the total number of jobs, so I just do something like `make -j300` and usually there is a figure that fails on the first run but not on the second...(we gotta fix that surf legend sometime).

Issue #321, opened 2014-03-03
problems with acidtest


bene86:
Hi,

on a complete new installation of Windows/Matlab I wanted to try the acidtest as explained in the readme file but I hat three problems.
1. the folder test/data needed to be created manually by me, otherwise there was an error.
2. eps2pdf did not find my ghostscript installation. I had to change line 164 in matlab2tikz_acidtest.m to eps2pdf(eps_file,'c:\Program Files\gs\gs9.10\bin\');
3. eps2pdf still did not work, thats why I also had to change the file eps2pdf.m. I inserted fullGsPath = [fullGsPath 'gswin64c.exe']; after line 79

Is any of this known?

PR #814, opened 2015-11-12
tools for regression testing

okomarov:
This is a dumber version of #506 (Tools for regression testing), which produces .pdf for comparison at two specified commits. Especially lacks point 6. in the reference above.

If the direction is correct, we can extend functionality in this PR.


test/regressionTest.m
```diff
@@ -0,0 +1,57 @@
+function regressionTest(commitBefore,commitAfter)
+    % Produce graphical output for two commits
+
+    % Work only with a clean working tree
+    [status, cmdout] = system('git status --porcelain');
+    if ~isempty(cmdout)
+        error('regressionTest:treeNotClean','Working tree is not clean.')
+    end
```
miscco: Maybe here we could opt to do a reset --hard in case of a dirty tree?

test/regressionTest.m
```diff
@@ -0,0 +1,57 @@
+function regressionTest(commitBefore,commitAfter)
```
miscco: So is this intended for developers during hacking or as a check against develop? In the later case i would omit the second argument and call the first branchName

test/regressionTest.m
egeerardyn: While technically correct, let's not take such drastic measures. In git itself, all operations will warn when the tree is dirty. We should do the same. At most, I think we can show a message akin to "please clean your working copy using `git reset --hard`".

test/regressionTest.m
egeerardyn: It can be used for both. 
But mostly this is to test whether a regression is present between commit X and Y. I wouldn't hardcode the `develop` branch (at most as the default argument when for `commitBefore`).

miscco:
Also to maybe ease the burden on the developer, we can do some imagemagic
http://www.imagemagick.org/script/compare.php

we would generate the pdf, convert to jpg and compare both results with -metric PSNR. That would help the poor eyes of the developer

egeerardyn:
@miscco That's a really good idea (except for JPG: ugly lossy compression; it's far cleaner to use PNG).

If short on time, I think we can do that in another PR (since the file management is quite a bit more complicated to implement that than what is in this PR).

egeerardyn:
:+1: on the code. I haven't tested it in practice, I will try to test once the WiP label is gone.

okomarov:
Good stuff the imagmagick!

This PR is at a very early stage and dunno when I'll be able to develop more, but the plan is:

- instead of graphical do the `runTestMatlab2tikz` on whole ACID first, save data output in a separate folder
- re-do the same for other commit
- git diff test by test
- isolate diffed, and produce some graphical output (and here we can add magick diffing)

The reporting part can be improved, but at the moment I am thinking to re-use the existing infrastructure and dump the diff outputs together with the testGraphical of the problematic test (those that don't have matching hashes).

test/regressionTest.m
okomarov: I would actually not suggest anything, since I dunno what the developer is up to. He might be working on a separate feature and just did not commit all changes yet.

okomarov:
The whole tool is intended to robustly identify what induces changes in the hashes when a new PR is submitted.

egeerardyn:
@okomarov With respect to the testing infrastructure: I think one of the current major limitations is that it doesn't allow to output to a different directory.

That would allow a few things:
 - parallel testing of different environments on the same checkout without any race conditions
 - move the complexity of file management from reporting tools (as this PR) and the developer (during manual testing) towards the testing framework where it would be more well-contained.

Essentially, a `diff` operation would then become something like
```matlab
git checkout before
testM2T(..., 'output','~/before')
git checkout after
testM2T(..., 'output','~/after')
makeDiffReport('~/before', '~/after')
```

test/regressionTest.m
egeerardyn: Equally fine :+1:

miscco:
Actually, we could also use the externalize library. for tikz. That way we could change testGraphical, to directly generate an png per test rather than one huge pdf for the whole suite (or both for all thats worth).

okomarov:
@egeerardyn agree, I am aware of the limitation but thought of taking the workaound of moving things afterawrds. However, your suggestion is definitely the way to go and it would be great if you have time to adapt testM2t to accept an output directory input.

egeerardyn:
@miscco If we rely on imagemagick already, I don't think we gain anything from using `external` (since that produces PDFs that we would need to convert anyhow).

The current workflow for `testGraphical` is actually:
 - produce a standalone TikZ file per test
 - produce the ACID TeX file with references to the compiled files
 - use the make file to compile each of the figures and then the report (akin to what you could get with `external`)

I certainly agree that we can adapt our testing framework to make outputting to PNG and so on easier (almost a year ago, this has been discussed as well: #321, #326).

miscco:
So i am unsure, whther you would like to address the imagemagick stuff here. I am unfortunately unfamiliar with the test suite . So my current glance at this let to the following idea.
1. Add the following to execute_plot_stage
```matlab
set(gcf, 'PaperUnits', 'Points');
status.plotStage.Dimension = ceil(get(gcf, 'PaperPosition'));
```
 This should get the actual printing dimensions of the figure

2. Create a execute_convert_stage function that
2.a checks whether the pdf is there
2.b converts the individual pdfs with
```matlab
system(['convert -density ',num2str(status.plotStage.Dimension(3),'x',
                            num2str(status.plotStage.Dimension(4)' test',
                            num2str(i),'-converted.pdf test',
                            num2str(i),'-converted.png]);
```

egeerardyn:
Related to the image diffing, I don't know if you guys have the Image Processing toolbox available (I have), but that would allow to do very similar things as the `compare` tool from `imagemagick`, e.g.

```matlab
function PSNRVal = compareAndPlot(file_before, file_after) 
    
    %% load files
    if ~exist('file_before','var') || isempty(file_before)
        file_before = 'pears.png';
    end
    before = im2double(imread(file_before));

    if ~exist('file_after','var') || isempty(file_after)   
        after  = imfilter(before,fspecial('gaussian'));
    else
        after  = im2double(imread(file_after));
    end
    
    %% compute differences
    % let's compute 1 - |a-b|, such that when a==b, the pixel value is 1
    % i.e. white. That way, differences show up as darker pixels. This seems
    % also more visible to the human eye the actual difference.
    delta  = imabsdiff(after, before);
    delta  = 1-delta./max(delta(:));
    
    % also compute a version with histogram equalization
    deltaBW = repmat(histeq(rgb2gray(delta)),1,1,3);
    
    %% metrics on the difference
    PSNRVal = psnr(after, before);
    [SSIMVal, similarity] = ssim(after, before);
    similarity = histeq(rgb2gray(similarity));

    %% produce figure
    figure;
    h(1) = subplot(3,3,[1 4 7]);
    imagesc(before);
    title('Before');
    
    h(2) = subplot(3,3,2);
    imagesc(delta)
    title(sprintf('Diff (PSNR = %g dB)', PSNRVal));
    
    h(4) = subplot(3,3,5);
    imagesc(deltaBW); colormap('gray')
    title('Diff (histogram equalized)');
    
    h(5) = subplot(3,3,8);
    imagesc(similarity); colormap('gray')
    title(sprintf('Structural Similarity (%g)', SSIMVal));

    h(3) = subplot(3,3,[3 6 9]);
    imagesc(after);
    title('After')

    linkaxes(h)
    
end
```

The plot looks something like this (for the default arguments):
screen shot 2015-11-18 at 16 25 44


Note that the differences are barely visible on the images themselves, since it is just a smoothing operator that I apply to the demo case.

Also, calculating `ssim` seems to be a lot slower than just the PSNR, so perhaps we should make it optional if we use this.

miscco:
I found the PSNR metric to be rather unsuited. In imagemagick there is the AE metric, that basically just tells you whether some pixels are different. That should be quite what we need.

For the toolbox i believe we have it, but i would rather not utilize an expensive toolbox that is matlab only when we have an opensource alternative, that works as well.

egeerardyn:
@miscco I agree with you that we shouldn't rely on the Image Processing Toolbox as the only mechanism, but I think MATLAB offers the great advantage that it is platform-independent (calling external tools from within MATLAB and Octave is often a cause of pain, in my experience).
I also suspect that Octave has similar routines (on MacPorts I see the `octave-image` package from OctaveForge).

I think the advantage of using MATLAB for comparison would be that we are more familiar with MATLAB (at least I am), such that we can do more advanced/fine-grained comparisons if we want to or need to. E.g. for the tests I have done, the `SSIM` seems to be a nice one.

With regard to choice of metric, I completely agree that PSNR is not an ideal one for all plots. It might be nice for photographs (or in our case: plots with color gradients if we can neglect the borders), but for plots with sharp edges it is hard to interpret.

I think the downside of something like [AE](http://www.imagemagick.org/script/command-line-options.php#metric) is that it is non-normalized and hence depends on the number of pixels used. It basically tells: "20 pixels are different", which tells little without knowing the size of the figure. Also, for figures with gradients it might be a bit too strict (akin to the hashes we currently use).

Probably the best strategy is to report multiple metrics for a well-balanced evaluation of different figures.

test/regressionTest.m
miscco: Fine with me too :+1:

miscco:
I had send around internally a simple bash script that generated the diffs between develop and a current branch. Could we adopts this here?

I thing testGraphical is in most cases an overkill, as it is hard to find a difference and when you find it, you always have to check the plain text diff too.

In general, i would love to have it as a matlab file rather than a bash script

egeerardyn:
@miscco MATLAB scripts are fine. However, the script you had immediately ran for all environments. That would become much more complicated to do from within MATLAB, I think.

okomarov:
I currently have no time to work on the PR

okomarov:
At the moment, the routine should produce a latex report with a 2x2 matrix of figures. In the first column, the reference figures generated by matlab, and the second column the m2t converted figures. On the first row, the first commit (current or base) and on the second row the other commit (comparison). 

Nothing fancy, but should help with visualizing regressions by automating all the git checkouts and make-ing of pdfs.

test/makeLatexReportRegression.m
```diff
@@ -0,0 +1,231 @@
+function makeLatexReportRegression(currentStatus, output, otherStatus)
+% generate a LaTeX report
```
egeerardyn: Could do with a tiny bit more of documentation, something like:
> MAKEKATEXREPORTREGRESSION makes a regression report in LaTeX
>  - currentStatus: ...
>  - output: output directory to store the report in
>  - otherStatus: ...

test/makeLatexReportRegression.m
```diff
@@ -0,0 +1,231 @@
...
+%
+% 
+    if ~exist('output','var')
```
egeerardyn: I think it's a bit weird to have `output` optional, while `otherStatus` is a required option. I would make `output` the last option to provide such that a defaulted value makes sense.

test/makeLatexReportRegression.m
```diff
@@ -0,0 +1,231 @@
...
+        output = m2troot('test','output','current');
+    end
+    % first, initialize the tex output
+    SM = StreamMaker();
+    stream = SM.make(fullfile(output, 'regression.tex'), 'w');
+
+    texfile_init(stream);
+
+    printFigures(stream, currentStatus, otherStatus);
+    printSummaryTable(stream, currentStatus);
```
egeerardyn: I think it's a bit weird to only show detailed stuff (summary, error messages, env info) for the current status. Wouldn't it make sense to show it also for the other (old?) situation? E.g. to see when we fix an unreliable test, introduce a new one, ...?

I think eventually we could do something similar to those parts as what you have done with the figures; but in the meanwhile just both overviews separately would be more than sufficient.

test/testGraphical.m
```diff
@@ -38,9 +38,9 @@
     %standalone mode!
     status = execute_type_stage(status, ipp);
 
-    if ~status.closeall && ~isempty(status.plotStage.fig_handle)
+    try
```
egeerardyn: Why `try`-`catch` instead of those explicit checks?
This would make the `closeall` option moot (which is fine by me), but I would like to understand why this approach is superior.

okomarov: Basically, there is a problem somewhere of the figure being already closed! This issue has already affected `testHeadless()`. In fact, there is a `try catch` there, together with if condistions ... and the code doesn't look very nice.

test/makeLatexReportRegression.m
okomarov: I agree, just wanted to get the core of the report out...

test/makeLatexReportRegression.m
okomarov: Actually, the function signature as is, would help merge this duplication of code into `makeLatexReport`, so left it out as is for now.

test/makeLatexReportRegression.m
okomarov: I want to see if this is useful and picks up use at all. 

We can already use it to inspect failed tets of existing PRs. The worklfow would be:

 -  Check out this PR
 -  run `regressionTest(commit to test against, commit of target PR, failed test indices)`

PR #815, opened 2015-11-13
Rework pixelate

miscco:
Currently pixelate is kind of buggy.

The reason is, that it assumes an explicit order in both dimensions, hidden in the diff(...). Instead use the matlab build in function unique, that finds us the unique pixels without any hassle.

P.S: It is not 100% mathematically correct, as there is the theoretical possibility, that the first and last pixel is duplicated *shrug*.

### DIFF
--- a/src/cleanfigure.m
+++ b/src/cleanfigure.m
@@ -554,18 +554,24 @@ function mask = pixelate(x, y, xToPix, yToPix)
     % The resolution is lost only beyond the multiplier magnification
     mult = 2;
 
-    % Convert data to pixel units, magnify and mark only the first
-    % point that occupies a given position
-    mask = [true; diff(round(x * xToPix * mult))~=0];
-    mask = [true; diff(round(y * yToPix * mult))~=0] | mask;
+    % Convert data to pixel units and magnify
+    dataPixel = round([x * xToPix * mult, ...
+                       y * yToPix * mult]);
 
-    % Keep end points or it might truncate whole pixels
+    % Find the unique pixels
+    [~, id_unique, ~] = unique(dataPixel, 'rows');
+
+    % Assume every pixel to be a duplicate
+    mask = false(size(x));
+
+    % Set the first, last, as well as unique pixels to true
+    mask(1)         = true;
+    mask(end)       = true;
+    mask(id_unique) = true(size(id_unique));
+
+    % Set NaNs to true
     inan         = isnan(x) | isnan(y);
-    df           = diff([false; inan; false]);
-    istart       = df == 1;
-    pend         = find(df == -1)-1;
-    mask(istart) = true;
-    mask(pend)   = true;
+    mask(inan)   = true;
 end
 % =========================================================================
 function mask = opheimSimplify(x,y,tol)
--- a/src/cleanfigure.m
+++ b/src/cleanfigure.m
@@ -558,16 +558,19 @@ function mask = pixelate(x, y, xToPix, yToPix)
     dataPixel = round([x * xToPix * mult, ...
                        y * yToPix * mult]);
 
-    % Find the unique pixels
-    [~, id_unique, ~] = unique(dataPixel, 'rows');
+    % Sort the pixels
+    [dataPixelSorted, id_orig] = sortrows(dataPixel);
 
-    % Assume every pixel to be a duplicate
-    mask = false(size(x));
+    % Find the duplicate pixels
+    mask = [true; diff(dataPixelSorted(:,1))~=0 | ...
+                  diff(dataPixelSorted(:,2))~=0];
+
+    % Unwind the sorting
+    mask = mask(id_orig);
 
     % Set the first, last, as well as unique pixels to true
     mask(1)         = true;
     mask(end)       = true;
-    mask(id_unique) = true(size(id_unique));
 
     % Set NaNs to true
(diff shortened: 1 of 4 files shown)


miscco:
The change in ACID 11 is the following:

```
<<<0.991672129152322	-0.0999999999999999\\	
```

So one Pixel is gone, maybe we need to add some more.

okomarov:
When you say is buggy, do you have an **example** showing where the `diff()` fails? Because all it does, it takes the last occurrence in a repetition, while the unique takes the first, hence it's just an **assumption**.

Also, unique is much **slower**.

Finally, I understand updating the hashes has got into the routine but in this case you cannot just dismiss an incongruence, which probably arises from unique taking first occurrence.

miscco:
So take the following pixel values
```
1    1
2    2
3    3
1    1
```

The current code doesnt catch this, as it assumes that xData and yData are strictly ascending.

okomarov:
It's doing `diff(...) ~= 0` so it does catch it.

miscco:
Hm i am unsure, whether we are on the same page. What i meant is that in the sequence above the 4th pixel will not be removed, although it is a duplicate.

miscco:
So the reason I catched this is that i have an unsreasonable large data set (1e6), that is really noisy. Therefore lineSimplify doesnt do anything. Therefore i though about manually using pixelate and found, that of those 1e6 data points it merely scratched a couple hundreds.

okomarov:
I see, now I get your point. I propose to test with issorted on x and if not sort first the data saving the sorting indexes, run diff on the sorted data, and sort back with the indices the mask.

How long pixelate takes with unique on your dataset?

miscco:
Actually i havent tested yet as i am at home. Sadly sort() is not applicable, as it sorts the rowd independently and doesnt give a guarantee about the ordering of the second column.

The problem i have is that it looks like that:

```
0    1
0    5
0    6
0    2
0    1
```

So neither  sorting for any column helps. One would need biconditional sorting. As i saw unique i kind of simply fell for that without checking for speed. I kind of assume, that in the background it does that kind of sorting.

miscco:
There seems to be 
```matlab
[tblB,index] = sortrows()
```
That sorts an cell array. However, we would have to do
```matlab
data = mat2cell([x, y], numel(x),2);
[~,id_unique] = sortrow(data);
```

miscco:
Hm dont know what octave is doing. I will have a look on it on monday probably.

src/cleanfigure.m
```diff
@@ -554,18 +554,28 @@ function simplifyLine(meta, handle, targetResolution)
...
+    mask(1)         = true;
+    mask(end)       = true;
+
+    % Set NaNs to true
     inan         = isnan(x) | isnan(y);
-    df           = diff([false; inan; false]);
-    istart       = df == 1;
-    pend         = find(df == -1)-1;
-    mask(istart) = true;
-    mask(pend)   = true;
```
okomarov: If i remember correctly you can have multiple start end due to NaN splits.

miscco: Actually i never understood the stasrt and end concept. It doesnt matter if a pixel is at the start of the data or not. The only impotant thing is, that it is occupied. That is also the only thing we are checking against.

okomarov: I think the intuition I got about this point is that the plot coordinates do not necessarily start at exactly the beginning of the pixel. So, if you have two plot coordinates which are within one pixel distance from each other, but they start in the middle of the pixel, then you will lose one of those, while you should have retained it. 

It is basically a coordinate reference problem and I just chose the safe option of keeping the ends of a line segment. In the end, having two additional coordinates per line segment is nothing compared to the sort bug you found.

miscco: @okomarov thx ;)

I think the easy solution is to pass
```
xData - xLim(1)
yData - yLim(1)
```
This should ensure, that everything is in the correct pixel, as it eliminates negative values, which could possibly come from different rounding for negative/positive values.

okomarov: I don't think this will solve the supposed discrepancy between the coordinate systems of the axis and the monitor. What we do, is we translate the axis units to pixels (to match the screen units), but we do not translate the origin of the axis to *match* the origin of the screen. 

Or maybe I am wrong and it is actually an approximation error around the origin. I'd say let's try and we can get rid of the dirty workaround.

miscco: I dont think, that this are screen pixels. What we are giving the routing are the dimensions of the *image*. As far as i understand the code, it associates all data points thei thepixel coordinates from the range [0, W] and [0, H]. Therefore, it should be completely independent of screen pixels.

okomarov: You are right :)

miscco:
So i couldnt find any difference to develop in my Octave 4 setup, which is rather strange.

miscco:
So i cant find any difference on my machine did someone else find something?

miscco:
Ping

egeerardyn:
@miscco I'm sorry I haven't had the time to look at this yet. I will try during the weekend.

egeerardyn:
Result looks fine to me:
develop and this PR on the same figure:
screen shot 2015-11-28 at 22 02 27

(red dots is one version, blue circles are the data points of the other).

I think this can be merged :+1: Thank you for tackling this!

PR #817, opened 2015-11-17
Put text removal into own function and prepare deprecation

miscco:
Hi,

as the title says, this puts text removal into its own subfunction, considerably cleaning up revursiveCleanup.

While we are at it, removed the intent parameter from recursiveCleanup, that might have some historic origin. Additionally, the early return for quiverplots seems to be nonfunctional. Remove it was well.

### DIFF
--- a/src/cleanfigure.m
+++ b/src/cleanfigure.m
@@ -6,6 +6,9 @@ function cleanfigure(varargin)
 %   CLEANFIGURE('handle',HANDLE,...) explicitly specifies the
 %   handle of the figure that is to be stored. (default: gcf)
 %
+%   CLEANFIGURE('pruneText',BOOL,...) explicitly specifies whether text
+%   should be pruned. (default: true)
+%
 %   CLEANFIGURE('targetResolution',PPI,...)
 %   CLEANFIGURE('targetResolution',[W,H],...)
 %   Reduce the number of data points in line objects by applying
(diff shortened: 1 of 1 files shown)


src/cleanfigure.m
```diff
@@ -77,47 +81,25 @@ function cleanfigure(varargin)
...
 
-    %display(sprintf([repmat(' ',1,indent), type, '->']))
-
-    % Don't try to be smart about quiver groups.
-    % NOTE:
-    % A better way to write `strcmp(get(h,...))` would be to use
-    %     isa(handle(h), 'specgraph.quivergroup').
-    % The handle() function isn't supported by Octave, though, so let's stick
-    % with strcmp().
-    if strcmp(get(h, 'Type'), 'specgraph.quivergroup')
```
egeerardyn: This might be related to HG1/Octave, not HG2 on which you are testing (where it indeed may be dead code).

I don't think we should remove it this hastily and certainly not mixed in with a lot of unrelated edits in the same commit.

miscco: True,

i was in the "Burn it with fire" mode

miscco: So in octave 4 this is also dead code.

egeerardyn:
:+1: on removing `indent` which seems to not be used anymore (or only during debugging). 
:-1: on doing quite a lot in a single commit. E.g. for quivers I'm quite sure this does something on HG1 (in HG1, quivers are stored as line objects with some internal structure). Probably it doesn't matter that much, since the internal structure will likely prevent any cleanup anyway, but we should really document that behavior instead to avoid that we add it back later.

If I remember correctly, an old-style quiver has a few line objects that break down: (1) the starting point marker, (2) the arrow body and (3) the arrow head. Since everything is quite simple, there is no room for optimization as far as I see; so it doesn't harm to break out early.

miscco:
Yeah, so i have put the quiverplot check back in. 

On the other side im kind of reluctant to break it into smaller patches, as i would have to manually revert the code, and start over with a fresh branch. I should keep in mind that moving then editing is the way to go.

egeerardyn:
@miscco You don't have to break it into smaller patches anymore (it isn't worth that hassle). 
Please just remember `git add --patch` for the future :smile:

PR #818, opened 2015-11-17
Parametrize output directory in testing

### LINKED ISSUE #816: Output directory parameter for testing
Currently (e.g. as discussed in #321, #326 and #814), there are some limitations to the testing framework that impede quick and easy iterations for us developers.

We should be able to set the output directory for use in testing. That would allow
   * parallel testing on a single checkout using different environments (mainly of importance on developer machines) without nasty race conditions
   * to move complexity from developer or reporting tools towards the testing suite. Currently, this complexity is basically file management (moving those files around by hand, getting outdated files out of the way, ...), but this is cumbersome, error-prone and mind-numbing.

egeerardyn:
This implements/fixes #816.

Idea of the commits:
 - passing an `output` parameter changes the output directory (default value is `test/output/current` to allow developers to place different test outputs in the `test/output` directory for good overview).
 - all over the code, I use absolute paths (or using the newly introduced `m2troot` function which effectively does the same).

Given the last point, I removed the directory change in the beginning and at the end, which were essentially a work-around for our lack of absolute paths in MATLAB.


egeerardyn:
@okomarov There you go: now one can specify the output directory for testing (I hope this can be made to merge with your latest PR #814 without much troubles)

I also found a few functions that we have but don't actively use. For `cleanFiles`, I will look into it in the near future since it is useful.
In the meanwhile, I appreciate any and all comments :smile:

Overall I'm quite pleased with the implementation, the only downside is that the output directory also has to be passed to `makeLatexReport`, but that is a necessary evil I guess.

okomarov:
I haven't tested but the code looks good.

egeerardyn:
I take the last comment as a :+1:.
In the meanwhile I have removed some sloppy things and inaccuracies of mine.

Unfortunately, this also means that #814 might need some changes before we can merge that PR.

Issue #820, opened 2015-11-24
pi in ticklabels are not exported in mathmode


murmlgrmpf:
I have a plot that uses multiples of pi as yticklabels. When I export the plot using matlab2tikz, it is not putting the \pi in math mode, what causes latex to throw an error.

Fix: parse ticklabels for latex strings that need to be in math mode

egeerardyn:
Could you please provide us with a bit of code to reproduce this problem?
Could you also tell what version of MATLAB/Octave you are using (tick labels and especially their interpeter have changed quite a bit during the last few versions).

murmlgrmpf:
I am running on Matlab r2015b. The following code reproduces the issue:

'''
ang = linspace(0,2*2*pi,100);
amp = sin(ang);

plot(ang,amp)

ax = gca;
ax.XTick =[0, pi,2*pi, 3*pi,4*pi];
ax.XTickLabel = {'0', '\pi','2\pi','3\pi','4\pi'};

matlab2tikz('test.tex','standalone',true,...
    'parseStringsAsMath',true,'interpretTickLabelsAsTex',true);
'''

murmlgrmpf:
leaving out the options does not seem to change the behavior.

murmlgrmpf:
I have tinkered around with the options 'interpretTickLabelsAsTex' 'parseStrings' and 'parseStringsAsMath' but no combination fixes the issue.

murmlgrmpf:
I have a workaround: put $\pi$ in the ticklabels fixes matlab2tikz export, but breaks the matlab plot.
Therefore I still think it would be better to fix this in matlab2tikz.

egeerardyn:
I can confirm that this happens using the code from the `master` branch (i.e. what you get when you download from the FileExchange), however, this bug is already solved in the current `develop` branch. 

So you can either wait for the next release (may take a few weeks/months) or download the [ZIP file from our current develop branch](https://github.com/matlab2tikz/matlab2tikz/archive/develop.zip), which is quite reliable, but not as well-tested as the version you find on FileExchange.

Thank you for reporting it, anyway!

PR #821, opened 2015-11-24
Tick label interpreter + cleanup

egeerardyn:
In #820, I noticed that the `interpretTickLabelsAsTex` parameter is no longer used anywhere within the code.

However, for R2014a and older, it still makes sense to use it.

In this PR, I have reintroduced the parameter via the default value (which is used in HG1). In HG2, nothing changes and the parameter is ineffective, hence, the warning procedure.

I also took a bit of time to clean up some bits of code in the vicinity: those should be functionally identical.

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -1508,7 +1508,7 @@ function interpreter = defaultTickLabelInterpreter(m2t)
     % determines the default tick label interpreter
     % This is only relevant in HG1/Octave. In HG2, we use the interpreter
     % set in the object (not the global default).
-    if m2t.cmdOpts.interpretTickLabelsAsTex
+    if m2t.cmdOpts.Results.interpretTickLabelsAsTex
         interpreter = 'tex';
     else
         interpreter = 'none';
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -1539,33 +1539,29 @@ function options = setAxisTicks(m2t, options, axis, ticks, tickLabels,hasMinorTi
     % a reasonable default.
     matlabDefaultNumMinorTicks = 3;
     if ~isempty(ticks)
-        options = opts_add(options, ...
-            [axis,'tick'], sprintf('{%s}', ticks));
+        options = opts_add(options, [axis,'tick'], sprintf('{%s}', ticks));
     end
     if ~isempty(tickLabels)
         options = opts_add(options, ...
             [axis,'ticklabels'], sprintf('{%s}', tickLabels));
     end
     if hasMinorTicks
-        options = opts_add(options, ...
-            [axis,'minorticks'], 'true');
+        options = opts_add(options, [axis,'minorticks'], 'true');
         if m2t.cmdOpts.Results.strict
-            options = ...
-                opts_add(options, ...
+            options = opts_add(options, ...
                 sprintf('minor %s tick num', axis), ...
                 sprintf('{%d}', matlabDefaultNumMinorTicks));
         end
     end
+
     if strcmpi(tickDir,'out')
-        options = opts_add(options, ...
-            'tick align','outside');
+        options = opts_add(options, 'tick align', 'outside');
     elseif strcmpi(tickDir,'both')
-        options = opts_add(options, ...
-        'tick align','center');
+        options = opts_add(options, 'tick align', 'center');
     end
+
     if isDatetimeTicks
-        options = opts_add(options, ...
-        ['scaled ' axis ' ticks'],'false');
+        options = opts_add(options, ['scaled ' axis ' ticks'], 'false');
     end
 end
 % ==============================================================================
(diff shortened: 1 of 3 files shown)


src/matlab2tikz.m
```diff
@@ -1460,17 +1463,7 @@ case guitypes()
...
-        dtsManager = get(handle, 'DatetimeDurationPlotAxesListenersManager');
-        oldState   = warning('off','MATLAB:structOnObject');
-        dtsManager = struct(dtsManager);
-        warning(oldState);
-
-        isDatetimeTicks = dtsManager.([upper(axis) 'DateTicks']) == 1;
-    catch
-        isDatetimeTicks = false;
-    end
+    isDatetimeTicks = isAxisTicksDateTime(handle, axis);
```
miscco: Sweet, i like the functions a lot

src/matlab2tikz.m
```diff
@@ -1519,33 +1539,29 @@ case guitypes()
     % a reasonable default.
     matlabDefaultNumMinorTicks = 3;
     if ~isempty(ticks)
-        options = opts_add(options, ...
-            [axis,'tick'], sprintf('{%s}', ticks));
+        options = opts_add(options, [axis,'tick'], sprintf('{%s}', ticks));
```
miscco: Please keep the line as it is, since you do not change it below

src/matlab2tikz.m
```diff
@@ -1519,33 +1539,29 @@ case guitypes()
...
     end
     if ~isempty(tickLabels)
         options = opts_add(options, ...
             [axis,'ticklabels'], sprintf('{%s}', tickLabels));
     end
     if hasMinorTicks
-        options = opts_add(options, ...
-            [axis,'minorticks'], 'true');
+        options = opts_add(options, [axis,'minorticks'], 'true');
```
miscco: Why is there no sprintf?

miscco: Nevermind im dumb. Still i would prefer to put it in a separate line

src/matlab2tikz.m
egeerardyn: Thanks!

src/matlab2tikz.m
egeerardyn: There is no need to break those lines, so I rather have everything on the same line if we can since it is slightly better for readability IMHO.

miscco: True, but in nearly all the rest of the code there is. So you generally have the pattern
```
         options = opts_add(options, ...
             [axis,'ticklabels'], sprintf('{%s}', tickLabels));
```

Therefore, in total it is more consistent to write it that way, so that  the actual command is on a single line. But this is advanced nitpicking :smiling_imp:

okomarov: The newline is really legacy stuff, probably when monitors were still not wide enough. I asked my self several times why we split it in the first place, but you know, there really is no incentive in beautyfying the whole code, unless you are already working on it.

egeerardyn: I agree that the added beauty is minimal. But I always had the same impression as you had: "why is that newline there?". During working on this PR, I scrolled over this part, it bothered me a few times too often, so I changed it to avoid it bothering me in the future.

Personally, I think it still makes sense to limit the width of a line: the longer a line becomes, the harder it becomes to keep track of each different line. Indirectly, it limits the indentation depth within a function (i.e. complexity). 

On modern wide screens, you indeed have all the width you could wish, but limiting that makes that you can have 2 windows next to each other.
Alternatively, you can turn such a screen and get 80 chars per line and about 80 lines as you can see below:
screen shot 2015-11-30 at 16 25 24


Personally, I avoid going over 80 chars unless readability of the alternative is worse.

miscco: So i would aggree to change it as you have and clean up the rest when someone has some spare time.

src/matlab2tikz.m
```diff
@@ -1519,33 +1539,29 @@ case guitypes()
...
         if m2t.cmdOpts.Results.strict
-            options = ...
-                opts_add(options, ...
+            options = opts_add(options, ...
                 sprintf('minor %s tick num', axis), ...
                 sprintf('{%d}', matlabDefaultNumMinorTicks));
```
miscco: While your at it could you merge the two sprintfs? It is a single instruction and should be inside one therefore.

miscco:
I really like the PR. I have one last minor nitpick about the two sprintfs, however that is rather optional.

src/matlab2tikz.m
egeerardyn: These two cannot be merged, as the code above is equivalent to:
```matlab
key = sprintf('minor %s tick num', axis);
val = sprintf('{%s}', matlabDefaultNumMinorTicks);
options = opts_add(options, key, val);
```

miscco: Ah true

egeerardyn: But the idea is quite valid: all those `sprintf`s can be a nuisance. 

I have some ideas in the back of my head to simplify those axis-related `sprintf`s away one day, but that's for another time.

egeerardyn:
I read the last comment as a :+1:, so merging this PR in.

src/matlab2tikz.m
miscco: maybe just make a  new issue, no need to do it alone

Issue #788, opened 2015-10-12
Test if lineStyles in pruneOutsideBox matter


miscco:
In pruneOutsideBox the following code is commented:
```matlab
%hasLines = ~strcmp(lineStyle,'none') && lineWidth>0.0;
%hasMarkers = ~strcmp(marker,'none');
```

Check whether this actually makes sense.

%TODO: test again with lines uncommented (see #787 and ce899382)

miscco:
So i made a little example, that should show why the current code is wrong.

```matlab
ydata_marker = [4, 4];
ydata_line   = [1, 1];
xdata = [-1, 1];
plot(xdata, ydata_marker, '*');
hold on;
plot(xdata, ydata_line);
hold off;
ylim([0, 5]);
xlim([0, 2]);
cleanfigure
pause;
xlim([-2, 2]);
```

Note, that normally one would expect, that the left marker would be removed, as it is not visible. This would be fixed with:
```matlab
  hasLines = ~strcmp(get(handle, 'lineStyle'),'none');
  hasMarkers = ~strcmp(get(handle, 'marker'),'none');
```
I am currently thinking about simplifying cleanfigure somehow, looking at #789.

PR #824, opened 2015-12-03
Check for lines in pruneOutsidebox

miscco:
Previously we didn't check whether it was a line or not #788. This might have the undesired behaviour, that markers outside of the visible box would not be removed.

By checking for lineplots, we can resolve this. Additionally, as markers are only plotted inside the box, we can simplify the code a bit.

This also adds a test that checks the behaviour.

### DIFF
--- a/src/cleanfigure.m
+++ b/src/cleanfigure.m
@@ -156,32 +156,28 @@ function pruneOutsideBox(meta, handle)
         return;
     end
 
-    %hasLines = ~strcmp(lineStyle,'none') && lineWidth>0.0;
-    %hasMarkers = ~strcmp(marker,'none');
-    hasLines = true;
-    hasMarkers = true;
+    % Check if the plot has lines
+    hasLines    = ~strcmp(get(handle, 'lineStyle'),'none')...
+                && get(handle, 'lineWidth') > 0.0;
 
     % Extract the visual limits from the current line handle.
-    [xLim, yLim]   = getVisualLimits(meta);
+    [xLim, yLim]= getVisualLimits(meta);
 
     tol = 1.0e-10;
     relaxedXLim = xLim + [-tol, tol];
     relaxedYLim = yLim + [-tol, tol];
 
-    numPoints = size(data, 1);
-
     % Get which points are inside a (slightly larger) box.
     dataIsInBox = isInBox(data, relaxedXLim, relaxedYLim);
 
-    % By default, don't plot any points.
-    shouldPlot = false(numPoints, 1);
-    if hasMarkers
-        shouldPlot = shouldPlot | dataIsInBox;
-    end
+    % Plot all the points inside the box
+    shouldPlot = dataIsInBox;
+    
     if hasLines
-        % Check if the connecting line is in the box.
+        % Check if the connecting line between two data points is visible.
         segvis = segmentVisible(data, dataIsInBox, xLim, yLim);
-        % Plot points which are next to an edge which is in the box.
+        
+        % Plot points which part of an visible line segment.
         shouldPlot = shouldPlot | [false; segvis] | [segvis; false];
     end
 
@@ -189,11 +185,8 @@ function pruneOutsideBox(meta, handle)
     id_replace = [];
     id_remove  = [];
     if ~all(shouldPlot)
-        % There are two options here:
-        %      data = data(shouldPlot, :);
-        % i.e., simply removing the data that isn't supposed to be plotted.
-        % For line plots, this has the disadvantage that the line between two
-        % 'loose' ends may now appear in the figure.
+        % For line plots, simply removing the data has the disadvantage 
+        % that the line between two 'loose' ends may now appear in the figure.
         % To avoid this, add a row of NaNs wherever a block of actual data is
         % removed.
 
--- a/src/cleanfigure.m
+++ b/src/cleanfigure.m
@@ -157,8 +157,8 @@ function pruneOutsideBox(meta, handle)
     end
 
     % Check if the plot has lines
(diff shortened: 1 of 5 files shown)


miscco:
So in the octave hash file the rlocusPlot hash was missing, which made the build fail. i have added it

src/cleanfigure.m
```diff
@@ -156,32 +156,28 @@ function pruneOutsideBox(meta, handle)
         return;
     end
 
-    %hasLines = ~strcmp(lineStyle,'none') && lineWidth>0.0;
-    %hasMarkers = ~strcmp(marker,'none');
-    hasLines = true;
-    hasMarkers = true;
+    % Check if the plot has lines
+    hasLines    = ~strcmp(get(handle, 'lineStyle'),'none')...
```
egeerardyn: Please use `CamelCase` for the keywords `LineStyle` and `LineWidth` (as per the MATLAB convention).

test/suites/ACID.Octave.3.8.0.md5
```diff
@@ -56,6 +56,8 @@ quiveroverlap : a5c76200c93b83f2480aca8563626bf9
 quiverplot : a279c46f4258aeab6f7b0ad599958005
 randomWithLines : 8095aa52dd58fae39a8d5718eebe55d5
 rectanglePlot : f63538597348033398ef9cc37dcf57cd
+removeOutsideMarker :
```
egeerardyn: These hashes don't make sense. Probably, those tests crashed during execution!

egeerardyn:
Please have a look at http://pixiebox.synology.me:8090/job/matlab2tikz/331/Runner=octave%20--eval/parsed_console/ to see what is happening in Octave. There are still some bugs left.

miscco:
I think the solution is rather simple:

```
wc -l ACID.MATLAB.8.3.md5 
99 ACID.MATLAB.8.3.md5

wc -l ACID.MATLAB.8.4.md5 
99 ACID.MATLAB.8.4.md5 

wc -l ACID.Octave.3.8.0.md5 
83 ACID.Octave.3.8.0.md5
```

So i put the test at the end of the suite and Octave doesnt have all the hashes. 

There are 2 simple solutions:
1. Put the test further at the beginning
2. Add hashes for the missing tests

Which one do you prefer?

test/suites/ACID.Octave.3.8.0.md5
miscco: They dont, you just get an comparison that tells you expected( )

src/cleanfigure.m
miscco: Ah yeah, the previous code had them as parameters.

egeerardyn:
It's quite normal that Octave has less hashes than MATLAB. In Octave, quite a lot of tests are skipped and the hash files don't store empty hashes (what would be the point in storing empty things for tests that don't get executed anyway).

If there is no stored hash in the hash table, we return `''` as you can see in `execute_hash_stage`, so there is no need to store empty hashes and I think that is the problem in your commits.

In general, I would advise against editing those hash files manually.
What I tend to do is:

```matlab
status = testHeadless();
makeTravisReport(status);
% inspect the results, if everything but the new test is fine, continue:
saveHashTable(status);
```

The files can be edited by hand, but you just have to be careful because the parser is quite brittle (as you see here).

miscco:
Yeah, currently it breaks immediately after the first test. So i am rather unsure what to do here.

egeerardyn:
You should be able to fix it by removing the lines with empty hashes from the files.

miscco:
@egeerardyn Do you have the function to check the plot with the markers from #815?

egeerardyn:
I just manipulated that by hand: plot it with both, copy the data in the same TikZ file and change one of the definitions.

miscco:
So for my case MATLAB2015b there is a huge difference in the outcome:

[ACID11.zip](https://github.com/matlab2tikz/matlab2tikz/files/52360/ACID11.zip)

However, all the points that are removed additionally are outside of the box.

acid11_devel
acid11_new

I checked the outcome by saving the individual images with the following code:

```matlab
ACID_devel = imread('ACID11_devel.png');
ACID_new   = imread('ACID11_new.png');

find (ACID_Devel-ACID_new);
```

That should theoretically find any pixel difference in the images

miscco:
So all the data points in the visible part of the image are still there. For the versions i can test, only external data points were removed.

So i updated the tests. Can somebody check 2014a/b?

miscco:
Ping

miscco:
So what is really strange, is that I run on your server in 2014a/b both develop and this PR. And testHeadless generated identical tex files!?

egeerardyn:
@miscco That is strange. Because that server is what the Jenkins install runs on and there I haven't noticed those files being the same.

Anyhow, I ran a little script to diff those
```zsh
#!/usr/bin/zsh
cd ~/matlab2tikz/test/output
DIFFER=cwdiff #or diff, wdiff
for f in `ls Markers2014a/**/*.tex`; do $DIFFER $f ${f/2014a/2014b}; done
```
and I do get quite some changes between both.

(note that the script will not work in bash, which is the default shell I set up for you, but you can easily use `zsh` instead).

miscco:
So i found my mistake, i hav > in the script instead of >>, which overwrote the file rather than append the output.

Here are the results:


[diff_Markers2014a.tex.zip](https://github.com/matlab2tikz/matlab2tikz/files/89175/diff_Markers2014a.tex.zip)
There are 2 tests that differ: ACID 11 and ACID 8.

For ACID 11 this is exactly what we want. Remove markers outside of the axes. For 8 it does basically the same, although that test is always really strange

miscco:
Ah i forgot, i cannot see any reason not to merge :smiling_imp:

PR #826, opened 2015-12-07
Travis GCE migration

egeerardyn:
Travis has [migrated](https://blog.travis-ci.com/2015-11-27-moving-to-a-more-elastic-future) from OpenVZ to GCE, so apparently, `gdb` is no longer available.

This PR is to fix this on our end, if we can.


egeerardyn:
Grm, `master` vs `develop`.

PR #827, opened 2015-12-07
Travis GCE migration

egeerardyn:
Travis has [migrated](https://blog.travis-ci.com/2015-11-27-moving-to-a-more-elastic-future) from OpenVZ to GCE, so apparently, `gdb` is no longer available.

This PR is to fix this on our end, if we can (this time against `develop` instead of `master`).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,9 @@
 language: c++
 # command to install dependencies
 before_install:
-  - #sudo apt-get install -y gdb # to capture backtrace of eventual failures
   - sudo add-apt-repository -y ppa:octave/stable
   - sudo apt-get update -qq
+  - sudo apt-get install -y gdb # to capture backtrace of eventual failures
   - sudo apt-get install octave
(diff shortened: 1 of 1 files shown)


miscco:
So there is no additional installation instead of gdb?

egeerardyn:
I think @okomarov added `gdb` for debugging purposes. I just quickly removed it to check whether this fixes everything (and it seems to do so).
We can further investigate this to make sure that we can retain the functionality.

egeerardyn:
Okay, I think the situation is fixed right now: `gdb` is still installed, but, the actual problem was that we used to `install gdb` before `update`d apt. In the new infrastructure, it seems the cache is empty and hence, `gdb` wasn't fount. The fix is easy: first update the cache, then install everything we need.

While I was at it, I also cleaned up the syntax a bit. It is not necessary to pass `-y` to `apt-get` commands, since Travis is configured in such a way that this flag is implied. Otherwise, there were some indentation inconsistencies, these have been fixed as well.

PeterPablo:
:+1:

Issue #569, opened 2015-03-11
Respect grid settings of axis


PeterPablo:
@egeerardyn spotted that grid options are currently [not handled](https://github.com/matlab2tikz/matlab2tikz/pull/567/files#diff-76782b0b358b916574ab8ad22a7ce6fbR889). It should be straight forward to implement. As such I add this issue to track this feature request and so we do not forget about it.

miscco:
Hm this looks interesting, gonna have a look at it next week.

miscco:
This should mostly be covered in #828

Issue #441, opened 2014-11-11
Factory workarounds


egeerardyn:
Since #377, we have noticed that some Factory settings in MATLAB do not map well to the (effective) standard settings for some objects.
This is not a major bug on our side, we would just generate code that is too verbose and wouldn't use TikZ defaults in those cases (just a minor annoyance).
We have to work-around this issue, in particular for the following values:
- R2014a (and older?): 
  - `FactoryAxesXColor` (and probably the Y and Z variants as well) are set to white, while the effective value is black.
- R2014b (and newer?):
  - `FactoryLegendEdgeColor` has value 20% gray, but the effective value is 15% gray

In short, we have to hard-code the effective values somewhere in the code. Unfortunately, those are dependent on the MATLAB version.

PeterPablo:
Just a short remark here: In my opinion it is not necessary to explicitly store some default values, since we can always compare the property (e.g., color of legend's edge) to its factory value `FactoryLegendEdgeColor`. If they are identical, we omit it. If they differ, we proceed as usual.

egeerardyn:
@PeterPablo I agree. But we have to take care of that edge case we discovered in #641: MATLAB sets `DefaultAxesColorOrder` by itself as well in HG1, so we cannot rely on `FactoryAxesColorOrder` there.

PeterPablo:
Yes. I'd say, we have a closer look at this, once #641 is merged.

CattusNiger:
Maybe what I'm going to say is a huge piece of shit (that's the good thing of having no idea, I can yell a lot of crap and nothing happens), but if in the end we all want to have the same output in tikz, no matter what Matlab or Octave version we have, wouldn't it be better to compare the properties of the plot with the default properties in the tikz library instead of the default in Matlab/Octave?

I don't know if this is possible/realistic/easy, but it was the first thing that come to my head once I reach this problem.

Hope this could help, or at least don't offend.

PeterPablo:
Thank you for sharing your thoughts. Unfortunately matlab and octave do
behave differently. We thus need to check against their default properties
and only if there is a discrepancy pass those properties on to pgfplots.

egeerardyn:
[...]

PR #828, opened 2015-12-07
Respect grids in m2t

miscco:
Hi, this is a run for #569.

I added checks for major and minor line styles, which seems to work fine. I haven't yet looked at other gird line options, however this should be straight forward.

The difference can be best seen in ACID(16).


src/matlab2tikz.m
```diff
@@ -892,26 +892,52 @@ case guitypes()
...
-        % Take over the grid line style in any case when in strict mode.
-        % If not, don't add anything in case of default line grid line style
-        % and effectively take Pgfplots' default.
-        defaultMatlabGridLineStyle = ':';
-        if m2t.cmdOpts.Results.strict ...
-                || ~strcmpi(matlabGridLineStyle,defaultMatlabGridLineStyle)
-            gls = translateLineStyle(matlabGridLineStyle);
+    % Draws the grids of an axis
+    % Check for major grids
+    hasGrid =  strcmpi(get(handle, 'XGrid'), 'on') ...
```
egeerardyn: Please use `isOn(...)` instead of `strcmpi(...,'on')`.

src/matlab2tikz.m
```diff
@@ -892,26 +892,52 @@ case guitypes()
...
+            || strcmpi(get(handle, 'YGrid'), 'on') ...
+            || strcmpi(get(handle, 'ZGrid'), 'on');
+
+    % Check for minor grids
+    hasMinorGrid =  strcmpi(get(handle, 'XMinorGrid'), 'on') ...
+                 || strcmpi(get(handle, 'YMinorGrid'), 'on') ...
+                 || strcmpi(get(handle, 'ZMinorGrid'), 'on');
+
+    % Default matlab grid style ommited for brevity
+    % defaultGridLineStyle = ':';
```
egeerardyn: I would leave `defaultGridLineStyle` as variable. You can define it a bit further and directly do the compare:
```matlab
defaultGridLineStyle = ':';
hasLineStyle = strcmpi(gridLineStyle, defaultGridLineStyle);
```

src/matlab2tikz.m
```diff
@@ -892,26 +892,52 @@ case guitypes()
...
+
+    % Check for the grid line styles
+    if hasGrid
+        % Get the line style
+        GridLineStyle = get(handle, 'GridLineStyle');
```
egeerardyn: Nitpicking mode: please use `gridLineSyle` instead of `GridLineStyle` (I prefer variables starting with lower case, akin to Java convention).

miscco:
As i am already on it, what other options should be included.  From what i find in the handle it seems like the following would be interesting:
```matlab
         GridAlpha: 0.1500
         GridAlphaMode: 'auto'
         GridColor: [0.1500 0.1500 0.1500]
         GridColorMode: 'auto'
         GridLineStyle: '-'
         MinorGridAlpha: 0.2500
         MinorGridAlphaMode: 'auto'
         MinorGridColor: [0.1000 0.1000 0.1000]
         MinorGridColorMode: 'auto'
         MinorGridLineStyle: ':'
   ```

So is alpha relevant? I think color should be in there

File: src/matlab2tikz.m [hunk 4]
```diff
@@ -892,26 +892,52 @@ case guitypes()
...
+            || isOn(get(handle, 'YGrid')) ...
+            || isOn(get(handle, 'ZGrid'));
+
+    % Check for minor grids
+    hasMinorGrid =  isOn(get(handle, 'XMinorGrid')) ...
+                 || isOn(get(handle, 'YMinorGrid')) ...
+                 || isOn(get(handle, 'ZMinorGrid'));
+
+    % Default matlab grid style ommited for brevity
+    % defaultGridLineStyle = ':';
```
miscco: So i commented it out, as the length of defaultGridLineStyle is rather lengthy in comparison with ':'

src/matlab2tikz.m
```diff
@@ -892,26 +892,52 @@ case guitypes()
...
-                m2t.axesContainers{end}.options,...
-                axisGridOpts);
+                m2t.axesContainers{end}.options, axisGridOpts);
         end
-    else
+    end
+
+    if hasMinorGrid
+        % Get the line style
+        MinorGridLineStyle = get(handle, 'MinorGridLineStyle');
```
egeerardyn: Same remark about starting with a capital letter: please don't. Also, in this case, I would say that:
 - the comments don't add anything (comments should explain why you do something in a particular way or explain the context of the code, not just tell what you do)
 - in this case, you can just as well do `glsMinor = translateLineStyle(get(handle,'MinorGridLineStyle'))`.

src/matlab2tikz.m
egeerardyn: Oh, now I come to think of it: `:` is only the default in HG1, we probably need to check the defaults to make this code work well in both HG1 and HG2.

src/matlab2tikz.m
miscco: Already done ;) seems it didnt refresh on your side

src/matlab2tikz.m
miscco: Done

src/matlab2tikz.m
miscco: So from my point of view, is there a good reason, why we need to check it? It doesnt really hurt to have the line grid style = {}
 and it kind of adds complexity

miscco:
So currently we check for girds in getAxisOptions

However, this adds xmajorgrid and friends for all axes individually. 

I wonder whether we keep it that way or add grid = mayor/minor directly in the drawGridOfAxes function

egeerardyn:
@miscco I would say the `...Alpha` values are relevant.

Personally, I prefer the approach where we add `xmajorgrid` and so on per axis, instead of `grid = major`. This keeps the code simpler on our end and the approach is more flexible, e.g. you can have minor grids in the X direction, major+minor in Y and no grids in Z. The notation `grid=both` is nice when writing TikZ code by hand, but it mainly adds corner cases to our codebase while the current approach works just fine.

That being said: if you really prefer the short notation, I would suggest you leave the implementation as is (i.e. produce `xmajorgrid` etc.) and write a function to perform the simplifications when they are appropriate (i.e. for a 2D plot with `xmajorgrid` and `ymajorgrid` set, replace those options with `grid=major`).

miscco:
@egeerardyn True enough, i thought about a check if all grids are on, but i think that is just an overkill.

src/matlab2tikz.m
egeerardyn: You should be able to get the same result with:

```matlab
[gridLineStyle, isDefault] = getAndCheckDefault('axes', handle, 'GridLineStyle', ':');
```

which should work for both HG1 and HG2. 

What that function does, is it first checks the MATLAB factory defaults for the relevant setting. If that setting is not found, it uses the `':'` as fallback. It retrieves the property from your object and `isDefault` will be true when the default value and the actual one are equal.

As far as I'm concerned, you can just pass in `':'` as a literal to the function, without giving it a name, just make sure you remove the commented-out code: it can only become outdated and then we have to dig around to see what to do with it.

miscco:
So for alpha and color i would intend the following:

1. Check for alpha =0.15 (default value)
2. Check whether GridColor = alpha * AxesColor

if either of them is untrue, add an appropriate grid option

egeerardyn:
@miscco: For the grid color, I just think you should check for equality to the default color. The alpha value shouldn't influence that process.

miscco:
Hm the axis color in the standard case are [1, 1, 1] (black) and the XGridColor is [0.15, 0.15, 0.15] which fits nicely with the default alpha value of 0.15.

egeerardyn:
I suspect that's a coincidence. You should check the default grid colors, not the default axes color (= color of the box if I'm correct)

E.g. on R2015b:
```matlab
>> get(groot,'DefaultAxesGridColor')
ans =
         0.15         0.15         0.15
>> get(groot,'FactoryAxesGridColor')
ans =
         0.15         0.15         0.15
>> get(groot,'FactoryAxesColor')
ans =
     1     1     1
```

In any case, now that I think of it, I remember I once filed a bug report with TMW about inconsistent behavior of these defaults (see also #441).

miscco:
So i went ahead and tried a more future proof approach using an opts structure. If this finds some support, i can easily add addtional options.

src/matlab2tikz.m
```diff
@@ -892,37 +892,48 @@ case guitypes()
...
+    % Draws the grids of an axis
+    gridOpts = opts_new();
+
+    % Check for major/minor grids
+    hasGrid =  [isOn(get(handle, 'XGrid'));
+                isOn(get(handle, 'YGrid'));
+                isOn(get(handle, 'ZGrid'))];
+
+    hasMinorGrid =  [isOn(get(handle, 'XMinorGrid'));
+                 	 isOn(get(handle, 'YMinorGrid'));
```
egeerardyn: Indentation is off here. Probably due to mixing tabs and spaces.

src/matlab2tikz.m
```diff
@@ -892,37 +892,48 @@ case guitypes()
...
+                 	 isOn(get(handle, 'ZMinorGrid'))];
+
+    % Check for the grid options
+    if any(hasGrid)
+        % Get the line style and translate it to pgfplots
+        gridLS = translateLineStyle(get(handle, 'GridLineStyle'));
```
egeerardyn: I would prefer to keep having the check against the default value (and also the strict switch) here, e.g.
```matlab
[gridLS, isDefault] = getAndCheckDefault('axes', handle, 'GridLineStyle', ':');
if ~isDefault || m2t.Results.strict
    gridOpts = opts_add(gridOpts, ...);
end
```
 (note: code is untested).

That way, we keep our TikZ code leaner and we translate ugly MATLAB settings to beautiful TikZ ones (in HG2, I admit the situation has improved a lot, but in HG1 the dotted grid lines are a menace).

src/matlab2tikz.m
okomarov: use the `formatWhitespace` function :)

egeerardyn:
I like the general trend of it. I love what you did with the `isOn` calls together with `any`, but I still have two minor comments with respect to some minor details in the commits:
 - some indentation is off
 - I feel we should keep checking the defaults and only include the statements when they deviate from the defaults

miscco:
So what i am unsure about is whether the defaults match with pgfplots. There is no guarantee, that the default linestyle in Matlab is the sae as in pgfplots

egeerardyn:
@miscco I agree, the defaults don't necessarily match between `pgfplots` and MATLAB. But that's actually what I like about it. 

IMHO, that's what the `strict` mode is for: to reproduce a figure as closely as possible, even if we end up reproducing all the ugly things from the MATLAB figure.

In non-strict mode (i.e. normal mode), I think we should match the intent. I.e. if a user doesn't change a setting, he probably doesn't care about it and we can just as well use the pgfplots defaults since those are generally better than MATLAB's (dotted grid lines in HG1, just off-black axes lines in HG2, ...). That mode also produces leaner TikZ code, such that a user can more easily change these graphs with `every ...` style options in the document.

I can agree that the latter is a rather vague concept, but I have the impression people like it that way (I certainly do). The downside is a bit more complex code, but it remains manageable.

miscco:
So this is nearly there, just some small nitpicks to discuss:

1. Minor Grid defaults. If i do check for the default value for minor grids, pgfplots will plot a solid line instead of a dotted one. Here i would say, that we should omit the check, as the user has to manually edit the file

2. ColorMode. I implemented global color options via GridColor and MinorGridColor. However, as the Matlab manual said, the individual axes can have different colors if {X, Y, Z}Color is set. I dont really know how to implement that, and even if it is worth it.

src/matlab2tikz.m
```diff
@@ -892,37 +892,111 @@ case guitypes()
...
-    else
+        if hasMinorGrid(i)
+            grid     = [xyz{i}, 'minorgrids'];
+            options = opts_add(options, grid, []);
+        end
+    end
+
+    % Check for global grid options
+    if any(hasGrid)
+        gridOpts = [];
```
egeerardyn: Please use `opts_new()` here. Most of our code uses `opts_*` for handling those option arrays.

src/matlab2tikz.m
```diff
@@ -892,37 +892,111 @@ case guitypes()
...
+        % Get the line style and translate it to pgfplots
+        [gridLS, isDefault] = getAndCheckDefault(...
+            'axes', handle, 'GridLineStyle', ':');
+        if ~isDefault || m2t.cmdOpts.Results.strict
+            gridOpts = translateLineStyle(gridLS);
```
egeerardyn:
```
gridOpts = opts_add(gridOpts, gridLS);
```

src/matlab2tikz.m
miscco: This is not an options array. The problem is, that these global options *all* end up within a 
```
grid style = { }
```
Therefore, i gather the options in a string beforehand

src/matlab2tikz.m
```diff
@@ -892,37 +892,111 @@ case guitypes()
...
+        end
+
+        % Get the color of the grid and translate it to pgfplots usable
+        % values
+        [gridColor, noColor] = getAndCheckDefault(...
+            'axes', handle, 'GridColor', [0.15, 0.15, 0.15]);
+        if ~noColor
+            [m2t, gridColor] = getColor(m2t, handle, gridColor, 'patch');
+            gridOpts = [gridOpts, ', ', gridColor];
```
egeerardyn:
```
gridOpts = opts_add(gridOpts, 'color', gridColor);
```

since I suspect the actual property name is either `color` (but it might be `draw`, since I typically confuse those).

src/matlab2tikz.m
```diff
@@ -892,37 +892,111 @@ case guitypes()
...
+        end
+
+        % Get the alpha of the grid and translate it to pgfplots
+        % NOTE: The color conversion does not take alpha values into
+        % account. Therefore, we have to add it explicitely when using
+        % colored grids
+        [gridAlpha, noAlpha] = getAndCheckDefault(...
+            'axes', handle, 'GridAlpha', 0.1);
+        if ~noAlpha || ~noColor
+            gridOpts = [gridOpts, ', opacity =', num2str(gridAlpha)];
```
egeerardyn:
```
gridOpts = opts_add(gridOpts, 'opacity', num2str(gridAlpha));
```

src/matlab2tikz.m
```diff
@@ -892,37 +892,111 @@ case guitypes()
...
+        end
+
+        if ~isempty(gridOpts)
+            options = opts_add(options, 'grid style', sprintf('{%s}', gridOpts));
```
egeerardyn:
```
gridOptions = opts_print(m2t, gridOpts, ',');
options = opts_add(options, 'grid style', ['{' gridOptions '}']);
```

src/matlab2tikz.m
```diff
@@ -892,37 +892,111 @@ case guitypes()
...
+            'axes', handle, 'MinorGridLineStyle', ':');
+        minorGridOpts = translateLineStyle(minorGridLS);
+
+        %Get the color of the grid and translate it to pgfplots usable
+        % values
+        [minorGridColor, noColor] = getAndCheckDefault(...
+            'axes', handle, 'MinorGridColor', [0.1, 0.1, 0.1]);
+        if ~noColor
+            [m2t, minorGridColor] = getColor(m2t, handle, minorGridColor, 'patch');
+            minorGridOpts = [minorGridOpts, ', ', minorGridColor];
```
egeerardyn: Same remark as for the major grid and usage of `opts_*`

src/matlab2tikz.m
```diff
@@ -892,37 +892,111 @@ case guitypes()
...
+        end
+
+        % Get the alpha of the grid and translate it to pgfplots
+        % NOTE: The color conversion does not take alpha values into
+        % account. Therefore, we have to add it explicitely when using
+        % colored grids
+        [minorGridAlpha, noAlpha] = getAndCheckDefault(...
+            'axes', handle, 'MinorGridAlpha', 0.1);
+        if ~noAlpha || ~noColor
+            minorGridOpts = [minorGridOpts, ', opacity =', num2str(minorGridAlpha)];
```
egeerardyn: Same remark as for the major grid and usage of `opts_*`

src/matlab2tikz.m
```diff
@@ -892,37 +892,111 @@ case guitypes()
...
+        end
+
+        if ~isempty(minorGridOpts)
+            options = opts_add(options, 'minor grid style', sprintf('{%s}', minorGridOpts));
```
egeerardyn: As for the major grid: remember to print the OptionArray `minorGridOpts` to string before passing this into the main `options`.

miscco:
@egeerardyn Thanks a lot, the opts array made it way better

egeerardyn:
@miscco That was fast :smile:. I hope it works, because I didn't open MATLAB to write the code.

miscco:
Some minor adjustments, but in general it worked. I omited all the empty braces ', []' at the end of opts_add is that suitable?

egeerardyn:
@miscco Great! (I ~~hate~~ utterly dislike those empty brackets)

miscco:
Maybe a bad habit, but i went ahead and cleared the other 10 or so occurences of opts_add(*, [])

egeerardyn:
I wouldn't call that a bad habit: as long as it's a commit of it's own, that is perfectly fine.

With respect to your two points above:

> Minor Grid defaults. If i do check for the default value for minor grids, pgfplots will plot a solid line instead of a dotted one. Here i would say, that we should omit the check, as the user has to manually edit the file

I would keep the check for the same reason as with major grid lines. This can be overridden by the user either using the `strict` parameter for m2t or by specifying the `every minor grid` option in pgfplots (e.g. in the document preamble):

screen shot 2015-12-08 at 10 37 52

> ColorMode. I implemented global color options via GridColor and MinorGridColor. However, as the Matlab manual said, the individual axes can have different colors if {X, Y, Z}Color is set. I dont really know how to implement that, and even if it is worth it.

I'm quite surprised, but it is possible according to the pgfplots manual:
screen shot 2015-12-08 at 10 39 10

That could make the code abstracted from x/y/z dimensions :smile: 
Anyhow, I think it's more important to first make sure the test suite is green again.

If you want to tackle X/Y/Z dependence in this PR, feel free to do so. Otherwise, it's equally fine to open a new issue/feature request for that (please do link back to this PR then).

miscco:
Hm, so in principle, it would make the code rather clumsy, as one had to check for global options first then for overwrite of local options.

Meh.

As far as i can tell, the test suite is rather happy with this. A problem might be, that {x, y, z}majorgrid and friends moved a little down the list, as i put the option into drawGrid rather than axisOptions.

here is the pdf on my installation

http://workupload.com/file/MOAnEOT8

Mind i skipped 42 as it breaks compilation of the acid.pdf all the time, dunno why, as the file itself compiles after editing

egeerardyn:
@miscco There is no need to tinker with the global options. Except for the `every ...` variants, all those can be specified in the axis. So adding a `major x grid style` instead of `major grid style` (as in this PR), will only affect the X direction and so on.

I have noticed that in HG1, there is no `GridColor` property and the color seems to be determined completely from the `X/Y/ZColor`. So perhaps it's the easiest to translate the X/Y/Z-dependent stuff another time.

miscco:
So i tested it and it doesnt really work as intended. the whole grid gets plastered in the XColor. I believe, this is kind of an advanced option that one should leave to the user to manually edit.

egeerardyn:
Let's deal with those advanced things another time.
I did a quick trial: converting a simple plot from m2t and adding the relevant option myself:
```latex
%
%The latest updates can be retrieved from
%  http://www.mathworks.com/matlabcentral/fileexchange/22022-matlab2tikz-matlab2tikz
%where you can also make suggestions and rate matlab2tikz.
%
\usepackage[T1]{fontenc}
\usepackage[utf8]{inputenc}
\usepackage{pgfplots}
\usepackage{grffile}
\pgfplotsset{compat=newest}
\usetikzlibrary{plotmarks}
\usetikzlibrary{arrows.meta}
\usepackage{amsmath}

\begin{document}
\begin{tikzpicture}

\begin{axis}[%
width=6.021in,
height=4.749in,
at={(1.01in,0.641in)},
scale only axis,
separate axis lines,
every outer x axis line/.append style={black},
every x tick label/.append style={font=\color{black}},
xmin=1,
xmax=10,
xmajorgrids,
every outer y axis line/.append style={black},
every y tick label/.append style={font=\color{black}},
major x grid style={color=green},
ymin=1,
ymax=10,
ymajorgrids,
]
\addplot [color=blue,solid,forget plot]
  table[row sep=crcr]{%
1	1\\
2	2\\
3	3\\
4	4\\
5	5\\
6	6\\
7	7\\
8	8\\
9	9\\
10	10\\
};
\end{axis}
\end{tikzpicture}%
\end{document}
```

test

It doesn't matter whether I write `color=green` or `draw=green` (which seems odd to me).

miscco:
@egeerardyn Can you run a testGraphical for the early matlab releases?

From my side on 2015b it seems fine.

egeerardyn:
@miscco You can already inspect a part on http://pixiebox.synology.me:8090/job/matlab2tikz/Runner=matlab2014a%20-r/360/parsed_console/ 
Due to the lack of `GridColor` property in HG1, `ACID(16)` seems to fail in the plot stage. You can probably fix this using:
```matlab
if isprop(gca,'GridColor')
    set(gca, 'GridColor', ...);
else
    %TODO equivalent HG1 settings (if those exist)
end
```
in that plot code.

miscco:
Should we also check in m2t for HG1?

egeerardyn:
@miscco What do you mean? If we should implement the grid conversion for HG1 as well? Then yes, but not necessarily in this PR (since in HG2, we also can extend the functionality towards X/Y/Z dependence).

In m2t, I don't think we need explicit HG1 checks thanks to the default values passed to `getAndCheckDefault`. We might want to double-check those values (I can do so tonight) and document it in the code to remind us in the future that HG1 uses the fallback path.

miscco:
Ah true, the default we set helps i think. 

I have checked in octave, and there the situtation is similar to HG1. Only {X, y, Z}Color and no grid specific options rather than GridLineStyle

miscco:
So here are the diffs for Octave4.0:
```
ACID5:
30d29
< xmajorgrids,
33,34c32,34
< ymajorgrids,
< axis background/.style={fill=white}
---
> axis background/.style={fill=white},
> xmajorgrids,
> ymajorgrids

ACID16
28,29d27
< xmajorgrids,
< xminorgrids,
33a32,34
> axis background/.style={fill=white},
> xmajorgrids,
> xminorgrids,
36c37
< axis background/.style={fill=white}
---
> minor grid style={dotted}

ACID24:
28d27
< xmajorgrids,
31d29
< ymajorgrids,
34d31
< zmajorgrids,
39c36,39
< axis z line*=left
---
> axis z line*=left,
> xmajorgrids,
> ymajorgrids,
> zmajorgrids

ACID55:
26d25
< xmajorgrids,
29d27
< ymajorgrids,
32d29
< zmajorgrids,
37c34,37
< axis z line*=left
---
> axis z line*=left,
> xmajorgrids,
> ymajorgrids,
> zmajorgrids

ACID56:
27d26
< xmajorgrids,
31d29
< ymajorgrids,
35d32
< zmajorgrids,
41c38,41
< axis z line*=left
---
> axis z line*=left,
> xmajorgrids,
> ymajorgrids,
> zmajorgrids

ACID57:
28d27
< xmajorgrids,
32d30
< ymajorgrids,
36d33
< zmajorgrids,
41a39,41
> xmajorgrids,
> ymajorgrids,
> zmajorgrids,

ACID60:
26d25
< xmajorgrids,
30d28
< ymajorgrids,
34d31
< zmajorgrids,
39c36,39
< axis z line*=left
---
> axis z line*=left,
> xmajorgrids,
> ymajorgrids,
> zmajorgrids

ACID83:
29d28
< xmajorgrids,
32,33c31,33
< ymajorgrids,
< axis background/.style={fill=white}
---
> axis background/.style={fill=white},
> xmajorgrids,
> ymajorgrids
```

So as far as i can see, the PR only shifted the {x, y, z}majorgrids and friends to the back of the axis options.

miscco:
The same for MATLAB2015b:
>!
```
ACID5:
32d31
< xmajorgrids,
37d35
< ymajorgrids,
40c38,40
< axis y line*=left
---
> axis y line*=left,
> xmajorgrids,
> ymajorgrids

ACID16
31,32d30
< xmajorgrids,
< xminorgrids,
38a37,39
> axis background/.style={fill=white},
> xmajorgrids,
> xminorgrids,
41c42,43
< axis background/.style={fill=white}
---
> grid style={red,opacity=0.15},
> minor grid style={dotted,blue,opacity=0.25}

ACID27:
28d27
< xmajorgrids,
34,35c33,35
< ymajorgrids,
< axis background/.style={fill=white}
---
> axis background/.style={fill=white},
> xmajorgrids,
> ymajorgrids

ACID55:
28d27
< xmajorgrids,
33d31
< ymajorgrids,
38d35
< zmajorgrids,
43c40,43
< axis z line*=left
---
> axis z line*=left,
> xmajorgrids,
> ymajorgrids,
> zmajorgrids

ACID56:
29d28
< xmajorgrids,
35d33
< ymajorgrids,
41d38
< zmajorgrids,
48c45,48
< axis z line*=left
---
> axis z line*=left,
> xmajorgrids,
> ymajorgrids,
> zmajorgrids

ACID57:
30d29
< xmajorgrids,
36d34
< ymajorgrids,
42d39
< zmajorgrids,
47a45,47
> xmajorgrids,
> ymajorgrids,
> zmajorgrids,

ACID59:
29d28
< xmajorgrids,
35d33
< ymajorgrids,
40d37
< zmajorgrids,
47c44,47
< axis z line*=left
---
> axis z line*=left,
> xmajorgrids,
> ymajorgrids,
> zmajorgrids

ACID60:
28d27
< xmajorgrids,
34d32
< ymajorgrids,
40d37
< zmajorgrids,
45c42,45
< axis z line*=left
---
> axis z line*=left,
> xmajorgrids,
> ymajorgrids,
> zmajorgrids

ACID62:
29d28
< xmajorgrids,
35d33
< ymajorgrids,
40a39,40
> xmajorgrids,
> ymajorgrids,

ACID75:
30d29
< xmajorgrids,
36d34
< ymajorgrids,
41d38
< zmajorgrids,
46c43,46
< axis z line*=left
---
> axis z line*=left,
> xmajorgrids,
> ymajorgrids,
> zmajorgrids

ACID82:
32d31
< xmajorgrids,
37d35
< ymajorgrids,
42d39
< zmajorgrids,
47c44,47
< axis z line*=left
---
> axis z line*=left,
> xmajorgrids,
> ymajorgrids,
> zmajorgrids

ACID83:
31d30
< xmajorgrids,
36d34
< ymajorgrids,
39c37,39
< axis y line*=left
---
> axis y line*=left,
> xmajorgrids,
> ymajorgrids
```

miscco:
Any way to hide the text like a spoiler?

egeerardyn:
Not that I know. The other option is to upload it to a `gist` (or pastebin or whatever) and post a link.

miscco:
So i changed it a bit, to not explicitly add opacity. This would have lead to more matlabish colors but the opacity was strange with minor grids.

this should be ready to merge now.

miscco:
So given this is a new feature, is there anything else we want to have inside?

PR #829, opened 2015-12-07
Minor cleanup

egeerardyn:
Just some minor cleanup:
 - removed an unused function
 - fixed spacing in a function (it seems our automatic whitespace tool is a bit too coarse at the moment), I will have a look whether we can't make `git` do some things in a pre-commit hook to avoid at least trailing whitespace.

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -5581,16 +5581,6 @@ function [val, isDefault] = getAndCheckDefault(type, handle, key, default)
     isDefault = isequal(val, default);
 end
 % ==============================================================================
-function opts = addIfNotDefault(m2t, type, handle, key, default, pgfKey, opts)
-    % sets an option in the options array named `pgfKey` if the MATLAB option is
-    % not a default value
-    % FIXME: this function is currently unused -- remove it in the future?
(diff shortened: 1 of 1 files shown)


miscco:
Seems obviously correct :+1:

egeerardyn:
Thanks for checking!

miscco:
Although i have to say, that I might use the addIfNotDefault function for the grid options. It seems rather handy, whenever one deals with coloring etc

egeerardyn:
Feel free to revert that commit when you would want to use it :smile:

In the past, I wrote that function exactly for such purposes, but now that I think of it: there are actually only a few properties that can be converted 1:1 from MATLAB to TikZ. E.g. colors require additional translation (e.g. `[1 0 0 ]` to `'red'`), such that the function has very few actual use cases.

miscco:
That is true, but it can be handy to check if it is the default value and only change it when not.

egeerardyn:
@miscco That's why we have the `getAndCheckDefault`.
The function I removed, `addIfNotDefault`, was a wrapper around `getAndCheckDefault` that just added the MATLAB value to an OptionArray if the value was non-default. So it's less useful.

Issue #831, opened 2015-12-08
Standardize function output in m2t


miscco:
So i went scanning the code and i found a lot of different ways functions wokr.

Some return a str, some write directly into the file via the fid, some  return [m2t, str]

I am sure, that this is due to historical growth. If there is some agreement, i would tackle that, to unify the way we return from functions and order the functions depending on their role.

egeerardyn:
@miscco I agree that there is currently little convention and the scope jumps a bit up and down.
This is all related to #477, #445 and #566 as well.

Your idea is very good, but let's wait a moment.
This afternoon, I have been looking into #393, for which I have already uniformized quite a lot of these handlers to
```matlab
[m2t, str] = handler(m2t, h, custom)
```
where `custom` is a newly introduced thingy.
There are a few places where this is not yet possible, most importantly:
 - `handleAllChildren` since that needs to return a `cellstr`
 - `drawHggroup` (just because it adds no value by itself)
 - `drawAxes` (i.e. second highest level)
 - `saveToFile` (i.e. highest level)

But I will make my experimental code public in a short while.

In particular, the last two need a lot of TLC to uniformize them: they build a faux object in `m2t` and certainly `saveToFile` writes the structure out to file. I think it would make sense in the mid-long run (say, 2.0 release) to make sure that those functions live by that interface and the concerns get separated in `drawAxes` and `saveToFile`, i.e.:
 - `drawAxes` should build a str as well
 - `saveToFile` should be split into:
     * `drawFigure` that builds the string representation, and
     * `saveToFile` that does the writing to disk.

The problem with migrating right away is that we use `m2t` as a kind of god object (global variable, or whatever you like to call that) as discussed in the other issue reports. This makes for subtle coupling that is not trivial to untangle. IMHO, the best moment to tackle such invasive operations, is to wait until there are no PRs open since it tends to produce a lot of merge conflicts.

PR #832, opened 2015-12-08
TikZ code customization

egeerardyn:
This is a partial implementation of #393 and as discussed in #831 as well, this encompasses a more unified calling convention of the `draw*` functions.

I am releasing this as **experimental**, i.e. the code should not be merged yet! This is just for review and discussion. In due time, I will rebase this code on top of develop.

TODO:
- [ ] also add a field for custom file names, e.g. such that we can do "curve X" should end up in a file named `x.tsv` instead of `figure1.tsv`.


egeerardyn:
Just FYI, I ran some timings on this PR as I expect a bit of a performance penalty: for every object for which we call `handleObject`, I basically build an `inputParser` to validate its custom data structure (i.e. I expect this to be quite slow, but it makes optimal use of pre-existing code and is very flexible).
screen shot 2015-12-09 at 10 37 53

From the figure, you can see that for most our test cases, there is a 10% performance penalty (which is less than I anticipated, so I'm quite happy with this). In any case, there are a few things we can try to speed this up a bit:
 - make the `inputParser` persistent
 - refactor `m2tcustom` such that it first checks in which mode it runs and only does the necessary work for each case. I suspect specifically handling the case `m2tcustom([])` may provide us with a decent speed-up already.

miscco:
Thats more than i expected, as i will be traveling a lot for the rest of the month i doubt i will find too much time reviewing this approach, given the other PRs.

egeerardyn:
Okay, I got rid of the performance penalty, I think (based on 5 repeated tests).
screen shot 2015-12-09 at 13 44 16

egeerardyn:
@LJeub I have noticed that you have been working further on this feature on your own repo. If you have some comments for this branch, I would be very happy discuss them with you and make sure we can converge back again.

PR #836, opened 2015-12-30
Tick label rotation

RPeschke:
allows to use rotated tick labels

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -1508,7 +1508,7 @@ end
 
 function options = setAxisTickLabelStyle(m2t, options,axis,handle)
  keywordTickLabelRotation = [upper(axis), 'TickLabelRotation'];
- Rotation=get(handle,keywordTickLabelRotation);
+  Rotation=getOrDefault(handle, keywordTickLabelRotation, 0);
  if Rotation ~=0
     options = opts_add(options, [axis,'ticklabel style'], sprintf('{rotate=%d}',Rotation));
  end
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -1505,13 +1505,13 @@ function [options] = getAxisTicks(m2t, handle, axis, options)
         
     options = setAxisTickLabelStyle(m2t, options,axis,handle);
 end
-
-function options = setAxisTickLabelStyle(m2t, options,axis,handle)
- keywordTickLabelRotation = [upper(axis), 'TickLabelRotation'];
-  Rotation=getOrDefault(handle, keywordTickLabelRotation, 0);
- if Rotation ~=0
-    options = opts_add(options, [axis,'ticklabel style'], sprintf('{rotate=%d}',Rotation));
- end
+% ==============================================================================
+function options = setAxisTickLabelStyle(m2t, options, axis, handle)
+    keywordTickLabelRotation = [upper(axis), 'TickLabelRotation'];
+    Rotation = getOrDefault(handle, keywordTickLabelRotation, 0);
+    if Rotation ~= 0
+        options = opts_add(options, [axis, 'ticklabel style'], sprintf('{rotate=%d}', Rotation));
+    end
 end
 % ==============================================================================
 function interpreter = defaultTickLabelInterpreter(m2t)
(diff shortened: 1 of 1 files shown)


egeerardyn:
@matlab2tikzBot: test this PR

src/matlab2tikz.m
```diff
@@ -1502,6 +1502,16 @@ case guitypes()
 
     options = setAxisTicks(m2t, options, axis, pgfTicks, pgfTickLabels, ...
         hasMinorTicks, tickDirection, isDatetimeTicks);
+        
+    options = setAxisTickLabelStyle(m2t, options,axis,handle);
+end
+
+function options = setAxisTickLabelStyle(m2t, options,axis,handle)
+ keywordTickLabelRotation = [upper(axis), 'TickLabelRotation'];
```
egeerardyn: This only exists in MATLAB R2014b and newer. In Octave and R2014a or older, this throws errors like `The name 'XTickLabelRotation' is not an accessible property for an instance of class 'axes'.`.

To avoid this, you can handle this with our `getAndCheckDefault` or `getOrDefault` functions:
```matlab
key = [upper(axis), 'TickLabelRotation'];
% either:
[rotation, isDefaultRotation] = getAndCheckDefault('axes', handle, key, 0);
if ~isDefaultRotation
    ...
end
% or:
rotation = getOrDefault(handle, key, 0);
if rotation ~= 0
    ...
end
```

egeerardyn:
I only have a few other minor comments apart from my other one:
 - very nice to extract this to a separate function, I like it
 - the indentation seems to be off (and a few commas without a space after them). Could you fix this?

Once those are addressed, I am completely in favor of incorporating this into the main version.

Note: if you want access to the test results (and have your PRs tested automatically), I have sent you an invite for our associate developers group (= no implied responsibilities, just access to the testing server).

src/matlab2tikz.m
```diff
@@ -1502,6 +1502,16 @@ case guitypes()
 
     options = setAxisTicks(m2t, options, axis, pgfTicks, pgfTickLabels, ...
         hasMinorTicks, tickDirection, isDatetimeTicks);
+        
+    options = setAxisTickLabelStyle(m2t, options,axis,handle);
+end
+% ==============================================================================
+function options = setAxisTickLabelStyle(m2t, options, axis, handle)
+    keywordTickLabelRotation = [upper(axis), 'TickLabelRotation'];
+    Rotation = getOrDefault(handle, keywordTickLabelRotation, 0);
```
miscco: I would rather have keywordTickLabelRotation removed and the corresponding assignment directly into the getOrDefault. It is only used once and doesnt bring any additional insight.

Also it would be nice to rename Rotation to tickLabelRotation or whatever suits you best.

miscco:
I had one nitpick, before commiting i would ask, whether there are other TickLabelStyles that we might consider. Given this is a nice generic encapsulated function it might be the right time to add them

src/matlab2tikz.m
egeerardyn: @miscco I disagree: introducing a variable (even if it's used only once) is a good thing because it makes the code self-documenting.
I'm also more in favor of `rotation` rather than `tickLabelRotation`: it is more terse and the scope of the function implies we are considering tick labels and not the rotation of something else (in a more generic function, however, I would totally agree with you to specify which rotation).

egeerardyn:
@miscco: I somewhat had the same nitpick in my mind but the other way around: the function name is more generic than what is actually done, but, as you said: that is an opportunity to add more. So I quite like it.

In this PR, I'm fine with just a `TODO` comment in the function to signal this and revisit it another time: tick label rotation is a specific feature, no need to add more to it if that would introduce more complicated code (I love that this PR is very simple). 

Anyhow, I think we would benefit from a more systematic approach to find which properties we do or do not translate and tackle such a TODO at that time.

miscco:
Yeah, a TODO would be fine with me.

src/matlab2tikz.m
miscco: whats irritating me is the length of keywordTickLabelRotation, for the small use it brings. maybe shorten it to keyword? nd then use rotation as variable?

egeerardyn: `keyword` sounds like a good compromise to me. Indeed, the name was somewhat on the long side. :+1:

egeerardyn:
@RPeschke Thank you for your contribution :+1:

PR #841, opened 2016-01-11
Generate automatic labels by tags

murmlgrmpf:
The autolabel option of matlab2tikz is working fine, but if you change your plots, because you plot an additional line, all labels change.
Therefore, it is convenient to add the label you want to have in tex to the Tag property of your line in matlab. When exporting with autolabels, the original labels are reproduced for compatibility and additionally the Tag will show up as a label.


miscco:
Please resubmit this PR against the develop branch rather than master.

murmlgrmpf:
Thanks for closing, I was already looking for a way to do this... sorry for the inconvenience

miscco:
No problem at all, I can't tell you how often that has happened to me in the beginning :smile:

PR #842, opened 2016-01-11
add labels to plot according to Tag property of line

murmlgrmpf:
The autolabel option of matlab2tikz is working fine, but if you change your plots, because e.g. you plot an additional line, all labels change.
Therefore, it is convenient to add the label you want to have in tex to the Tag property of your line in matlab. When exporting with autolabels, the original labels are reproduced for compatibility and additionally the Tag will show up as a label.

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -1763,20 +1763,19 @@ function [m2t, generatedCodeSoFar, labelCode] = addLabel(m2t, generatedCodeSoFar
     if ~exist('generatedCodeSoFar','var') || isempty(generatedCodeSoFar)
         generatedCodeSoFar = '';
     end
+    
     if m2t.cmdOpts.Results.automaticLabels
-        [pathstr, name] = fileparts(m2t.cmdOpts.Results.filename); %#ok
-        labelName = sprintf('addplot:%s%d', name, m2t.automaticLabelIndex);
-        labelCode = sprintf('\\label{%s}\n', labelName);
-        m2t.automaticLabelIndex = m2t.automaticLabelIndex + 1;
-
-        userWarning(m2t, 'Automatically added label ''%s'' for line plot.', labelName);
-        generatedCodeSoFar = [generatedCodeSoFar, labelCode];
-        
         if ~isempty(lineTag)
             labelName = sprintf('%s', lineTag);
-            labelCode = sprintf('\\label{%s}\n', labelName);
-            generatedCodeSoFar = [generatedCodeSoFar, labelCode];
+        else
+            [pathstr, name] = fileparts(m2t.cmdOpts.Results.filename); %#ok
+            labelName = sprintf('addplot:%s%d', name, m2t.automaticLabelIndex);
+            m2t.automaticLabelIndex = m2t.automaticLabelIndex + 1;
         end
+        labelCode = sprintf('\\label{%s}\n', labelName);
+        generatedCodeSoFar = [generatedCodeSoFar, labelCode];
+        
+        userWarning(m2t, 'Automatically added label ''%s'' for line plot.', labelName);
     end
 end
 % ==============================================================================
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -1653,7 +1653,6 @@ function [m2t, str] = drawLine(m2t, h, yDeviation)
     lineStyle            = get(h, 'LineStyle');
     lineWidth            = get(h, 'LineWidth');
     lineOptions          = getLineOptions(m2t, lineStyle, lineWidth);
-    lineTag              = get(h,'Tag');
     [m2t, markerOptions] = getMarkerOptions(m2t, h);
 
     drawOptions = opts_new();
@@ -1677,7 +1676,7 @@ function [m2t, str] = drawLine(m2t, h, yDeviation)
     m2t = jumpAtUnboundCoords(m2t, data);
 
     [m2t, str] = writePlotData(m2t, str, data, drawOptions);
-    [m2t, str] = addLabel(m2t, str, lineTag);
+    [m2t, str] = addLabel(m2t, str, h);
 end
 % ==============================================================================
 function [m2t, str] = specialDrawZplaneUnitCircle(m2t, h, drawOptions)
@@ -1758,13 +1757,14 @@ function [data] = getXYZDataFromLine(m2t, h)
     end
 end
 % ==============================================================================
-function [m2t, generatedCodeSoFar, labelCode] = addLabel(m2t, generatedCodeSoFar, lineTag)
+function [m2t, generatedCodeSoFar] = addLabel(m2t, generatedCodeSoFar, h)
     % conditionally add a LaTeX label after the current plot
     if ~exist('generatedCodeSoFar','var') || isempty(generatedCodeSoFar)
         generatedCodeSoFar = '';
     end
     
     if m2t.cmdOpts.Results.automaticLabels
+        lineTag = get(h,'Tag');
         if ~isempty(lineTag)
             labelName = sprintf('%s', lineTag);
         else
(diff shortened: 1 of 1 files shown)


src/matlab2tikz.m
```diff
@@ -1770,6 +1771,12 @@ function assertRegularAxes(handle)
 
         userWarning(m2t, 'Automatically added label ''%s'' for line plot.', labelName);
         generatedCodeSoFar = [generatedCodeSoFar, labelCode];
+        
+        if ~isempty(lineTag)
+            labelName = sprintf('%s', lineTag);
```
miscco: In the code above we have 

```matlab
labelName = sprintf('addplot:%s%d', name, m2t.automaticLabelIndex);
```
Isnt the addplot missing here?

miscco: Also i think, that automaticLabels and special tags are rather exclusive, so i would check for an label first and in the else clause add the automatic label

miscco:
@egeerardyn Why is there no jenkins test?

murmlgrmpf:
Well at least for me the addplot: prefix is just showing up in the tex labels: \label{addplot:DuctBode11}
So I chose not to append it.

The Jenkins test is passing!?

miscco:
@murmlgrmpf The Travis test is passing, but that only covers Octave as far as i know.

So the problem i see, is that the addplot comes from the auto clause. However, you want to give it an explicit name. So you should have a clause 
```matlab
if ~empty(lineTag)
 labelName = sprintf('addplot:%s', lineTag);
...snip
else
   [pathstr, name] = fileparts(m2t.cmdOpts.Results.filename); %#ok
   labelName = sprintf('addplot:%s%d', name, m2t.automaticLabelIndex);
   ...snip
end
```

murmlgrmpf:
@miscco What is the point of adding addplot: as a prefix to all labels? As far as i see, it is just bloating them.

miscco:
@murmlgrmpf Yeah, my mistake i have worded it correxctly above i think

murmlgrmpf:
Is it a bug? should the prefix be removed somewhere along the export?

murmlgrmpf:
I did not use an if else construction but rather just add another label for compatibility considerations.

miscco:
It is a bug as far as your code adds 2 labels and it only works when autolabels are enabled, which is most likely not intentional?

murmlgrmpf:
It was intended to add a second label instead of adding a different one.

miscco:
Hm, i dont know what the others think about it, but i would be strongly in favor of one label either automatic or explicit.

murmlgrmpf:
As of the autolabel: there may be use cases, where the tags are used in a different way, thus I would not export the tags as labels by default. adding another option would bloat the interface thus the solution I proposed.

murmlgrmpf:
A different approach would be to have some identifier in the Tag that matlab2tikz interprets and then adds the label:
Tag = 'tikzLabel:MyLabel'
and then append the label even when autolabel is disabled.
Better solution?

miscco:
Could you first provide a minimal workin example, so we get a better idea of the problem?

murmlgrmpf:
Where should I put the mwe mfile?

miscco:
If it is minimal, you can paste it here in 

```
"```matlab"
code
"```"
```

murmlgrmpf:
```matlab
x = linspace(0,10,10)
y = x.^2
hl = plot(x,y)

hl.Tag = 'tikzLabel:MyLine'

matlab2tikz('mweAutoLabelsByTags.tex','standalone',true)
```

murmlgrmpf:
Then I would parse the tag field in matlab2tikz for the 'tikzLabel:' flag and add the label accordingly.

egeerardyn:
@missco: Jenkins tests only run for team members for security reasons (otherwise any person can execute code on my server) as documented in [our wiki](https://github.com/matlab2tikz/matlab2tikz/wiki/Jenkins).
@matlab2tikzBot: test this PR

miscco:
Strange, there is always a m2t function you never knew about. So you want the directly include the tikz file into your LaTeX document and add the label inside?!

In that case i am definitely against the double label.

I would suggest the following:

1. You change 'automaticLabels' to 'addLabels' still being a boolean
2. You put it in an if else clause, so that only a specific label OR an automatic generated one wlll be added.

murmlgrmpf:
Allright with changing the name and adding the if else. Should i change this pull request or write another?

miscco:
Update the PR, thats the easiest way

murmlgrmpf:
Do you really want to change the name of the switch from automaticLabels to addLabels? It would potentially break existing code, though it would be cleaner.

murmlgrmpf:
btw: Yes, I directly include the tex in my document and I use reftikz.

egeerardyn:
I like this PR (haven't looked in detail, though), but I was aware that some label configurations are not always useful.

I thought about those labels a good while back (after @okomarov tackled legends for `plotyy`), and at that time, I had the impression we would be best off with "hierarchical" labels, i.e. labels that encode the location of the object in the HG tree structure that is a figure (the tag could indeed be a part of that).

@murmlgrmpf Please just add the code on top of your `AutoLabelsByTag` branch and push the new commits to your personal remote (likely `origin`), i.e. just update this PR (note: not force-pushing!).

miscco:
@murmlgrmpf: Hm, damn you backwards compatibility. The best solution is indeed, to just keep automaticLabels although it doesnt describe what it does any more and save the rename for some 2.0 release

egeerardyn:
Also, let's not change the name of the switch just yet. I agree that the proposed name is cleaner, but we should follow our practice of deprecation:
 - for now allow both parameters,
 - signal the old parameter as deprecated (e.g. [line 260](https://github.com/matlab2tikz/matlab2tikz/blob/master/src/matlab2tikz.m#L260))
 - write code to translate the old parameter to the new one (e.g. [line 302](https://github.com/matlab2tikz/matlab2tikz/blob/master/src/matlab2tikz.m#L302)),
 - use the new one everywhere else in the code.

murmlgrmpf:
Is the request now meeting your expectations?

murmlgrmpf:
We could still consider adding a prefix flag in order to avoid that someone accidentally uses Tags in matlab in combination with autolabels.
But even if this happens it would be easy to debug, as there are warnings anyway.

miscco:
@murmlgrmpf I think, this is a rather advanced feature, so accidents will most likely be not fatal.

As it is, it looks good. Although i would appreciate it, if you would follow @egeerardyn last comment about adding the new label and prepare the old one for deprecation. Adding explicit labels is in nearly every case the way to go.

@egeerardyn 
What do you think about checking for a Tag?

egeerardyn:
I don't think many people use those automatic labels (let alone those tags, which are typically reserved for special plot function anyway). Personally, I have thought before about using those Tags as ways to declare the labels and I know that it may possibly introduce duplicates, but I just love the idea of being able to specify it like that. So I would greenlight that idea.
If people are unhappy with it, we can remove it again and introduce that together with `m2tcustom` in #832 (just another field called `Label`).

I just have 2 comments:
 - we can still deprecate this `automaticLabels` switch (see above)
 - I would rather move the responsibility of getting the tag to `addLabel`, i.e. such that it has a function signature:
```matlab
[m2t, str] = addLabel(m2t, str, handle)
```
rather than the current
```matlab
[m2t, str] = addLabel(m2t, str, get(handle, 'Tag'))
```.

But that is nitpicking from my end.
Anyhow, as far as I'm concerned this is :+1:

**Comment by miscco** [comment]
:+1: for the 
```matlab
[m2t, str] = addLabel(m2t, str, handle)
```

murmlgrmpf:
I have changed the signature according to @egeerardyn proposal.
As of the depreciation warning: should I move the definition of automaticLabels to line 260ff or should I put the depreciation warning to the place where it is defined now?

murmlgrmpf:
Now I have deprecated automaticLabels added addLabels, changed the documentation and enabled both switches in the code.
In future one will just need to remove the parameter definition and the deprecation notice.

egeerardyn:
:+1: Thank you for your contribution!

PR #846, opened 2016-01-19
Cleanup of draw options gathering

miscco:
So i looked into bar plots due to the issues we have with them. I found, that a lot of the draw options handling can be generalized for different plots, e.g. histogram/barplots/areas

So i did that and cleaned up various other occurrences of opt_print while i was at it.

I did not bother to check all the differences yet, as i would like to know whether there is interest at all int that PR

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -2325,7 +2325,6 @@ end
 % ==============================================================================
 function [m2t, str] = drawImage(m2t, handle)
     str = '';
-
     if ~isVisible(handle)
         return
     end
@@ -2551,6 +2550,9 @@ end
 % ==============================================================================
 function [m2t, str] = drawContourHG2(m2t, h)
     str = '';
+    if ~isVisible(h)
+        return
+    end
 
     % Retrieve ContourMatrix
     contours = get(h,'ContourMatrix')';
@@ -2621,6 +2623,10 @@ function [m2t, str] = drawFilledContours(m2t, h, contours, istart, nrows)
     % defined by inclusion, i.e. its members are contained within one
     % outer contour. The outer contours of two groups cannot include
     % each other.
+    str = '';
+    if ~isVisible(h)
+        return
+    end
 
     % Split contours in cell array
     cellcont = mat2cell(contours, diff([istart; nrows+1]));
@@ -2717,8 +2723,8 @@ function [m2t, str] = drawFilledContours(m2t, h, contours, istart, nrows)
         % Print out
         drawOpts = opts_print(m2t, drawOptions,  ',');
         tabOpts  = opts_print(m2t, tableOptions, ',');
-        str      = sprintf('\\addplot[%s] table[%s] {%%\n%s};\n', ...
-                           drawOpts, tabOpts, table);
+        str      = sprintf('%s\\addplot[%s] table[%s] {%%\n%s};\n', ...
+                           str, drawOpts, tabOpts, table);
     end
 end
 % ==============================================================================
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -4267,8 +4267,8 @@ function [m2t, drawOptions] = getPatchDrawOptions(m2t, h, drawOptions)
     faceColor          = get(obj, 'FaceColor');
     [m2t, drawOptions] = setColor(m2t, h, drawOptions, 'fill', faceColor, 'none');
 
-    % FaceAlpha
-    faceAlpha = get(h, 'FaceAlpha');
+    % FaceAlpha (Not applicable for MATLAB2014a/b)
+    faceAlpha = getOrDefault(h, 'FaceAlpha', 'none');
     if ~isNone(faceColor) && isnumeric(faceAlpha) && faceAlpha ~= 1.0
         drawOptions = opts_add(drawOptions, 'fill opacity', sprintf(m2t.ff,faceAlpha));
     end
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -1801,7 +1801,7 @@ function [m2t,str] = plotLine2d(m2t, opts, data)
     end
 
     % Print out
-    tabOpts = opts_print(m2t, tableOptions, ',');
+    tabOpts = opts_print(m2t, tableOptions, ', ');
     str     = sprintf('\\addplot [%s]\n %s table[%s]{%s};\n',...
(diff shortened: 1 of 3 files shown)


src/matlab2tikz.m
```diff
@@ -1784,20 +1787,23 @@ function assertRegularAxes(handle)
 function [m2t,str] = plotLine2d(m2t, opts, data)
     errorbarMode = (size(data,2) == 4); % is (optional) yDeviation given?
 
-    str = '';
+    errorBar = '';
```

egeerardyn:
Could you give a sumary what you have done in this big commit?
As far as I see:
 - moved some functions around
 - some renaming of variables
 - some refactoring of `opts_print` calls and `*Opts` vs `*Options` (:+1: for that)
 - handle some properties of areaseries, barplots, ... in a separate function (:+1: again)

However, I think in this big commit the latter is the most important, but I tend to lose track of those changes in that big commit. Please, in the future, do keep everything as separate commits. It's easier to review and it allows to track regressions more easily. I rather get 50 tiny commits (that each produce working code) than 1 big one (which does not produce working code yet).

okomarov:
We should just make it a rule. Moving functions should be addressed in separate PRs! By moving I mean cut/paste with NO edits (not even a single char). Also, whitespace edits should be done in separate PRs. Basically, in these 2 cases, we could merge straight away or even the author of the PR, once he gets green light, he merges right away.

It is basically impossible to review this PR without opening Matlab, but often all is needed is a quick overview of the code, which can be done in github directly.

egeerardyn:
@okomarov Good points. Let's put those in the `contributing.md` file.

miscco:
Yeah, as always one thing lead to the other ;)

I wanted to have a look at drawBarSeries as there are some issues with barplots. I compared it with its brethren drawArea, drawHistogram and friends and realized, that a lot of code is not only shared, but one function checked for Opacity, the other did not and so on and so forth. 

The one code movement i mande was drawArea, as this i closely related to drawHistogram and drawBarplot at least from the involved code, but i can revert that.

Also i found, that at the end of the function we often have some sprintf with an opt_print as argument, what i belive is bad style. Therefore i went havoc and changed everything into:
```
opts_print(Option1)
...
opts_print(OptionN)
str = sprintf()
```

miscco:
Ah and i renamed one str to errorbar, as that function created an errorBar.

I would also remove all useless occurences of 
```
str='';
```
That are in a lot of functions and only needed, when there is an early exit

egeerardyn:
Overall, I think you did a great job, but in the future I would advise against doing all those things in the same commit. 

Maybe a little tip to avoid just that: you can do `git add --patch --interactive` to pull apart multiple unrelated edits in your working copy and only commit parts of them to git. I find it that useful that on my computer, `git pa` is an alias to that and I use it very often. Because such humongous commits come to be very naturally, but they should be avoided if possible.

If you can get all tests back to green (and an overview of what has changed in the output per test case and per environment), this is a :+1:.

miscco:
So here are the differences between develop branch and this PR for matlab2014a,/b and octave (Your server is very helpfull for that :smile: )

[diff_DrawOpts.zip](https://github.com/matlab2tikz/matlab2tikz/files/97588/diff_DrawOpts.zip)

So for matlab2014a and octave there is some reordering in the argument list of addplot and i missed a space after a the commata in opts_print. I will change that in a separate PR, once this is merged, to catch all occurrences of opts_print. it does increase readability somehow.

For 2014b it is rather strange. Now ACID(10) and ACID(98) seem to generate less tables?! I will investigate that later

egeerardyn:
If the current situation is that the options are `\addplot [fill=..., xbar, bar width=...]`, I would prefer to have the old ones back: `\addplot [xbar, bar width=..., fill=...]` if this can be done without a lot of work.

I like having `xbar` etc. first, because that conveys the essence of those plots (it's a bar plot) rather than details (it has color ABCDEF).

Please add the space after the comma already where you introduced the regression, that will get rid of many problems already (without updating any hashes).

Afterwards, I would suggest we just place a default separator argument in `opts_print` and get rid of all these arguments elsewhere (except maybe for the ones in `axes`, since those are something like `sprintf(', \n')` if I remember correctly.  But that's something that's perhaps best for another PR (since that may cause a lot of merge conflicts).

miscco:
Ah i found the error, contour plots used the sprintf('%S\\addplots...) to add multiple contours...

miscco:
@egeerardyn  I have changed the ordering and hopefully recovered all spaces. ;)

miscco:
So as far as i can tell the differences amount to:
```
< \addplot[ybar,bar width=0.8,draw=black,fill=mycolor1,area legend] plot table[row sep=crcr] {%
---
> \addplot[ybar,bar width=0.8,fill=mycolor1,draw=black,area legend] plot table[row sep=crcr] {%
```

```
< \addplot[fill=mycolor1,fill opacity= 0.6,draw=black,ybar interval,area legend] plot table[row sep=crcr] {%
---
> \addplot[ybar interval,fill=mycolor1,fill opacity= 0.6,draw=black,area legend] plot table[row sep=crcr] {%
59c59
< \addplot[fill=mycolor2,fill opacity= 0.6,draw=black,xbar interval,area legend] plot table[row sep=crcr] {%
---
> \addplot[xbar interval,fill=mycolor2,fill opacity= 0.6,draw=black,area legend] plot table[row sep=crcr] {%
```

Is that an acceptable change?

okomarov:
The changes look harmless but in other cases the ordering of the options might matter and the pgfplots manual should be consulted.

The code looks ok to me.

egeerardyn:
Code looks OK to me. @okomarov is right that the order matters in some cases, but I would be surprised if this is one of them.

I had a look at the manual and couldn't find anything:
 - `xbar interval` is just `xbar` with some coordinate magic on top
 - `xbar` sets `xbar legend` and some coordinate stuff
 - `fill`, `fill opacity`, `draw` are TikZ primitives as far as I know
 - `area legend` overrides which kind of image is shown, but doesn't do anything else

The only thing I wonder is, maybe we should get rid of that `area legend`? I think the `xbar legend` one looks nicer (even if you think the horrible pattern in the area legend away):
screen shot 2016-01-20 at 19 03 34
screen shot 2016-01-20 at 19 02 58

So as far as I'm concerned, :+1: for changing the order.

miscco:
So everything is working now. For philosophical reasons, I would prefer if someone else merges my PRs

egeerardyn:
:+1: Thank you!

PR #847, opened 2016-01-22
Cleanup gathering of lineoptions

miscco:
We had a function that generated lineOptions, but strangely put the getters for those options outside the function, passing them as arguments. 

Clean that up and use the getLineOptions whenever possible. Also utilize the isLineVisible function, when LineStyle and LineWidth are only used to check for visibility.

I have kept the generally used output [m2t, Options] however i am unsure if we shouldnt reduce it to just Options when m2t is not changed


egeerardyn:
:+1: I have nothing to add, remove or complain about. Great work!

PeterPablo:
I once tried to cleanup `[m2t, Options]` to be `Options` for cases where the `m2t` structure is not touched. Well done! :+1:

PR #848, opened 2016-01-22
M2t cleanup

miscco:
So I went ahead and started some more cleaning of the code. Generally i tried to first remove all warnings in matlab.

I also encapsulated functions that were passed unnecessary arguments. For example addLegend received the full string of the plot as an argument, when it only appended the legend. So just return a string without the input and append in the calling function.

I have not yet made the change for adding a whitespace after commata, as that would change the hashes.


src/matlab2tikz.m
```diff
@@ -551,13 +551,14 @@ function addStandalone(m2t, fid, part)
     ff  = colorFormat;
     if ~isempty(names)
         m2t.content.colors = sprintf('%%\n%% defining custom colors\n');
+        colorDef = cell(1, length(names)+1);
```
egeerardyn: Why `length(names) + 1`?

src/matlab2tikz.m
```diff
@@ -1163,22 +1164,21 @@ case guitypes()
...
+    str  = getOrDefault(h, 'displayname', '');
 
     % HG1: autogenerated legend strings, i.e. data1,..., dataN, do not populate
     % the 'displayname' property. Go through 'userdata'
-    if isempty(string)
+    if isempty(str)
         ud     = get(m2t.axesContainers{end}.LegendHandle,'userdata');
         idx    = ismember(ud.handles, h);
-        string = ud.lstrings{idx};
+        str    = ud.lstrings{idx};
```
egeerardyn: Spacing can be adjusted here.

src/matlab2tikz.m
```diff
@@ -1677,16 +1674,18 @@ function assertRegularAxes(handle)
...
+    [m2t, labelString] = addLabel(m2t, h);
+    
+    str = [dataString, labelString];
 end
 % ==============================================================================
-function [m2t, str] = specialDrawZplaneUnitCircle(m2t, h, drawOptions)
+function [m2t, str] = specialDrawZplaneUnitCircle(m2t, drawOptions)
     % Draw unit circle and axes.
 
     % TODO Don't hardcode "10", but extract from parent axes of |h|
```
egeerardyn: In the light of this comment, I would keep passing in this `h` into the function. If you want, you can add an `%#ok` after the function declaration to suppress the message (it will be re-enabled automatically once we do use `h`)

src/matlab2tikz.m
```diff
@@ -1762,12 +1766,10 @@ function assertRegularAxes(handle)
     end
 end
 % ==============================================================================
-function [m2t, generatedCodeSoFar] = addLabel(m2t, generatedCodeSoFar, h)
+function [m2t, labelCode] = addLabel(m2t, h)
```
egeerardyn: :+1: really nice!

src/matlab2tikz.m
```diff
@@ -2848,7 +2850,7 @@ function assertRegularAxes(handle)
         annotPanes   = findall(m2t.currentHandles.gcf,'Tag','scribeOverlay');
         children = allchild(annotPanes);
         if iscell(children)
-            children = [children{:}]
```
egeerardyn: Good find. I don't know whether that path is covered, maybe we should place a `TODO: is this dead code?` here to doublecheck that?

src/matlab2tikz.m
```diff
@@ -3099,16 +3103,14 @@ function assertRegularAxes(handle)
...
     % Different from the default MATLAB behavior. To fix this, one could use
     % /pgfplots/after end axis/.code.
-
-    str = '';
-
+    
     % - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
     % get required properties
-    String = get(handle, 'String');
+    textString = get(handle, 'String');
```
egeerardyn: Somehow, I don't like the name `textString` (not that `String` was any better, though), but I think it is somewhat pleonastic: text is a string (or a string is text).

How about `contents`? Or `label`? Or just `text`?

egeerardyn:
I like it. It brings back memories, as this is exactly how I started out to tinker with m2t (back then we only had the graphical tests, so I remember my frustrations with trying to get the dimensions right).

Anyhow, this PR makes a very old private branch of mine superfluous :+1:. Thank you for cleaning that up!
I think some of the `%#ok` things depend on the settings one uses for `mlint` (e.g. complaining about `catch` vs `catch ME`), so perhaps we should provide the settings file in the repository as well.

src/matlab2tikz.m
miscco: I believe we can even make something like 
```matlab
str = strjoin([colorDef, sprintf('%%\n')], '');
```

src/matlab2tikz.m
miscco: I will have a look what it does

src/matlab2tikz.m
miscco: :+1:

src/matlab2tikz.m
miscco: Yeah, content might be better, i couldnt think of anything better given that the function is called drawText

miscco:
So i found, that we have the function join(m2,...) that liks to 
```
m2tstrjoin(cellstr, delimiter, m2t.ff)
```
And that function basically calls strjoin with 'floatFormat' mt.ff. What is the significance of that format?

PeterPablo:
This relates to the data format that can be specified by the user.

miscco:
What are the usecases symbolic languages like Chinese or Korean?

egeerardyn:
@miscco No. Rather which precision and formatting of the data (think `sprintf`/`fprintf`): `%0.15f`, `%8.6g`, `%0.3e`, ...
That is an important part that keeps our hash-based testing viable (that, and it is an important feature to keep the file size down and compilable by LaTeX).

PeterPablo:
`m2t.ff` is the shortcut to the float format (https://github.com/matlab2tikz/matlab2tikz/blob/3a1ee10dee3204e15565e95720424ed9e8a8caca/src/matlab2tikz.m#L283).

miscco:
But as far as i understand it coorectly, we only need that in those steps, that write to file?

PeterPablo:
that's not correct, as the content that goes into the file is assembled in
numerous places, but the actual writing of the file takes place only once
at the very end. the assembly steps rely on the number format.

egeerardyn:
> But as far as i understand it correctly, we only need that in those steps, that write to file?

@miscco I don't really understand what you're trying to say there. Can you clarify what "that" is?

miscco:
So addPlotyyReference is a tricky one. I reverted the change of that function, to try later more. 
I also changed getLegendString() to include the prettyprint and join that we always do with the output

miscco:
Kill it with fire
```matlab
m2t.axesContainers{end}.options = ...
            opts_add(m2t.axesContainers{end}.options,
```

miscco:
So I like your name better. The fact, that we simply replace code 1:1 makes me feel rather save. Tampering with m2t is always dangerous though. I have added the separating line, but kept it at its place, as that is the most logical one i can think of.

egeerardyn:
> So I like your name better. The fact, that we simply replace code 1:1 makes me feel rather save. Tampering with m2t is always dangerous though. I have added the separating line, but kept it at its place, as that is the most logical one i can think of.

:+1: I agree with you that the replacement is safe and probably it will help us even if we try to untangle the `m2t` object since the locations of access are more concentrated now.

> Renamed numEnvironements to envCounter

:+1:

I'm happy with this whole PR. Are you still working on it (please put a work-in-progress label then), otherwise this can be merged.

miscco:
@egeerardyn Thanks a lot for the reviews, I thought about reworking matlab2pgfplotsColormap to remove colors of the colormap rathern than clampering it together, but meh, thats something for another day.

egeerardyn:
@miscco Let's do the colormaps stuff another day. There is a TODO somewhere in there, so we will look at it again in the future.

src/matlab2tikz.m
```diff
@@ -712,19 +714,24 @@ case guitypes()
...
 
         % append the environment
-        pgfEnvironments{end+1} = str;
+        pgfEnvironments{envCounter} = str;
+        envCounter = envCounter +1;
     end
 end
 % ==============================================================================
-function [m2t, str] = addPlotyyReference(m2t, str, h)
+function [m2t, ref1, ref2] = addPlotyyReference(m2t, h)
```
okomarov: @miscco ref1 and ref2 are actually wrong if they refer to axis 1 and 2. Ref1 should is the latex label and ref2 should is the reference to that label.

miscco: @okomarov Isnt it the other way around 
```matlab
ref1 = sprintf('\\label{plotyyref:leg%d};\n\n', labelNum);

refs{ii}= sprintf('\\addlegendimage{/pgfplots/refstyle=plotyyref:leg%d}\\addlegendentry{%s};\n',...
                              ii, lString);
```
I would have thought, that ref1 is the labelReference and ref2 would be label

miscco: @okomarov screw the last one, too few coffee and too much work

PR #849, opened 2016-01-25
Simplify splitLine

miscco:
So this one poped up, when i already had set the cleanup as ready.

Was easier than expected and simplifies the function tremendously


egeerardyn:
:+1:

Issue #845, opened 2016-01-19
Stacked bar charts


redfixie:
Hi - Stacked bar charts don't compile correctly, the last 'stack' goes on top of all others. In order to create stacked bar charts after 
*\begin{axis}[%*
you must add the line
*ybar stacked,*

egeerardyn:
Thank you for reporting this.
Do you by any chance have a piece of MATLAB code to reproduce this and a screenshot of what you are seeing?

I'm going on a limp here, but I think your version of `pgfplots` might be somewhat outdated. I.e. `matlab2tikz` should already be able to reproduce stacked plots as you see below:
<img width="1114" alt="screen shot 2016-01-19 at 11 08 33" src="https://cloud.githubusercontent.com/assets/177360/12415663/3cc7b0da-be9d-11e5-8794-32411b19dee3.png">
You should be able to reproduce the right plot by going into the directory `test/suites` within your matlab2tikz installation, run `ACID(34)` and then `matlab2tikz` on that figure (you can use the flags `'standalone',true` for easier compilation).

So can you also check the version of the software you are using?
 - matlab2tikz (this gets output to the terminal during execution),
 - MATLAB(type `version` in MATLAB),
 - pgfplots, by compiling the following document with LaTeX:
```latex
\documentclass{standalone}
\usepackage{tikz,pgfplots}
\begin{document}
TikZ: \expandafter\csname ver@tikz.sty\endcsname \\
pgfplots: \expandafter\csname ver@pgfplots.sty\endcsname \\
\end{document}
```
That should produce a document that looks like
<img width="345" alt="untitled" src="https://cloud.githubusercontent.com/assets/177360/12415816/164d1426-be9e-11e5-8d02-020e09b7a6c9.png">

egeerardyn:
Since this is likely a problem with an outdated pgfplots version, I will close this. 
If you are using a recent version of pgfplots and still have this issue, feel free to reopen.

Issue #840, opened 2016-01-08
Bar plot with small Y-Range


SimonEbner:
I noticed matlab2tikz has an issue when the Y-Range of values used in a bar-plot is small. The following matlab code

```
filepath =  ...
Y1 = randn(100,1);
Y2 = 1e-5*Y1;
% Std: 1
subplot( 121 );
[N, x] = hist(Y1, 15);
bar(x,N);
% Std: 1e-5
subplot( 122 );
[N, x] = hist(Y2, 15);
bar(x,N);

matlab2tikz( filepath, 'parseStrings', false, ...
    'ExtraAxisOptions', 'xlabel near ticks, ylabel near ticks', ...
    'showInfo', false, ...
    'standalone', true );
```

produces these matlab plots
![matlab](https://cloud.githubusercontent.com/assets/570249/12200488/5a4e61d4-b61f-11e5-9753-8345f22d280c.png)

but the corresponding tikz image looks like this:
![screenshot from 2016-01-08 15 49 39](https://cloud.githubusercontent.com/assets/570249/12200504/7253506e-b61f-11e5-9d5f-660b6c2640b1.png)

egeerardyn:
I think this is related to #785. We should certainly look at this at the same time.

miscco:
So i looked a bit into that and the problem seems to be the lack of precision within LaTeX. Currently we only consider the first three digits (formatDim), which i this case leads to a barwidth of 0.

However, manually editing the war width to the calculated value (2.8e-06) doesnt fix the problem, so i believe this is intrinsic to LaTeX. I bet the simplest solution would be to increase the XData by e5 and fake the xlabels

SimonEbner:
Thank you very much for the update. I will workaround as proposed by you.
Am 26.01.2016 21:06 schrieb "Michael Schellenberger Costa" <
notifications@github.com>:

egeerardyn:
You can also set the bar width to a physical unit (e.g. `2mm`) in pgfplots, maybe that helps at least to edit the figure.

miscco:
Playing around i found, that for that specific case removing the bar width would also work. Would you think, that this is a solution?

SimonEbner:
Thanks so much to both of you for the suggestions. I'll try these two options tomorrow and will get back to you.

Issue #785, opened 2015-10-07
Misplacement of bars by using bar(x,y)


elektronxyz:
Hello,

I discovered an issue in the current development branch. Using Matlab bar(x,y) with a random x-axis lead to an overlapping of the bars due to wrong values of "bar shift=" in the tikz file. 

Example Code:
x=13.64:0.060000:14.12;
y = rand(9,4);
bar(x , y);
matlab2tikz('barxy_placement_error_result.tikz','width','\figW','height','\figH');
           

Matlab Result:
![barxy_error_reference](https://cloud.githubusercontent.com/assets/15018040/10343849/55ce81ea-6d20-11e5-8481-3f90a1ea864d.jpg)

Latex Result:

![barxy_error_result](https://cloud.githubusercontent.com/assets/15018040/10343911/aff1c04c-6d20-11e5-8d8d-72f5a0905bc9.jpg)

egeerardyn:
Thank you for reporting this and providing an MWE! 
I can confirm this behavior on R2015.

miscco:
I am pondering, whether this should be part of cleanfigure or not. Essentially, it is a question of normalization and regularization of the figure data, which hints at cleanfigure. On the other hand cleanfigure might not be used everywhere.

egeerardyn:
It should *not* go in `cleanfigure` since that would probably require us to map a `barseries` object with N bars and M groups to M `barseries` objects with N bars and 1 group. That would require to cope properties, some of which might be protected/private, some of which are impossible to translate from 1 object to M objects (e.g. event handlers, appdata, ...). I.e. a lot of trouble.

From within `matlab2tikz`, it makes more sense to tackle this. We just need to make sure that we calculate the right shifts and bar widths in `matlab2tikz` and then we're done. So we don't need to write back to HG objects, which keeps the complexity down.

Personally, I think `cleanfigure` should only be used for rare cases where you have "special" "data" (can be text as well) in the figure (causing LaTeX crashes (outside of the range), long compilation (too much data), ...). Those are not conversion bugs, but rather workarounds for implementation limitations in LaTeX. `matlab2tikz` on the other hand should always produce semantically valid TikZ/LaTeX code (even though a practical implementation of LaTeX/pgfplots might fail to compile it without actually calling `cleanfigure`).

In other cases, e.g. #644, the distinction whether something should go in `cleanfigure` or `matlab2tikz` is less obvious.

miscco:
[...]

Issue #524, opened 2015-01-16
Tools for debugging


egeerardyn:
Since we have been looking at automatic testing, a lot of version-dependent output has disappeared from out TikZ output, but this is crucial to help us bug-hunt some issues.

This feature request is to overcome that hiatus. We would need a functionality in Matlab2tikz that gives us access to the same information as before and possibly even a bit more:
* current version of M2T (and git commit-ish if present)
* MATLAB/Octave version
* OS and version on which the user is running
* :new: installed toolboxes?
* :new: parameters passed to M2T?
* :new: LaTeX information: e.g. LaTeX distribution, pgfplots version, TikZ version, ... (when LaTeX can be found on the PATH).
* :new: contents of the user's `get(0,'Default')`

Personally, I would prefer just a flag added to M2T to print such information to stdout or a file and have the user copy that information to a bug report. I think that's the most user-friendly approach and it gives us access to the internal state of M2T (e.g. version).

If there are things that I've missed in this list, please just add them :smile:

PR #851, opened 2016-01-26
Fix barplots

miscco:
Hi,

this is a try on #785 and hopefully later also #845 and #840 . 

Currently barplots are stacked on top of each other. This can be avoided by removing the "bar shift" argument and adding "xbar/ybar" to the axis environement, as following the pgfplots manual this calculates the shifts automatically.
```latex
/pgfplots/xbar={<shift for multiple plots>} (style, default 2pt)
This style sets /tikz/xbar and some commonly used options concerning horizontal bars 
for the complete axis. This is automatically done if you provide xbar as argument to an 
axis argument, see above. The xbar style defines shifts if multiple plots are placed 
into one axis. It draws bars adjacent to each other, separated by
 <shift for multiple plots>. Furthermore, it sets the style bar cycle list and sets tick and
legend appearance options.
The style is defined as follows.
```
```latex
\pgfplotsset{
    /pgfplots/xbar/.style={
        /tikz/xbar,
        bar cycle list,
        tick align=outside,
        xbar legend,
       /pgfplots/bar shift auto={#1},
    },
}
```

Also add barplots to octave, as the patch workaround is rather ugly

### TODO
 - [x] Is xbar/ybar in the addplots necessary  ->NO
 - [x] Add bar schift auto?
 - [x] Investigate barplot width (#840) -> Most likely a precision problem
 - [x] Stacked barplots? (#845) -> Seems to stem from outdated pgfplots

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -2856,7 +2856,8 @@ function cl = guessOctavePlotType(h)
 
         % quiver plots
         % TODO: maybe check for existence of `udata` and absence of `ldata`
-
+    elseif hasProperties(h, {'bargroup','barwidth', 'barlayout'}, {})
+        cl = 'specgraph.barseries';
         % unknown plot type
     else
         cl = 'unknown';
@@ -3801,6 +3802,9 @@ function [m2t, str] = drawBarseries(m2t, h)
     else
         barType = 'ybar';
     end
+    m2t = m2t_addAxisOption(m2t, barType);
+
+    % Get the draw options for the layout
     [m2t, drawOptions] = setBarLayoutOfBarSeries(m2t, h, barType, drawOptions);
 
     % Get the draw options for the bars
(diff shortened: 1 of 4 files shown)


egeerardyn:
@miscco: The problem with `xbar`/`ybar` in the `axis` environment is that it then becomes impossible to make plots with both bars and lines.

I suspect the only real solution is by determining the bar width in m2t (or doing the dirty thing we did before: if I remember correctly, just translate the underlying patches in HG1/Octave and unsupported in HG2)

miscco:
Would adding a sharp plot in that case help? This overrides the ybar command. Optionally, one could rather add 
```
bar shift auto
```
To the axis, as that only counts for bar plots

egeerardyn:
`sharp plot` would be a mess for consistency, but if there is an `bar shift auto` option, that could be a good first try. Eventually, we should, however, also support changing the bar widths.

miscco:
@egeerardyn first :smiling_imp:

egeerardyn:
@miscco Can you also make sure you test it with differently scaled values on the X-axis/Y-axis (our test cases work well when those are around 1, but not with other scales (as you can see from the different bug reports)).

okomarov:
`bar shift auto` requires pgfplots 1.13 and #845 is probably due to outdated version of the user.

miscco:
True, but the bug in bar shift is still fixed, so we should add a note for pgfplots?

egeerardyn:
@miscco There is a function `needsPgfplotsVersion` that is there especially for signaling that a minimal version is needed. I think in the future that we should be even more verbose about that version, since bugs regarding outdated pgfplots version (as #845) are quite common among our users. Recently, one of my colleagues who was starting out with m2t was having a similar problem (but at my department, quite a lot of computers still have pgfplots 1.3 or 1.4).

In the long run, I hope (e.g. #524) that we can add some information about the `pgfplots` version that is available on the system and show clear warnings.

miscco:
So given that there is the check for the pgfplots version and it does fix a bug are there any resentments in me updating the hashes and merging it?

egeerardyn:
Oh, no resentments :smile: If the produced files compile fine and produce decent figures, then mess around with the hashes as you please and merge :+1:

miscco:
Given that this has been acked by @egeerardyn I will merge this tomorrow.

egeerardyn:
This time I'm faster 😛 great job though!

Issue #852, opened 2016-01-27
surface plot not working


pfbuxton:
Hello,

Here is a test code which fails:

<code>
clear;

x = linspace(0,10,100);
y = x.^2;
c = y.^5;

surface(...
  'XData',[x(:) x(:)],...
  'YData',[y(:) y(:)],...
  'ZData',zeros(length(x(:)),2),...
  'CData',[c(:) c(:)],...
  'FaceColor','none',...
  'EdgeColor','flat',...
  'Marker','none', ...
  'LineWidth',4);

matlab2tikz( 'test.tikz', 'standalone',true );

dos( [ 'pdflatex test.tikz"' ] );

</code>

pfbuxton:
I have found an alternative which unfortunately also does not work (at least on my system):

```matlab
clear;

x = linspace(0,10,100);
y = x.^2;
c = y.^5;

patch( [x fliplr(x)],[y+1.0 fliplr(y)-1.0], [c fliplr(c)], 'linestyle','none' );

matlab2tikz( 'test.tikz', 'standalone',true );

dos( [ 'pdflatex test.tikz"' ] );
```

miscco:
That one seems like a simple one.
isFaceColorFlat

is not defined in getPatchShape.

I will submit a patch for that

miscco:
So the second one should be fine by defining isFaceColorFlat, however it seems the first one has some problems, as it does not compile correctly. It seems some color keys are incorrect:
```
! Package pgfkeys Error: The key '/tikz/color' requires a value. I am going to 
ignore this key.
```

The culprit has this form:
```latex
\addplot[%
mesh,
color,colormap={mymap}{[1pt] rgb(0pt)=(0.2081,0.1663,0.5292); rgb(1pt)=(0.211624,0.189781,0.577676); rgb(2pt)=(0.212252,0.213771,0.626971); 
...
rgb(62pt)=(0.9661,0.951443,0.0755333); rgb(63pt)=(0.9763,0.9831,0.0538)},mesh/rows=2]
table[row sep=crcr, point meta=\thisrow{c}] {%
```

miscco:
Here the reswults for the first plot:
![issue_852_matlab](https://cloud.githubusercontent.com/assets/3017120/12624559/0410cf26-c530-11e5-8a6d-b723f2aedc82.jpg)

And the resulting pdf
[...]

PR #853, opened 2016-01-27
Fix missing definition of isFaceColorFlat

miscco:
As the title says, it is not defined in getPatchShape. See #852 

To preserve encapsulation, get it from patchOptsion


miscco:
For correctness i should mention, that this does not fix the first figure presented in #852. There some color definitions seem to be astray.

src/matlab2tikz.m
```diff
@@ -2305,6 +2305,7 @@ function assertRegularAxes(handle)
             userInfo(m2t, '\nMake sure to load \\usepgfplotslibrary{patchplots} in the preamble.\n');
 
             % Default interpolated shader,not supported by polygon, to faceted
+            isFaceColorFlat = isempty(strfind(opts_get(patchOptions, 'shader'),'interp'));
```
okomarov: Well done for spotting. Our current ACID never hits this condition, hence it never raised an error. There is a patch suite that might actually have some plots of this type.

Issue #855, opened 2016-02-04
"forget plot" breaks latex


WDuck:
Its me again, same example (Matlab R2015b) as an our ago but other issue.
```matlab
s=tf('s');
T=1/(s+1);
stepplot(T);
cleanfigure;
legend('Test Function', 'Location', 'SouthEast');
matlab2tikz('stepplot_test.tikz', 'height', '\figureheight', 'width', '\figurewidth');
```

Now get the  wanted result including the legend text with matlab2tikz `development` branch:
```latex
% This file was created by matlab2tikz.
%
%The latest updates can be retrieved from
%  http://www.mathworks.com/matlabcentral/fileexchange/22022-matlab2tikz-matlab2tikz
%where you can also make suggestions and rate matlab2tikz.
%
\definecolor{mycolor1}{rgb}{0.00000,0.44700,0.74100}%
%
\begin{tikzpicture}

\begin{axis}[%
width=0.925\figurewidth,
height=0.942\figureheight,
at={(0.026\figurewidth,0\figureheight)},
scale only axis,
separate axis lines,
every outer x axis line/.append style={white!40!black},
every x tick label/.append style={font=\color{white!40!black}},
xmin=0,
xmax=9,
every outer y axis line/.append style={white!40!black},
every y tick label/.append style={font=\color{white!40!black}},
ymin=0,
ymax=1,
axis background/.style={fill=white},
legend style={at={(0.97,0.03)},anchor=south east,legend cell align=left,align=left,draw=white!15!black}
]
\addplot [color=mycolor1,solid, forget plot]
  table[row sep=crcr]{%
0	0\\
0.0460517018598701	0.0450074139785546\\
...
9.02613356453466	0.999879773556536\\
};
\addlegendentry{Test Function};

\addplot [color=black,dotted, forget plot]
  table[row sep=crcr]{%
-0.9	1\\
9.89999999999998	1\\
};
\end{axis}

\begin{axis}[%
width=0.951\figurewidth,
height=\figureheight,
at={(0\figurewidth,0\figureheight)},
scale only axis,
xmin=0,
xmax=1,
xtick={\empty},
xlabel={Time (seconds)},
ymin=0,
ymax=1,
ytick={\empty},
ylabel={Amplitude},
hide axis,
title style={font=\bfseries},
title={Step Response},
axis x line*=bottom,
axis y line*=left
]
\end{axis}
\end{tikzpicture}%
```

[...]

PR #856, opened 2016-02-04
Only add forget plot if it is not the last plot of an axis

### LINKED ISSUE #799: ACID test 42 produces uncompilable code
I have found, that ACID(42) produces uncompilable code for me in MALTAB2015b.

the culprit is the following:
```
\addplot [color=mycolor2,solid,line width=1.5pt,forget plot]
```

This is the last plot in the axis. removing the "forget plot" lets the code compile fine. Even if i *only* remove it there in the last plot of that axis.

miscco:
It seems, that the intention was, that the last plot of an axis never gets an "forget plot" however, this requires an && rather than a ||.

As far as I know this fixes #799 and #855

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -1764,7 +1764,7 @@ function [m2t, str] = writePlotData(m2t, data, drawOptions)
             % entry for the *last* occurrence of the plot series.
             % Hence the condition k<length(xDataCell).
             %if ~isempty(m2t.legendHandles) && (~m2t.currentHandleHasLegend || k < length(dataCell))
-            if ~m2t.currentHandleHasLegend || k < length(dataCell)
+            if ~m2t.currentHandleHasLegend && k < length(dataCell)
                 % No legend entry found. Don't include plot in legend.
                 hiddenDrawOptions = maybeShowInLegend(false, drawOptions);
                 opts = opts_print(m2t, hiddenDrawOptions, ',');


miscco:
@okomarov can you have a look at it, its just the one line and from your comment it seems the correct choice.

egeerardyn:
@okomarov does not have the Control Toolbox available. I will do the review instead (I'm thinking whether the change really makes sense: almost ALL plots are affected by this, which goes against my intuition).

miscco:
Yeah, i am currently running it through your server, to test what is happening there. However, if the only problem is, that the last plot doesnt get a forget plot anymore, then it should be valid?!

okomarov:
If I remember correctly, from one figure of a bode plot that @egeerardyn provided some time ago,  these special plots create legend on empty layers. In other words, the actual lines are on a gca that has no legend and the legend is on another gca. There are two reasons why this is a clusterf*** (again if I recall correclty): these two axes do not have a plotyy structure (which we can detect), and their ordering (z-stacking) is actually the opposite of that of plotyy...so good luck with that ahaha

egeerardyn:
I think this can't fix the issue.
What this If is supposed to do is to avoid that split plots (plots that we split to avoid latex buffer overflows) only get a single legend entry.

So we need to add "forget plot" when either:
- the plot has no legend, or
- we are writing any section but the last section of the plot.

Which, as I see it, is exactly what the old code does.

egeerardyn:
@okomarov @miscco I will try to dissect one of those bode plots with uiinspect and figure2dot and post the results when I have a computer in front of me. Then we can document them once and for all.

okomarov:
Anotehr pointer, that links to my previous comment is this commit https://github.com/okomarov/matlab2tikz/commit/92c6b269869f899f440eefac823ac07eed6a4e7c

egeerardyn:
So, for a `stepplot` such as in #855, it looks like this:
stepplot
screen shot 2016-02-04 at 18 31 36

WDuck:
```
%
%The latest updates can be retrieved from
%  http://www.mathworks.com/matlabcentral/fileexchange/22022-matlab2tikz-matlab2tikz
%where you can also make suggestions and rate matlab2tikz.
%
\definecolor{mycolor1}{rgb}{0.00000,0.44700,0.74100}%
%
\begin{tikzpicture}

\begin{axis}[%
width=0.925\figurewidth,
height=0.942\figureheight,
at={(0.026\figurewidth,0\figureheight)},
scale only axis,
separate axis lines,
every outer x axis line/.append style={white!40!black},
every x tick label/.append style={font=\color{white!40!black}},
xmin=0,
xmax=9,
every outer y axis line/.append style={white!40!black},
every y tick label/.append style={font=\color{white!40!black}},
ymin=0,
ymax=1,
axis background/.style={fill=white},
legend style={at={(0.97,0.03)},anchor=south east,legend cell align=left,align=left,draw=white!15!black}
]
\addplot [color=mycolor1,solid]
  table[row sep=crcr]{%
0	0\\
0.0460517018598701	0.0450074139785546\\
0.0921034037197419	0.0879891606440708\\
0.138155105579612	0.129036410043893\\
0.184206807439482	0.168236228897294\\
0.230258509299354	0.205671765275678\\
0.276310211159224	0.24142242497077\\
0.322361913019096	0.275564039924959\\
0.368413614878966	0.308169029081007\\
0.414465316738836	0.339306551992344\\
0.460517018598708	0.369042655519742\\
0.506568720458578	0.397440413925574\\
0.552620422318448	0.424560062662772\\
0.59867212417832	0.450459126142302\\
0.64472382603819	0.475192539750152\\
0.69077552789806	0.498812766372652\\
0.736827229757932	0.521369907677283\\
0.782878931617802	0.542911810385045\\
0.828930633477672	0.563484167759754\\
0.874982335337544	0.583130616529584\\
0.921034037197414	0.60189282944642\\
0.967085739057284	0.619810603679356\\
1.01313744091716	0.636921945229817\\
1.05918914277703	0.653263149547387\\
1.1052408446369	0.668868878517328\\
1.15129254649677	0.683772233983081\\
1.19734424835664	0.698004827959718\\
1.24339595021651	0.711596849687259\\
1.28944765207638	0.724577129666104\\
1.33549935393625	0.736973200810382\\
1.38155105579612	0.748811356848964\\
1.42760275765599	0.760116708097975\\
1.47365445951586	0.770913234723148\\
1.51970616137573	0.781223837604971\\
1.5657578632356	0.791070386914523\\
1.61180956509548	0.800473768503041\\
1.65786126695535	0.809453928203604\\
1.74996467067509	0.826219917124995\\
1.84206807439483	0.841510680753823\\
1.93417147811457	0.855456022925345\\
2.02627488183431	0.868174326144299\\
2.11837828555405	0.879773556538202\\
2.21048168927379	0.890352180385626\\
2.30258509299354	0.899999999999949\\
2.39468849671328	0.908798916064359\\
2.48679190043302	0.916823622889686\\
2.57889530415276	0.924142242497037\\
2.6709987078725	0.930816902908065\\
2.76310211159224	0.936904265551942\\
2.90125721717185	0.945045912614201\\
3.03941232275147	0.952136990767704\\
3.17756742833108	0.958313061652936\\
3.31572253391069	0.963692194522963\\
3.4538776394903	0.96837722339829\\
3.63808444692978	0.973697320081024\\
3.82229125436927	0.978122383760486\\
4.00649806180875	0.981802991413883\\
4.2367565711081	0.985545602292527\\
4.46701508040746	0.988518463785018\\
4.74332529156668	0.991290364100429\\
5.06568720458578	0.99369042655519\\
5.43410081946474	0.995634841677592\\
5.89461783806345	0.997245771296656\\
6.49328996224177	0.998486438751561\\
7.32222059571944	0.99933930655199\\
8.65771994965569	0.999826219917123\\
9.02613356453466	0.999879773556536\\
};
\addlegendentry{Test Function};

\addplot [color=black,dotted]
  table[row sep=crcr]{%
-0.9	1\\
9.89999999999998	1\\
};
\end{axis}

\begin{axis}[%
width=0.951\figurewidth,
height=\figureheight,
at={(0\figurewidth,0\figureheight)},
scale only axis,
xmin=0,
xmax=1,
xtick={\empty},
xlabel={Time (seconds)},
ymin=0,
ymax=1,
ytick={\empty},
ylabel={Amplitude},
hide axis,
title style={font=\bfseries},
title={Step Response},
axis x line*=bottom,
axis y line*=left
]
\end{axis}
\end{tikzpicture}%
```
Is the result with the Matlab example from #855.
Seems to be a fix. =) 
Also no Latex errors!
Thankyou!

Kind regards,
Philipp

miscco:
As @egeerardyn pointed out, it is rather a hack for that single case than a fix.

WDuck:
@miscco Yea, another isue is hat the xlabel and ylabel are getting stuck in that axe at the bottom of the file and so do not appear in the pdf file. Seems that I have to get deeper into it myself. Is there a flowchart or something for an easier entry into the code?

okomarov:
I'd start from the point in the code I gave with the commit (again if my memory is not failing me), and try to create an exception to that workflow that handles special plots, like bode, steplot etc... Be warned, it's not an easy entry point for contributing to m2t.

I'd probably _consider_ intervening on the figure itself, in order to adapt it to the general case, i.e. linking the legend to the main axis with the lines, rather than to the empty layer. Although against the general principle of not touching the .fig, I fear that adapting the current legend design, would make it even more convoluted than it already is (because of backward and octave compatibility).  

@egeerardyn It would be useful if fig2dot, for legends, reported the referenced axis

egeerardyn:
@okomarov Personally, I'm against changing the data structure of the figures, since I regard those figures as "read-only" variables. That simplifies testing, as that basic assumption assures that
```matlab
plot(...)
matlab2tikz('a.tex')
matlab2tikz('b.tex')
```
produces the same output in `a.tex` and `b.tex`. If we do change the internal data structures of a figure, that is no longer guaranteed and we would need to test such behavior for every test case as well.

Note, there are some locations where we already change the internal data structure, but those changes are reverted before we return from those functions. For the legend, I fear that it will be as much work to revert any simplifying changes and to handle the special cases.

I guess you have a more intimate knowledge of the legend handling, but I would presume that there should be a good way to link plot objects and legend entries in a straightforward way, i.e.
 - either the legend contains a link to an object, or
 - the legend contains a link to the parent of the object (e.g. a group), or
 - recursion over both rules above.
That should be something that we can code in a concise manner without many exceptions. I mean, functions such as [`clickableLegend`](http://www.mathworks.com/matlabcentral/fileexchange/21799-clickablelegend-interactive-highlighting-of-data-in-figures) are able to link those together, even for bode plots, so we should be able to do the same.

For `figure2dot`, it may indeed prove useful to also include more detailed legend information in there. Good point! I will have a look at it over the weekend (`figure2dot` is not that complex).

egeerardyn:
I had a quick look at how `clickableLegend` does things and it seems to use features provided by `legend` itself to determine the links. According to the help of [`legend`](http://nl.mathworks.com/help/matlab/ref/legend.html):

> `[LEGH,OBJH,OUTH,OUTM] = legend(...)` returns 
>   - a handle `LEGH` to the legend axes; 
>   - a vector `OBJH` containing handles for the text, lines, and patches in the legend; 
>   - a vector `OUTH` of handles to the lines and patches in the plot; and 
>   - a cell array `OUTM` containing the text in the legend.

miscco:
So it seems this is an obviously wrong approach. As it is referenced with the respective issues I will close it.

PR #858, opened 2016-02-05
figure2dot: also show "Axes" property

egeerardyn:
@okomarov What you wanted, I hope :smile: 
This shows the related Axes for legend objects.
Confirmed working in both HG1 and HG2 (don't know about Octave).

E.g. for 
```matlab
hold all
plot(1:10)
plot(2:11)
plot(3:12)
legend('a','b')
```

untitled 3

this produces the following DOT figure in HG2 (HG1 also shows the property, but the overall data structure is more complicated). The green highlight is what this PR adds:

### DIFF
--- a/src/figure2dot.m
+++ b/src/figure2dot.m
@@ -84,6 +84,7 @@ function figure2dot(filename, varargin)
             end
             label = addProperty(label, 'Handle', sprintf('%g', double(h)));
             label = addHGProperty(label, h, 'Title', '');
+            label = addHGProperty(label, h, 'Axes', []);
             label = addHGProperty(label, h, 'String', '');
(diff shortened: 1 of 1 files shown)


okomarov:
Thanks! 
Minor thing, the numeric handle for Axes has one digiti less of precision.

egeerardyn:
The precision is set decided automatically. Also, I have the impression there is some structure to the handle values, but I don't know whether the last few digits can actually differ. I have the impression it doesn't matter that much (unless we have figures with hundreds/thousands of objects).

okomarov:
I think so too, good to merge :+1:

egeerardyn:
Thanks for reviewing!

Issue #837, opened 2016-01-01
Systematic approach to track what is supported


egeerardyn:
Right now, we don't really have an overview of what features of HG graphics we do and do not support, e.g. do we translate the `XTickLabelRotation` of an axes or not (and in which versions of Octave/MATLAB do we handle it).

So I'm imagining constructing some sets of tables:
 - which object types (axes, figure, line, ...) we currently handle for which environment (HG1, HG2, Octave)
 - which properties (XLabel, Title, ...) are present for each object type and in what environments we translate them.
 - which specific plot calls (e.g. `histogram`, `hist`, `bode`, ...) we do support in which environment

I think there are a few levels of support we can offer:
 - :white_check_mark: full : _(almost) nothing more to be done_
 - :baby_symbol: partial : _workable in some cases, but some rough edges_
 - :heavy_exclamation_mark: todo : _not implemented yet_
 - :x: none : _will not implement, implementation is impossible (e.g. `light` objects)_
 - :heavy_minus_sign: n/a: _a certain property/object is not present in that environment (e.g. `XTickLabelRotation` in HG1/Octave)
 - unknown: _we haven't tracked down what level of support we have_

So what I have in mind are tables that could look like:

| Property of `axes`         | HG1                | HG2                      | Octave             |
|----------------------------|--------------------|--------------------------|--------------------|
| `{X,Y,Z}Label`             | :white_check_mark: | :white_check_mark:       | :white_check_mark: |
| `{X,Y,Z}TickLabelRotation` | :heavy_minus_sign: | :heavy_exclamation_mark: | :heavy_minus_sign: |
| `UserData`                 | :x:                | :x:                      | :x:                |

PR #860, opened 2016-02-07
Fix scatter: markerEdgeColor unknown + some cleanup

egeerardyn:
This should fix something that was reported on FEX:

> **Comment from FEX: 07 Feb 2016 by Raph**
> Using scatter() I get the error
>
>> Undefined function or variable 'markerEdgeColor'. 
>> ```
>> Error in matlab2tikz>getScatterOptsColormap (line 3237) 
>> [m2t, ecolor] = getColor(m2t, h, markerEdgeColor,'patch');
>> ```
>
> Adding the line 
> ```matlab
> markerEdgeColor = get(h, 'MarkerEdgeColor'); 
> ```
> helps, maybe you can include this or provide a better fix.

Instead of the perfectly good fix suggested above, I took the opportunity to move some variables into a struct such that other functions can take fewer parameters. The introduced functions do not have any other call sites (according to MATLAB highlighting).

Note that the surrounding code could use some TLC, e.g. the other place where we do a `translateMarker` could be simplified together with this. But that's for another day and PR.


src/matlab2tikz.m
```diff
@@ -3514,14 +3514,7 @@ function warnFacetedInterp(m2t)
...
 
-    matlabMarker = get(h, 'Marker');
-    markerFaceColor = get(h, 'MarkerFaceColor');
-    markerEdgeColor = get(h, 'MarkerEdgeColor');
-    hasFaceColor = ~isNone(markerFaceColor);
-    hasEdgeColor = ~isNone(markerEdgeColor);
-    markOptions = opts_new();
-    [tikzMarker, markOptions] = translateMarker(m2t, matlabMarker, ...
-        markOptions, hasFaceColor);
+    rawMarker = getMarkerStruct(m2t, h);
```
miscco: rawMarker seems a bad name, how about markerOptions or markerOpts?

src/matlab2tikz.m
```diff
@@ -3530,18 +3523,17 @@ function warnFacetedInterp(m2t)
...
-    sData = translateMarkerSize(m2t, matlabMarker, sqrt(sData)/2);
+    sData = translateMarkerSize(m2t, rawMarker.style, sqrt(sData)/2);
 
     drawOptions = opts_new();
     if length(cData) == 3
         [m2t, drawOptions] = getScatterOptsOneColor(m2t, h, drawOptions, ...
-                                                markOptions, tikzMarker, ...
-                                                cData, sData, constMarkerkSize);
+                                                rawMarker, cData, sData, ...
+                                                constMarkerkSize);
```
miscco: Is constMarkerSize necessary here, as it is false all the time?

src/matlab2tikz.m
egeerardyn: I fully agree that `rawMarker` is not optimal. 
However, we typically use `*Options` and `*Opts` for data that is structured in a particular way rather than such a struct, so I think those are not any better.

miscco:
I love this, as i thought about doing something similar for other types as well.
I think i generallized it for barplot, but lines could also benefit forom something like that

src/matlab2tikz.m
egeerardyn: I don't think it's constant all the time: `sData` could be a vector.
But you do have a point that it could be calculated from `sData` again.

src/matlab2tikz.m
miscco: Yeah, i couldnt think of a better one. Maybe we could go with markerStruct and add similar one for lineStruct, pathcStruct etc?

egeerardyn: How about `markerInfo` or just `marker`?

src/matlab2tikz.m
miscco: Given
```
constMarkerkSize = length(sData) == 1;
```
I think it is alwase false in this branch.  getScatterOptsColormap handles the vector case

src/matlab2tikz.m
miscco: Info sounds great :+1:

egeerardyn:
@miscco Thanks! Indeed, in applying these changes, I think we can declutter our function workspaces quite a bit and for some functions we really need to limit the number of inputs/outputs being passed around.
I think this is something that would translate very well into an object-oriented framework, if we take care to think the changes over a bit.

miscco:
Yeah, the main benefit is also convergence. How often do we generate a table- or a lineplot with early identical options. The same for some cases, where FaceColor is missing the other without edgeColor etc

src/matlab2tikz.m
egeerardyn: Note that `sData` is the size data, not the color data (`cData`). By looking into `getScatterOptsOneColor`, you see the value is actually being used.

miscco:
So I had some nitpicks. I don't really care too much about passing the struct in the one occurrence, but I am heavily in favour of pulling any opts_print into its own line.

once that is addressed you get my :+1:

miscco:
Ah ok, I take that back somewhat.

We have a function
```
[m2t, drawOptions] = getMarkerOptions(m2t, h)
```

I would think, there is a ton of duplication going on here. So maybe incorporate getMarkerInfo there?

For future reference, what is your opinion on the preferable approach.

have get*Info to return a struct that is then processed into drawOpts by get*Options, or put everything into one function. I am definitely in favour of splitting into two.

egeerardyn:
I think that doing this `get*Info` for everything may declutter the code quite a bit, but we should not do this blindly for everything. Probably we will stumble upon quite some duplication while making those changes.

miscco:
Yeah, i think the important part will be to make sure, that one catches all the corner cases with getOrDefault etc. 

But in general i think this might prevent a lot of subtle bugs, that might happen otherwise.

As a second point, what do you think about adding the booleans too aka:
```
faceColor
hasFaceColor
```

Therefore, one could also have all the checks like
```matlab
if *Info.hasEdgeColor
...
end
```

egeerardyn:
@miscco: I would suggest we try to keep the pattern as I did here (or extensions thereof): so `*Info.has*` is a good idea. The only downside is that for some objects, e.g. `MarkerFaceColor` might be `auto` which is handled differently and so on. This is something we would need to keep track off in a good way (i.e. #837).

But having uniform access to those properties is certainly something that would benefit uniformity of both the code and the features we support for different HG objects.

src/matlab2tikz.m
```diff
@@ -3569,8 +3561,20 @@ function warnFacetedInterp(m2t)
...
+function marker = getMarkerInfo(m2t, h)
+    % gets marker-related options as a struct
+    marker                        = struct();
+    marker.style                  = get(h, 'Marker');
+    marker.FaceColor              = get(h, 'MarkerFaceColor');
+    marker.EdgeColor              = get(h, 'MarkerEdgeColor');
+    marker.hasFaceColor           = ~isNone(marker.FaceColor);
+    marker.hasEdgeColor           = ~isNone(marker.EdgeColor);
+    [marker.tikz, marker.options] = translateMarker(m2t, marker.style, ...
+        opts_new(), marker.hasFaceColor);
```
miscco: Again the tiniest of nitpicks, can you put marker.options = opts_new();  on a single line and then pass marker.options into translateMarker?

I am just against passing functions as arguments

miscco:
@egeerardyn Yeah, stuff like 'auto' and friends can make the life harder, but that could bealleviated by something like *Info.isDefault*, which is often the

miscco:
Sorry, wrong button :(

src/matlab2tikz.m
egeerardyn: Technically, this is not passing a function as an argument. And especially with `opts_*`, you should look at those as objects in an early stage where `opts_new()` is something like `cell()`, `''`, ...

But I think I have an idea to accommodate to your comment: passing that in as a defaulted value should allow to make the code cleaner at the other call site.

miscco: Yeah, but generally we have the pattern 
```
[m2t, *Opt] = Func(m2t, h, *Opt, Arg)
```
Which we should stick to, or as you said put it inside the function as below.
```
[m2t, *Opt] = Func(m2t, h, Arg)
```

egeerardyn: I will upload a new version shortly (but I have messed up some hashes, so I will try to fix those first). The new syntax will follow the pattern `[opts] = getMarkerInfo(m2t, h, opts)`

src/matlab2tikz.m
```diff
@@ -6329,6 +6327,14 @@ function printAll(m2t, env, fid)
     end
     opts = opts_append(opts, key, value);
 end
+function opts = opts_addOpts(m2t, opts, key, subOpts, sep)
```
miscco: So you will hate me for that, but can we rename it to
```
opts_addSubOpts
```
Because this is generally a suboption of some axis option. So we can differentiate, when we have curly braces around

Also i cannot think of any case where we would not use ',' as a separator.

src/matlab2tikz.m
```diff
@@ -6329,6 +6327,14 @@ function printAll(m2t, env, fid)
...
+    % add a key={Opts} pair to an options array
+    if ~exist('sep','var') || ~ischar(sep)
+        sep = ',';
```
miscco: Please use ', ' with a whitespace to improve readability of the tikz code

src/matlab2tikz.m
egeerardyn: I thought the same (except for the `opts_addSubOpts`), but the name change is fine with me. Not passing `sep` has been implemented in the meanwhile.

src/matlab2tikz.m
egeerardyn: I absolutely agree with you that `', '` is more readable (and I prefer that as well, although I also quite like the `sprintf(',\n')` syntax in some cases).
However, I'm a bit reluctant to change this since that will change the hashes right now. I'd rather keep it for another PR such that this PR remains purely:
 - refactoring, and
 - a bug fix
without any other side effects.

miscco:
:+1: for keeping the separator as ','. Though we should change that in a future PR.

miscco:
All in all a very nice PR. Loving it!

PeterPablo:
:+1: As far as I can see. (Not runtime tested, but Jenkins and Travis are happy).

egeerardyn:
Thanks for reviewing this, guys!

Issue #859, opened 2016-02-07
Scatter plots: changing color and size


egeerardyn:
**This has been requested on FEX: 07 Feb 2016, by Raph:**
> When the colors of the scatters are set by a colormap, the scaling of the scatter markers is also not working. When all scatters are having the same color it works fine.

The idea is indeed simple, but currently we are using `meta` data to change the size and the color. As such, changing one of them also would set the other (or produce weird results).
Given that `pgfplots` has quite some customization options, I think this is something that should really be feasible. The coding is probably heavier on the TikZ side than on the MATLAB side.

miscco:
Can you give a MWE?

egeerardyn:
I don't have an MWE unfortunately.

okomarov:
I suspect is again about using cdata in scatter, which we do not support.

egeerardyn:
@okomarov I'm not sure (I don't use [scatter](http://nl.mathworks.com/help/matlab/ref/scatter.html?searchHighlight=scatter) myself), but I think this may already happen if one provides both changing sizes and changing colors, i.e.

```matlab
scatter(x,y,sData,cData)
``` 

where both `sData` and `cData` are vectors.
Otherwise, indeed this would be a duplicate of #433.

rengesser:
Hi, thanks for raising this issue. It was me on matlab central asking for it. A MWE is here:

x=1:10;
y=rand(size(x));
z=cos(x);
subplot(1,2,1)
scatter(x,y,400,z','filled') % scatter with changing color
subplot(1,2,2)
scatter(x,y,400,'k','filled') % only black scatter
matlab2tikz('mwe_scatter.tex','standalone',true)
saveas(gcf,'mwe_scatter_mat_fig.png','png')

The scatter size is not changing (see attached png file), but in tikz the scatter size for the colored scatters are much smaller (see attached pdf file)

![mwe_scatter_matlab](https://cloud.githubusercontent.com/assets/17144266/12920853/b3d1cefa-cf4a-11e5-97f6-35fa81c18de0.png)
[mwe_scatter_tikz.pdf](https://github.com/matlab2tikz/matlab2tikz/files/123253/mwe_scatter_tikz.pdf)

egeerardyn:
@rengesser Thank you for providing an MWE! I hope you don't mind I have taken the liberty to mark it up and placed a PNG of the TikZ output for convenience.

[...]

PR #861, opened 2016-02-09
Quick fix for #859

egeerardyn:
This is just a quick and "dirty" fix for #859.

This is a work-in-progress (I don't want any additional TODOs in this PR).

### What needs to be done:
I think we can clean up the scatter plot stuff a lot by breaking down the different settings into different functions instead of how everything is organized right now (i.e. 1 color vs N colors? 1 size vs N sizes?). That produces a mess to make sure that every property is handled since a property is handled in a few different locations.

I imagine that we can simplify everything by starting from a basic table output for the data that we want to generate:
```
x y [z] [r] [g] [b] [colorValue] [size]
```
where `[*]` denotes an optional value. Optional in the sense that it can be omitted when it is constant over the whole dataset or when it is not relevant.

Generating such a table is not that hard, the only special thing is to keep track of which columns are (not) relevant.

Once the table is taken care of, we can think about telling to `pgfplots`:
- pick the right marker symbol, and so on,
- use the column `size` for sizes (or just use the default value)
- use the columns `r`, `g` and `b` for the color (for a CData matrix)
- use the column `v` to use the raw colormap

Each of those bullets can be a separate function, which is well-contained and has a single responsibility.


miscco:
So are there any TODOs here?

I think the current state is already quite an improvement so i would merge it rather sooner than later or it will just bitrott.

egeerardyn:
@miscco You're right. I'm currently performing the backwards merge and will update this PR. We can tackle the other TODOs after release 1.1.

src/matlab2tikz.m
```diff
@@ -3543,60 +3543,144 @@ function warnFacetedInterp(m2t)
...
+    
+    dataInfo   = getDataInfo(h, 'X','Y','Z','C','Size');
+    markerInfo = getMarkerInfo(m2t, h);
+    
+    if isempty(dataInfo.C) && strcmpi(getEnvironment(), 'Octave')
+        dataInfo.C = get(h, 'MarkerEdgeColor');
     end
 
-    markerInfo = getMarkerInfo(m2t, h);
     %TODO: check against getMarkerOptions() for duplicated code
```
miscco: Is that TODO still applicable?

src/matlab2tikz.m
```diff
@@ -3543,60 +3543,144 @@ function warnFacetedInterp(m2t)
...
 
-    constMarkerkSize = length(sData) == 1; % constant marker size
+    dataInfo.Size = tryToMakeScalar(dataInfo.Size, m2t.tol);
 
     % Rescale marker size (not definitive, follow discussion in #316)
     % Prescale marker size for octave
     if strcmpi(getEnvironment(), 'Octave')
-        sData = sData.^2/2;
+        dataInfo.Size = dataInfo.Size.^2/2;
     end
```
miscco: Is that check only needed for scatter plots or can we put that into getDataInfo?

src/matlab2tikz.m
```diff
@@ -3543,60 +3543,144 @@ function warnFacetedInterp(m2t)
...
+    hasDifferentColors = numel(dataInfo.C)    ~= 3;
+    isaScatter         = hasDifferentSizes || hasDifferentColors;
+    if isaScatter
+        drawOptions = opts_add(drawOptions, 'scatter');
+    end
+    %TODO: we need to set the scatter source
+    drawOptions = opts_add(drawOptions, 'only marks');
+    drawOptions = opts_add(drawOptions, 'mark', markerInfo.tikz);
+
+    if length(dataInfo.C) == 3
```
miscco: I would say that  length(dataInfo.C) ==  numel(dataInfo.C). So we should decide for one

src/matlab2tikz.m
```diff
@@ -3543,60 +3543,144 @@ function warnFacetedInterp(m2t)
...
+        % gets options specific to scatter plots with a single color
+        % No special treatment for the colors or markers are needed.
+        % All markers have the same color.
+        [m2t, xcolor, hasFaceColor] = getColorOfMarkers(m2t, h, 'MarkerFaceColor', dataInfo.C);
+        [m2t, ecolor, hasEdgeColor] = getColorOfMarkers(m2t, h, 'MarkerEdgeColor', dataInfo.C);
+
```
miscco: The indentation below seems wrong or there are some ends missing.

src/matlab2tikz.m
```diff
@@ -3543,60 +3543,144 @@ function warnFacetedInterp(m2t)
...
+                sprintf('%.4fpt', dataInfo.Size)); % FIXME: investigate whether to use `m2t.ff` 
+        else
+            %TODO: warn the user about this. It is not currently supported.
+        end
+
+        drawOptions = opts_add(drawOptions, 'scatter src', 'explicit');
+        drawOptions = opts_addSubOpts(drawOptions, 'scatter/use mapped color', ...
+                                   markerOptions);
+        % Add color map.
+    m2t = m2t_addAxisOption(m2t, matlab2pgfplotsColormap(m2t, m2t.current.colormap), []);
```
miscco: indentation is wrong here

src/matlab2tikz.m
egeerardyn: `length` and `numel` do something slightly different:
 - `length(x) == max(size(x))`
 - `numel(x) == prod(size(x))`

I had the impression it was something else (i.e. that `length` was `y = size(x); y(find(y~=1,1))`: the size of the first non-singleton dimension), so I think here having `length` is indeed dangerous.

src/matlab2tikz.m
egeerardyn: I agree.

src/matlab2tikz.m
egeerardyn: Will check on that.

src/matlab2tikz.m
```diff
@@ -3543,60 +3543,144 @@ function warnFacetedInterp(m2t)
...
     end
     % - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
     % Plot the thing.
-    [env, data, sColumn] = organizeScatterData(m2t, xData, yData, zData, sData);
+    [env, data, metaPart, columns] = organizeScatterData(m2t, dataInfo);
```
miscco: Is that metaPart used anywhere?

src/matlab2tikz.m
```diff
@@ -3606,6 +3690,31 @@ function warnFacetedInterp(m2t)
...
+        dataInfo.(name) = get(h, [name 'Data']);
+    end
+end
+% ==============================================================================
+function value = tryToMakeScalar(value, tolerance)
+    % make a vector into a scalar when all its components are equal
+    if ~exist('tolerance','var')
+        tolerance = 0; % do everything perfectly
+    end
+    if all(abs(value - value(1)) <= tolerance)
```
miscco: There is a maltab function "uniquetol". If that also exists for octave we could use it here.

src/matlab2tikz.m
```diff
@@ -3621,116 +3730,34 @@ function warnFacetedInterp(m2t)
...
+    
+    xData = dataInfo.X;
+    yData = dataInfo.Y;
+    zData = dataInfo.Z;
+    cData = dataInfo.C;
+    sData = dataInfo.Size;
+    
+    % add the actual data
     if ~m2t.axes{end}.is3D
         env = 'addplot';
```
miscco: Could you define 
```
 columns     = {'x','y'};
```
And below the full one, i think this would be clearer.

miscco:
So i had some minor nitpicks. 

i think this can be merged rather soon.

src/matlab2tikz.m
egeerardyn: Has been introduced in R2015a, so this is a bit too new for use to make use of. Also, the goal is slightly different:
 - `uniquetol` returns the unique values up to some tolerance
 - this function checks whether all values are equal (up to some tolerance).

I agree that both can be used to accomplish the same, but we don't actually need that new function for this purpose.

miscco: Ok, i just checked for unique and saw uniquetol. Actually i didnt check when it was introduced

src/matlab2tikz.m
egeerardyn: I think this is quite specific to scatter plots. For scatter plots the `sData` (marker size) is specified proportionally to the area of the marker, for other plots, it's proportional to the dimensions (i.e. `sqrt(area)`) of the marker, so the handling should be somewhat different.

Anyhow, the sizing isn't done perfectly, so I would refrain from moving those things around until we have a further look at #316 where those issues were discussed as well.

src/matlab2tikz.m
egeerardyn: Not yet I think (I was in the middle of refactoring).

src/matlab2tikz.m
egeerardyn: It is still applicable: both functions are nontrivial and way too complex to see immediately whether they do the same or not. I.e. when refactoring those, we should check for duplication and maybe even try to merge them.

miscco: So I would say we leave it in and save the refactoring for after the 1.1 release.

src/matlab2tikz.m
miscco: Im not 100% decided whether we should keep it or just remove it for the 1.1 release as that would be cleaner.

egeerardyn: Sorry, I edited my post. The `metaPart` is being used.

src/matlab2tikz.m
```diff
@@ -3543,60 +3543,144 @@ function warnFacetedInterp(m2t)
...
 
     drawOptions = opts_new();
-    if length(cData) == 3
-        [m2t, drawOptions] = getScatterOptsOneColor(m2t, h, drawOptions, ...
-                                                markerInfo, cData, sData, ...
-                                                constMarkerkSize);
-    elseif size(cData,2) == 3
-        drawOptions = getScatterOptsRGB(m2t, drawOptions);
+    
+    %% Determine if we are drawing an actual scatter plot
```
miscco: Superflous %

egeerardyn: This is by design (MATLAB cell mode).

src/matlab2tikz.m
```diff
@@ -3543,60 +3543,144 @@ function warnFacetedInterp(m2t)
...
+                                         markerInfo.options);
+            if hasEdgeColor
+                markerOptions = opts_add(markerOptions, 'draw', ecolor);
+            else
+                markerOptions = opts_add(markerOptions, 'draw', xcolor);
+            end
+            if hasFaceColor
+                markerOptions = opts_add(markerOptions, 'fill', xcolor);
+            end
+            % for changing marker size, the 'scatter' option has to be added
```
miscco: I dont really know that much about scatter plots, but these two lines below and the if else clause following it seem redundant. What is the difference between

color = xcolor
and 
scatter/use mapped color = xcolor

Could it be that the former should be part of the respective clause below?

src/matlab2tikz.m
miscco: Oh didnt know we use that in m2t

src/matlab2tikz.m
egeerardyn: It could very well be, but if you look in the old code (below), such a check was already present. 

Personally, I'd say that further clean up should happen in #925 to get this PR merged,  then I have a few days to carry out changes to my thesis and hopefully during the weekend, I can address the other PRs before a 1.1 release.

miscco: Second that. Cant wait to get the corrections for my thesis.

src/matlab2tikz.m
egeerardyn: Not at a lot of places yet, but we have to start somewhere 😄

egeerardyn:
@miscco Thank you for the review :+1: if you are happy with the current state, I think this PR can be merged

miscco:
Jepp this is definitely a nice cleanup of the mess scatter plots are.

PR #862, opened 2016-02-10
Simplified getAndCheckDefault

miscco:
The type of the figure is nearly always covered in the provided handle. Therefore instead of passing it explicitly deduce it from the handle.

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -926,24 +926,24 @@ function m2t = drawGridOfAxes(m2t, handle)
     if any(hasGrid)
         gridOpts = opts_new();
         % Get the line style and translate it to pgfplots
-        [gridLS, isDefault] = getAndCheckDefault(...
-            'axes', handle, 'GridLineStyle', ':');
+        [gridLS, isDefault] = getAndCheckDefault(handle, ...
+                                'GridLineStyle', ':');
         if ~isDefault || m2t.cmdOpts.Results.strict
             gridOpts = opts_add(gridOpts, translateLineStyle(gridLS));
         end
 
         % Get the color of the grid and translate it to pgfplots usable
         % values
-        [gridColor, defaultColor] = getAndCheckDefault(...
-            'axes', handle, 'GridColor', [0.15, 0.15, 0.15]);
+        [gridColor, defaultColor] = getAndCheckDefault(handle, ...
+                                      'GridColor', [0.15, 0.15, 0.15]);
         if ~defaultColor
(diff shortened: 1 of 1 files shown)


src/matlab2tikz.m
```diff
@@ -5610,9 +5610,9 @@ function warnFacetedInterp(m2t)
     end
 end
 % ==============================================================================
-function [val, isDefault] = getAndCheckDefault(type, handle, key, default)
+function [val, isDefault] = getAndCheckDefault(handle, key, default)
     % gets the value from a handle of certain type and check the default values
-    default   = getFactoryOrDefault(type, key, default);
+    default   = getFactoryOrDefault(handle.Type, key, default);
```
PeterPablo: This fails in octave and MATLAB HG1. You might get this working using `get(handle, 'Type')` (untested).

miscco: Yepp, just pushed that change, thx

miscco:
So it seems the culprit here is getLegendOpts, which now produces an additional fill=white.

I am unsure, whether this is incorrect, as the results look the same

miscco:
Yeah, does someone has the option to debug this for matlab2014a. Somehow it does not accept [1 1 1] as default, although that is the default in matlab2014a.

egeerardyn:
@miscco: my guess is that in R2014a, there was no real object "legend", so perhaps the defaults from "axes" are taken (which is what "legend" produces back then)?

miscco:
@egeerardyn The strange thing is, that the factory default FactoryLegendColor exists and we provide white as a default fallback. Strange thing.

Sadly as i do not have Matlab2014a and investigating this headless on you server is tedious at best, I am not sure how to proceed.

egeerardyn:
@miscco Normally, you can use X11-forwarding to get the MATLAB GUI from the server onto your computer (it's probably quite slow, but it's better than nothing). If you're using Windows, you can try Xming as X11 server, it works well (or at least it did when I last used it, a few years ago).

okomarov:
After Egon's remark, I do not think that omitting the type in `getAndCheckDfl()` is much of a gain. To fully automate it, we would need to implement the type detection which is already carried out in several stages while traversing the figure with m2t. This inference is not trivial and re-introducing the complexity of type detection in a low level utility (which at that point should be of basic complexity) is like re-doing some of the work unnecessarily. I mean, it is not as bad as it sounds, but unfortunately this is the price we pay for backward compatibility.

miscco:
Given the trouble this does on older matlab versions I will close this PR. Linewrapping is not that bad of a thing compared to the headache all this special cases mace

Issue #863, opened 2016-02-15
Suggestion: Cleaning up stairs plots


bobpepin:
Hi,
I found the following useful to clean up a stairs plot where the function is constant over some interval. It replaces for example `[(x_1, y_1), (x_2, y_1), ... (x_k, y_1), (x_{k+1} y_2)]` by `[(x_1, y_1), (x_{k+1} y_2)]`.

```matlab
% h is a handle to a 'stairs' object
I = [true (diff(h.YData) > 0)];
set(h, 'XData', h.XData(I), 'YData', h.YData(I));
```

Regards,
Bob

egeerardyn:
Thank you for this suggestion! We will do our best to integrate it into `cleanfigure`

miscco:
As far as I know, the lineSimplification in the development branch should do exactly that. Have you tried it?

okomarov:
@miscco stair plots are not currently simplified (line 130 of cleanfigure)

miscco:
Is there any reason for that?`Just add ing 
```
% Simplify the lines by removing superflous points
simplifyLine(meta, h, cmdOpts.targetResolution);
```
should do the trick, or are there weird handle types involved?

okomarov:
Probably HG1 vs HG2 compatibility (type detection and potential changes in the graphical object) + Octave differences. Also, this feature has been fairly low-prioprity, since stairs plots need a very limited amount of data (unless @bobpepin has a specific use case he wants to share), hence very unlikely to generate out-of-memory in latex.

egeerardyn:
I agree with @okomarov that it is quite low-priority (IMHO since few people use `stair` plots), but I also think this is a quick fix (there are some things we could tweak, but not a lot: `==` instead of `<` and maybe also check the other direction). I especially like that it's just a really simple algorithm and it is lossless compression (unlike `simplifyLine` which can be lossy). So in my book, this is a no-brainer: we should add it.

Since `stairs` and `plot` use a different strategy to interpolate between each pair of points, `simplifyLine` is not applicable.
In the example below, `simplifyLine` should remove all points except the first and the last. 
For `line` that is desired behavior. For `stairs` it is not.

```matlab
x = 1:10
y = 2*x
figure, hold all
stairs(x,y)
plot(x,y)
legend('stairs','line')
```
[...]

PR #865, opened 2016-02-17
Added a simplification algorithm for stair plots

miscco:
This is the simplification function proposed in #863.

I have also modified ACID(28) to add a use case for this simplification algorithm.

### DIFF
--- a/src/cleanfigure.m
+++ b/src/cleanfigure.m
@@ -129,7 +129,9 @@ function recursiveCleanup(meta, h, cmdOpts)
             limitPrecision(meta, h, cmdOpts.scalePrecision);
         elseif strcmpi(type, 'stair')
             % Remove data points outside of the visible axes
-            pruneOutsideBox(meta, h);
+            pruneOutsideBox(meta, h);            
+            % Remove superfluous data points
+            simplifyStairs(meta, h);
             % Limit the precision of the output
             limitPrecision(meta, h, cmdOpts.scalePrecision);
         elseif strcmp(type, 'text') && cmdOpts.pruneText
@@ -446,6 +448,31 @@ function simplifyLine(meta, handle, targetResolution)
     end
 
     % Remove the data points
+    removeData(meta, handle, id_remove);
+end
+% =========================================================================
+function simplifyStairs(meta, handle)
+   % Some data might not lead to a new step in the stair, e.g.
+   % [(x_1, y_1), (x_2, y_1), ... (x_k, y_1), (x_{k+1} y_2)] 
+   % equals [(x_1, y_1), (x_{k+1} y_2)]
+
+    % Extract the data
+    xData = get(handle, 'XData');
+    yData = get(handle, 'YData');
+
+    % Dont do anything if the data is empty
+    if isempty(xData) || isempty(yData)
+        return;
+    end
+
+    % Check which point do lead to a new step
+    xStep = [true, (diff(xData) ~= 0)];
+    yStep = [true, (diff(yData) ~= 0)];
+
+    % Plot only points, that generate a new step
+    id_remove = find(~xStep | ~yStep);
+
+    % Remove the superfluous data
     removeData(meta, handle, id_remove);
 end
 % =========================================================================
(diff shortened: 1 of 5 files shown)


miscco:
So after the discussion in #863 i added a check for monotonicity

src/cleanfigure.m
```diff
@@ -449,6 +451,47 @@ function simplifyLine(meta, handle, targetResolution)
...
+function simplifyStairs(meta, handle)
+   % Some data might not lead to a new step in the stair, e.g.
+   % [(x_1, y_1), (x_2, y_1), ... (x_k, y_1), (x_{k+1} y_2)] 
+   % equals [(x_1, y_1), (x_{k+1} y_2)]
+
+    % Extract the data
+    xData = get(handle, 'XData');
+    yData = get(handle, 'YData');
+
+    % Dont do anything if the data is empty
```
egeerardyn: Nitpick: Dont -> Don't

src/cleanfigure.m
```diff
@@ -449,6 +451,47 @@ function simplifyLine(meta, handle, targetResolution)
...
+    if isempty(xData) || isempty(yData)
+        return;
+    end
+
+    % Check for identic data points
```
egeerardyn: Nitpick: identic -> identical (or coinciding, repeated, ...)

src/cleanfigure.m
```diff
@@ -449,6 +451,47 @@ function simplifyLine(meta, handle, targetResolution)
...
+
+    % Do not remove the last data point
+    xNoDiff(end) = false;
+    yNoDiff(end) = false;
+
+    % Identic data might lead to a visible result if the sign in the other
+    % dimension changes , e.g. the sequence  [0, 1], [0, -1], [0, 2]
+    % Therefore check for a sign change in the difference between data
+    % points (The sign changes if diff(sign)~=0)
+    xIsMonotone  = [true, diff(sign(diff(xData)))==0, true];
```
egeerardyn: I think we better extract this to a function.
Moreover, I have my doubts about the correctness (either of the code or my understanding of this)

As I remember, we need to check that the whole sequence is monotonous: `x(i) <= x(j) if i < j for each i != j` (or `x(i) >= x(j) ... `)
And this function doesn't seem to do that, it rather checks for locally strict monotonicity, i.e. `x(i) < x(i+1) for each i` (or `x(i) > x(i) ...`).

Perhaps we should add some documentation in the code, e.g. some ASCII plots or data sequences which highlights what kind of points can be removed (somewhere before we actually compute anything).
```matlab
% E.g. in the plot below, the data points are indicated by `o` and `x`. We can remove the `x` points without altering the lines. So this means, either:
%  - one of the coordinates is not changing
%  - ...
%
%       x--o--+
%       |     |
% o--x--o     o--o
%
```

miscco: I dont think monotony is required over the full plot. The only thing that matters is when the monotony between to consequtive pairs of points changes. thats when additional lines might occur. 

Consider your expanded example
```
%       o--x--o
%       |     |
% o--x--o     o--x--o
%       |
%       o
```
This plot is not globally monoton, but that does not mean we cannot remove data points.

egeerardyn: I agree that we don't need global monotonicity to able to simplify the graph.
Sorry, I had a long day yesterday, so I think my reasoning skills were not ideal.

miscco:
So are there any further comments on this or can i update the hashes?

egeerardyn:
If the before/after plots look the same, you can certainly update the hashes :+1: I have no further comments (except: well done!)

miscco:
So the diff shows that only the additional stair i introduced in the test is added.

Whoever, what i cannot grasp is why for Matlab2014a and Octave it seems to add the not simplified stair. I checked my octave 4.0 version and it recognizes the stair plot as a line plot, so that lineSimplify should take care of it. Sadly i cannot really test it. Still it is an obvious improvement for all other cases.

 [diff_simplifyStairs.zip](https://github.com/matlab2tikz/matlab2tikz/files/154457/diff_simplifyStairs.zip)

miscco:
So given that it works and there are no further comments i would merge this tomorrow.

Issue #867, opened 2016-02-24
Contour plots: line width + color is off


rengesser:
I have some problems with exporting contour plots. 

- [x] Changing the LineWidth in contour plots is ignored by matlab2tikz:
```matlab
[X Y Z]=peaks;
subplot(1,2,1); contour(X,Y,Z,'LineWidth',1)
subplot(1,2,2); contour(X,Y,Z,'LineWidth',5)
matlab2tikz('contour_test.tex','standalone',true)
``` 
<img width="1127" alt="contour_linewidth_mwe" src="https://cloud.githubusercontent.com/assets/17144266/13286591/b2ba8386-db02-11e5-92db-e111f77a6314.png">
In matlab the two plots are having different line widths, but in matlab2tikz they don't. Adding the option "line width=5pt" to the addplot command in pgfplots would help.

- [x] A second problem for me is, that the color of a single contour line is ignored:
```matlab
[X Y Z]=peaks;
contour(X,Y,Z,[0 0],'r') % only one red colored contour line at level 0
matlab2tikz('contour_test2.tex','standalone',true)
``` 
<img width="1014" alt="contour_color_mwe" src="https://cloud.githubusercontent.com/assets/17144266/13286894/59b70aaa-db04-11e5-8d52-9ce64704db2d.png">
In matlab the line is red, in pgfplots it is black or blue.

miscco:
We started doing some improvements to the options gathering , e.g. get*Info, which should help us catching those cases too. However i am not sure how fast we can tackle this.

miscco:
Please try  #872, which should fix this issue.

rengesser:
Works perfect for me, thanks a lot.

miscco:
As the fix is merged i am closing this.

PR #872, opened 2016-03-01
Consider line options in contour plots

### LINKED ISSUE #795: LineSpec for contour lines
**Imported from FileExchange (21 Oct 2015 Sebastian Bomberg)**

I'm facing a little issue regarding iso contours though. 
Apparently, the line spec of couture lines is ignored.

```matlab
Z = peaks();
subplot(1,2,1)
hPC = pcolor(Z);

hPC.FaceColor = 'interp'; 
hPC.EdgeColor = 'interp'; 
hPC.LineStyle = 'none';

hold all;
contour(Z, '-k', 'ShowText', 'on');

hold off;

subplot(1,2,2)
contour(Z, '-k', 'ShowText', 'on');
matlab2tikz('pcolorAndContour.tex','standalone',true);

```
counter lines should appear black but get drawn with the same colormap as the patches from pcolor. Thus you can't see them. It becomes more clear in the 2nd subplot.

I'd really appreciate your help.

Sebastian

miscco:
So this a rather clean approach on #867

I hope it also fixes #795 but did not check yet.


miscco:
So as far as I can see the difference here is that we now do not clutter a "solid" every time we draw a line, which I would consider an improvement. 

Here are the diffs of the tex files:
[diff_ContourLines.zip](https://github.com/matlab2tikz/matlab2tikz/files/152449/diff_ContourLines.zip)

okomarov:
It is not easy to check the diff_zip. I am saying that because it would be easier to look at graphical differences first and in case of a clean PR like this, it means that the merge can be accomplished merge faster.

miscco:
@okomarov  Thats true, however it is more difficult to actually compare the images for visual difference. We tried that once and it is really tricky.

in contrast the diff is rather straigthforward.

okomarov:
Well, we have the runTestGraphical with side by side comparison. If there is no perceptible difference, in a sense that is all that matters. After all this is a visual conversion utility. The diff in the code matters to the developers, but hardly to the end user.

miscco:
Here is the pdf with the differences between develop branch an the PR
[acid_Contour.zip](https://github.com/matlab2tikz/matlab2tikz/files/153000/acid_Contour.zip)

I cannot spot any difference, although lack of coffee seems possible

src/matlab2tikz.m
```diff
@@ -1872,10 +1872,12 @@ function assertRegularAxes(handle)
     lineOpts = opts_new();
 
     % Get the options from the handle
-    lineStyle = get(h, 'LineStyle');
     lineWidth = get(h, 'LineWidth');
 
-    if ~isNone(lineStyle) && (lineWidth > m2t.tol)
+    % Get the line style and check whether it is the default one
+    [lineStyle, isDefaultLS] = getAndCheckDefault('Line', h, 'LineStyle', '-');
```
okomarov: This is for some reason returning false and causing many tests to fail because it adds the `solid` option:

```
test15-converted.tex
59c59
< \addplot [color=mycolor1,solid,forget plot]
---
> \addplot [color=mycolor1,forget plot]
```

src/matlab2tikz.m
```diff
@@ -2600,6 +2602,18 @@ function assertRegularAxes(handle)
             plotOptions = opts_add(plotOptions,'contour/labels','false');
         end
 
+        % Get line properties
+        [m2t, lineOptions] = getLineOptions(m2t, h);
```
okomarov: Actually ignore my previous comment that I deleted, it is this line that adds unconditionally the `solid`:

```
[~,lineOpt] = getLineOptions(m2t,h)
lineOpt = 
    'solid'    ''
```

We should probably test for default here too?

miscco: I cant follow you where is that code? In any case it should not make any difference.

Also i think you read the diff wrong. The "solid" options are removed from the code not inserted. Previously we would have splattered them everywhere.

okomarov: Yep, read the diff wrong. Looks good. 

We should edit some contours to cover the cases in #867

miscco:
I have rebased against new develop. Given that it fixes the issues and there are no further comments/regressions, I will update the hashes and merge it tomorrow if nobody objects.

egeerardyn:
@miscco Code likes fine to me, thank you for your work on this! :+1: once the tests are green again.

Just a tiny recommendation: if you write "fixes #issue" in your PR or commits ([or similar phrases](https://help.github.com/articles/closing-issues-via-commit-messages/)), this will automatically:
- link the PR and issue in GitHub, (close the issue when we merge into `master`, but we tend to do that ourselves already)
- assign you for the referenced issue, and
- add the "work-in-progress" and remove the "ready-for-action" labels via [waffle.io](https://waffle.io/matlab2tikz/matlab2tikz) for the referenced issue.

miscco:
Hm thats strange that jenkins is ok but travis fails. I will check that tomorrow.

@matlab2tikzBot please test this.

miscco:
Does anybody has a clue whats going on here? jenkins is fine, which means at least octave 3.8 is happy. However the same cannot be said for octave 4.0 with travis. I run it locally at my computer and i get the same diff as from @egeerardyn server.

Ideas?

[diff_Contour_LinesOctave4.txt](https://github.com/matlab2tikz/matlab2tikz/files/158685/diff_Contour_LinesOctave4.txt)

miscco:
Ah i think i found the error, travis is considering unreliable tets, for which i did not update the hashes. However i do not understand why. For example ACID(1) is unreliable for 
```
stat.unreliable = isOctave || isMATLAB();
```

So it should always be recognized as that. Some ideas on that? Should i update the hashes?

miscco:
@matlab2tikzBot Please test this.

egeerardyn:
The outcome from Travis build 1007 is:

## Test summary
Test results for m2t commit c9d3748d7e running with Octave 4.0.0 on Linux 3.13.0-63-generic.

```matlab
suite = @ACID;
alltests = [1:100];
reliable = [2:5 7 9:16 19:20 22:24 28 32 35:37 39:40 42:44 46:57 59:62 64:77 79:85 87:90 93 97:100];
unreliable = [1 6 8 17:18 21 25:27 29:31 33:34 38 41 45 58 63 78 86 91:92 94:96];
failReliable = [9 87:90];
passUnreliable = [17 25:27 45 63 78];
skipped = [12 42:44 46:49 54 59 62 71:72 75:76 80 82 98];
```

| Unreliable |    7 |   19 |    0 |    26 |
|   Reliable |   51 |    5 |   18 |    74 |
|      Total |   58 |   24 |   18 |   100 |

Build fails with 5 errors. :heavy_exclamation_mark:

E.g. test 1 is seen as unreliable by Travis, so I think the problem is with the PR, not Travis (judging from build 1007).

miscco:
Ah now i am getting it, those tests do not run under your server, thats why i did not get the error from jenkins.

```
octave:9> ACID(9)
penny data set not found. Skipping.
```

miscco:
Here is the diff for the 5 tests that failed.

[diff_Octave4.txt](https://github.com/matlab2tikz/matlab2tikz/files/184384/diff_Octave4.txt)

Dumb me did fliped the ordering, but i think one can see that only "solid" is removed everytime.

egeerardyn:
:+1: Thanks for tackling this!

PR #880, opened 2016-04-06
miscellaneous fixes

patrickkwang:
There are a variety of things here:
* corrected the issue noted on 861 around ```hide axis``` (try ```imshow()``` with axis labels)
* added coloring of ticks and axis labels (e.g. test 20)
  * reference: [test20-reference.pdf](https://github.com/matlab2tikz/matlab2tikz/files/206597/test20-reference.pdf)
  * master: [test20-converted.pdf](https://github.com/matlab2tikz/matlab2tikz/files/206596/test20-converted.pdf)
  * now: [test20-converted.pdf](https://github.com/matlab2tikz/matlab2tikz/files/206598/test20-converted.pdf)
* moved the check for ```draw=none``` from ```getMarkerOptions``` to ```getLineOptions```
* switched from using ```draw=red``` for _marker colors_ to using ```mark options={red}``` so as not to fail when line and markers are different colors or when lines are absent
  * in ```setColor()```, prevented attempting to set the color of a line that already has ```draw=none```
  * generalized the allowed syntax for ```opts_print()``` to allow setting marker colors with ```setColor()```
* added limited support for ```stem3()```, still need to handle ```hgtransform``` and baseplane

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -2153,9 +2153,6 @@ function [m2t, str] = drawPatch(m2t, handle)
     % Line options
     [m2t, lineOptions] = getLineOptions(m2t, handle);
     drawOptions = opts_merge(drawOptions, lineOptions);
-    if isNone(handle.Marker) && (isNone(handle.LineStyle) || isNone(handle.EdgeColor))
-        drawOptions = opts_add(drawOptions,'draw opacity','0');
-    end
 
     % No patch: if one patch and single face/edge color
     isFaceColorFlat = isempty(strfind(opts_get(patchOptions, 'shader'),'interp'));
@@ -2290,6 +2287,9 @@ function [m2t, options] = setColor(m2t, handle, options, property, color, noneVa
     % that is set when the color == 'none' (if it is omitted, the property will not
     % be set).
     % TODO: probably this should be integrated with getAndCheckDefault etc.
+    if opts_has(options,property) && isNone(opts_get(options,property))
+        return
+    end
     if ~isNone(color)
         [m2t, xcolor] = getColor(m2t, handle, color, 'patch');
         if ~isempty(xcolor)
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -1367,6 +1367,18 @@ function [m2t, opts] = getTitleOrLabel_(m2t, handle, opts, labelKind, tikzKeywor
         end
         str = join(m2t, str, '\\[1ex]');
         opts =  opts_add(opts, tikzKeyword, sprintf('{%s}', str));
+    
+        % set color of axis labels
+        color = object.Color;
+        [m2t, col] = getColor(m2t, handle, color, 'patch');
+        if strcmpi(labelKind,'title')
+          % TODO: color titles
+        else
+          opts = ...
+              opts_add(opts, ...
+              ['every axis ',lower(labelKind(1)),' label/.append style'], ...
+              ['{font=\color{',col,'}}']);
+        end
     end
 end
 % ==============================================================================
@@ -1471,11 +1483,6 @@ function [m2t, options] = getAxisOptions(m2t, handle, axis)
             opts_add(options, ...
             ['every ',axis,' tick/.append style'], ...
             ['{',col,'}']);
-        % set color of axis labels
-        options = ...
-            opts_add(options, ...
-            ['every axis ',axis,' label/.append style'], ...
-            ['{font=\color{',col,'}}']);
     end
     % - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
     % handle the orientation
@@ -2295,7 +2302,11 @@ function [m2t, options] = setColor(m2t, handle, options, property, color, noneVa
         if ~isempty(xcolor)
             % this may happen when color == 'flat' and CData is Nx3, e.g. in
             % scatter plot or in patches
-            options = opts_add(options, property, xcolor);
+            if isempty(property)
+                options = opts_add(options, xcolor);
+            else
+                options = opts_add(options, property, xcolor);
+            end
         end
     else
         if exist('noneValue','var')
@@ -6458,9 +6469,7 @@ function str = opts_print(opts, sep)
     nOpts = size(opts,1);
     c = cell(1,nOpts);
     for k = 1:nOpts
-        if isempty(opts{k,1})
-            c{k} = sprintf('%s', opts{k,2});
-        elseif isempty(opts{k,2})
+        if isempty(opts{k,2})
             c{k} = sprintf('%s', opts{k,1});
         else
             c{k} = sprintf('%s=%s', opts{k,1}, opts{k,2});
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -859,8 +859,9 @@ function m2t = drawAxes(m2t, handle)
 
     if ~isVisible(handle)
         % Setting hide{x,y} axis also hides the axis labels in Pgfplots whereas
-        % in MATLAB, they may still be visible. Well.
-        m2t = m2t_addAxisOption(m2t, 'hide axis');
+        % in MATLAB, they may still be visible. Instead use the following.
+        m2t = m2t_addAxisOption(m2t, 'axis line style={draw=none}');
+        m2t = m2t_addAxisOption(m2t, 'ticks=none');
         %    % An invisible axes container *can* have visible children, so don't
         %    % immediately bail out here.
         %    children = allchild(handle);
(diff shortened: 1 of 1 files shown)


egeerardyn:
@matlab2tikzBot: test this PR

src/matlab2tikz.m
```diff
@@ -4016,7 +4039,11 @@ function warnFacetedInterp(m2t)
 % ==============================================================================
 function [m2t, str] = drawStemSeries(m2t, h)
     [m2t, str] = drawStemOrStairSeries_(m2t, h, 'ycomb');
-
+    
+    % TODO: handle baseplane with stem3()
+    if m2t.axesContainers{end}.is3D
+      return
```
egeerardyn: Doesn't this cause 3D stem/stair plots to be omitted completely?

src/matlab2tikz.m
```diff
@@ -859,8 +859,9 @@ case guitypes()
 
     if ~isVisible(handle)
         % Setting hide{x,y} axis also hides the axis labels in Pgfplots whereas
-        % in MATLAB, they may still be visible. Well.
-        m2t = m2t_addAxisOption(m2t, 'hide axis');
+        % in MATLAB, they may still be visible. Instead use the following.
+        m2t = m2t_addAxisOption(m2t, 'axis line style={draw=none}');
```
egeerardyn: Please use:
```matlab
m2t = m2t_addAxisOption(m2t, 'axis line style', '{draw=none}');
m2t = m2t_addAxisOption(m2t, 'ticks', 'none');
```
instead, since those are easier to track (those make sure that `axis line style` and `ticks` are not being set elsewhere).

src/matlab2tikz.m
```diff
@@ -6418,7 +6458,9 @@ function printAll(m2t, env, fid)
     nOpts = size(opts,1);
     c = cell(1,nOpts);
     for k = 1:nOpts
-        if isempty(opts{k,2})
+        if isempty(opts{k,1})
+            c{k} = sprintf('%s', opts{k,2});
```
egeerardyn: If this case occurs, then probably there is something else going wrong. The internal data structure of those option arrays should prevent this case from happening.

src/matlab2tikz.m
patrickkwang: No, ```m2t``` and ```str``` from ```drawStemOrStairSeries``` are still returned. In the 3D case, the next lines dealing with the baseline are skipped.

src/matlab2tikz.m
patrickkwang: Yes, this was a bit of a hack on my part. To allow the use of ```setColor()``` for marker colors in the ```mark options={red}``` format, I add an entry like this to the options array. I figured it would help me and was unlikely to break anything else. I'd appreciate any alternative suggestions.

src/matlab2tikz.m
patrickkwang: I see. Will do.

src/matlab2tikz.m
egeerardyn: Oh, yeah, indeed. I was being confused by the code folding of GitHub. :+1:

src/matlab2tikz.m
egeerardyn: I think I have another way that keeps that data structure cleaner.

In `setColor`, check whether the property name is empty. If it is, do `options = opts_add(options, color)`, otherwise, do what we do now.

src/matlab2tikz.m
```diff
@@ -2267,12 +2294,19 @@ function assertRegularAxes(handle)
...
         [m2t, xcolor] = getColor(m2t, handle, color, 'patch');
         if ~isempty(xcolor)
             % this may happen when color == 'flat' and CData is Nx3, e.g. in
             % scatter plot or in patches
-            options = opts_add(options, property, xcolor);
+            if isempty(property)
+                options = opts_add(options, xcolor);
+            else
+                options = opts_add(options, property, xcolor);
+            end
```
patrickkwang: I believe this is what you were suggesting.

src/matlab2tikz.m
```diff
@@ -1366,6 +1367,18 @@ case guitypes()
...
+        color = object.Color;
+        [m2t, col] = getColor(m2t, handle, color, 'patch');
+        if strcmpi(labelKind,'title')
+          % TODO: color titles
+        else
+          opts = ...
+              opts_add(opts, ...
+              ['every axis ',lower(labelKind(1)),' label/.append style'], ...
+              ['{font=\color{',col,'}}']);
+        end
```
patrickkwang: Since axis label colors can differ from the axis colors, I moved the check for axis label colors from ```getAxisOptions()``` to ```getTitleOrLabel_()```. I've just realized that ```object.Color``` is probably not backwards compatible, so I should probably fix that.

src/matlab2tikz.m
miscco: Didnt we have a helper for the Option={Arg1,...} case?

miscco:
@egeerardyn There seems to be another case where Octave4.0 only tests fail. Is there the posibility to run Octave4.0 on your server so we can include it into jenkins?

src/matlab2tikz.m
```diff
@@ -1366,6 +1367,18 @@ case guitypes()
         end
         str = join(m2t, str, '\\[1ex]');
         opts =  opts_add(opts, tikzKeyword, sprintf('{%s}', str));
+    
+        % set color of axis labels
+        color = object.Color;
+        [m2t, col] = getColor(m2t, handle, color, 'patch');
+        if strcmpi(labelKind,'title')
+          % TODO: color titles
```
miscco: While you are at it. there is the title style = {} option, where we could set the color. That might be nice to

src/matlab2tikz.m
```diff
@@ -1366,6 +1367,18 @@ case guitypes()
...
+        else
+          opts = ...
+              opts_add(opts, ...
+              ['every axis ',lower(labelKind(1)),' label/.append style'], ...
```
miscco: Every axis seems like a bug in waiting. What if i want to color the axes differently. I think you have to check for the axis and add x/y/z axis styles.

miscco:
BTW as a general remark, I see that you directly worked on your develop branch. This is not really advisable, as you will get in trouble once an other PR is merged before yours.

Generally you shoud create your own branch for every project your working on.

src/matlab2tikz.m
egeerardyn: @patrickkwang  Exactly. :+1:
@miscco Not that I know of. Note that this is within some function related to colors where one of the arguments was `key` in `key=value`. The patch makes sure that we can signal that we need just `value` in the TikZ output.

src/matlab2tikz.m
egeerardyn: @patrickkwang :+1:, but indeed, `color = object.Color` should be replaced by `color = get(object, 'Color')` for backwards compatibility.

egeerardyn:
> @egeerardyn There seems to be another case where Octave4.0 only tests fail. Is there the posibility to run Octave4.0 on your server so we can include it into jenkins?

@miscco This should be feasible. I haven't looked into installing both Octave 4 and 3.8 at the same time, but it should be possible. If I find the courage this weekend, I will have a go at it (but it seems my server is down right now, will check tonight). Worst-case we drop support for Octave 3.8: 4.0 has been out for a year now and it is free software. Moreover, the testing suite is a burden on our side, I don't expect that our few Octave users will notice anything.

src/matlab2tikz.m
miscco: We have, id introduced it in one of my later PRs

```matlab
function opts = opts_addSubOpts(opts, key, subOpts)
    % add a key={Opts} pair to an options array
    formatted = ['{' opts_print(subOpts) '}'];
    opts      = opts_add(opts, key, formatted);
end
```

src/matlab2tikz.m
patrickkwang: ```lower(labelKind(1))``` does this. I'm actually not sure what ```every``` contributes here, but my tests for single-axis coloring work.

miscco: Every axis y seems strange, but it would appear in every y axis i assume? Maybe just drop the every

patrickkwang: It does not seem to work without the ```every```. My TikZ experience is insufficient to explain why.

src/matlab2tikz.m
egeerardyn: @miscco That introduces `key={k=v,k=v,...}` not `key`. Anyhow, there are sections in this PR where `opts_addSubOpts` could come in handy, sure. I will address those upon merge, that's small stuff to fix (it's more complicated to do it using that function, but it provides consistency in the output). IMHO, it's not something to bother a first-timer with (the same for our test suite).

From within the `setColor` function, there was no mechanism to enforce `key` instead of `key=value` until now.

src/matlab2tikz.m
egeerardyn: The `every X` seems strange, but that's how it's supposed to be. In any case (or as far as I understand and have some clarity), those `every X` statements are still contained within the `axis` (and `tikzpicture`) scope, so it shouldn't propagate further to other scopes.

I think it looks nicer without the `every`, so probably this excerpt from the manual helps a bit to pick an/any alternative:
screen shot 2016-04-08 at 19 12 23

src/matlab2tikz.m
```diff
@@ -4455,14 +4489,19 @@ function warnFacetedInterp(m2t)
...
         fontStyle = sprintf('%s\\bfseries',fontStyle);
     end
     if strcmpi(get(handle, 'FontAngle'), 'Italic')
         fontStyle = sprintf('%s\\itshape',fontStyle);
     end
+    if ~all(get(handle, 'Color')==0)
+        color = get(handle, 'Color');
+        [m2t, col] = getColor(m2t, handle, color, 'patch');
+        fontStyle = sprintf('%s\\color{%s}', fontStyle, col);
+    end
```
patrickkwang: After further exploration, I decided the right thing to do is to move the coloring for text, axis labels, and titles all into here.

egeerardyn:

miscco:
As this was taken over in #908 I am closing this for now

Issue #874, opened 2016-03-21
Plotyy: PlotyyLabelNum not set


egeerardyn:
During the conversion of some of my old FIG files to TikZ, I encountered a problem with one of those.
Now, this might be a very uncommon situation, since the FIG files were generated with an older version of MATLAB (2012a/b, I suppose), yet I call the conversion in R2016a.

As I start from a FIG file, I don't have code to reproduce this and I'm just dumping this here to ensure that we keep track of it.

To reproduce the issue:
 - FIG file is available at https://dl.dropboxusercontent.com/u/452005/m2t/issue874/pdfs.fig

Code used during conversion:

```matlab
files = dir('*.fig');

asTikz = @(file) strrep(file.name, '.fig','.tikz');

for file = files(:)'
    fprintf(' %s \t -> \t %s \n', file.name, asTikz(file));
    
    hFig = open(file.name);
    finallyCloseFigure = onCleanup(@() close(hFig));
    try
        matlab2tikz(asTikz(file), 'figurehandle', hFig, ...
                    'width', '\figurewidth', ...
                    'height', '\figureheight', ...
                    'standalone', false, ...
                    'externalData', true,...
                    'showInfo', false);
    catch ME
        fprintf(2,'Error "%s" during conversion of %s\n%s\n\n', ...
                ME.identifier, file.name, getReport(ME));
    end
end
```

The figure should look something like:
<img width="549" alt="screen shot 2016-03-21 at 11 15 05" src="https://cloud.githubusercontent.com/assets/177360/13916052/89437156-ef58-11e5-99a2-2df07d9d4632.png">

I currently **work around** the issue by means of a patch:
```matlab
function m2t = getPlotyyReferences(m2t, axisHandle)
%% ... help text ...

m2t.PlotyyLabelNum = []; %% ADD THIS LINE AT THE BEGINNING

%% ... other code ...
```
which prevents m2t from crashing, but might have unintended side-effects.
Probably the solution is to be safer about the `m2t.PlotyyLabelNum` variable, e.g. initialize it globally.

egeerardyn:
After #883, this workaround is no longer needed. However, the figure doesn't compile perfectly (at the tip of #883 at the time of writing), since both axes are in the wrong order for LaTeX:
 - the first axis defines the legend (i.e. references some styles)
 - the second axis actually defines the styles:

```latex
% This file was created by matlab2tikz.
%
%The latest updates can be retrieved from
[...]

PR #883, opened 2016-04-08
Fix plotyy multiple traces

### LINKED ISSUE #882: MATLAB error when exporting two-axes plot w/ multiple plots per axis
Hello,

I have encountered a (rather seldom) MATLAB error when I tried to export a two-axes `plotyy` where there were multiple plots at a single axis, in combination with a legend.

A simple non-working example is given by ([test_plotyy2tikz.m](https://github.com/matlab2tikz/matlab2tikz/files/209941/test_plotyy2tikz.m.txt))

    x = 0:0.01:20;
    y1 = 200*exp(-0.05*x).*sin(x);
    y2 = 0.8*exp(-0.5*x).*sin(10*x);
    y3 = 0.8*exp(-0.5*x).*sin(x);

    figure
    %plots y2 and y3 both on secondary axis
    plotyy(x,y1,x,[y2; y3]);
    legend('y = e^{-0.05 x} \sin(x)', 'y = e^{-0.5 x} \sin(10 x)', 'y = e^{-0.5 x} \sin(x)');
    
    matlab2tikz('test_plotyy.tex');

This leads to the following MATLAB error: ([full stack trace](https://github.com/matlab2tikz/matlab2tikz/files/209942/plotyytikz_stacktrace.txt)  attached)

> Index exceeds matrix dimensions.
>
> Error in matlab2tikz>addPlotyyReference (line 754)
>            ref = m2t.axesContainers{end}.PlotyyReferences(ii);

Diving into matlab2tikz, the error seems to be raised because the lines 749--762 are executed twice, apparently because of the two plots on the secondary axis:

    749: elseif m2t.currentHandleHasLegend
    750:    % We are on the secondary axis
    [751--752]
    753:    for ii = 1:labelNum
    754:        ref = m2t.axesContainers{end}.PlotyyReferences(ii);
    [755--757]
    758:    end
    [759--761]
    762:    m2t.axesContainers{end}.PlotyyReferences = [];

But, for the assignment in line 762 the access of the `ii`-th element in line 750 fails the second time. My workaround now was to modify line 749 to

    749: elseif m2t.currentHandleHasLegend && ~isempty(m2t.axesContainers{end}.PlotyyReferences)

ensuring that the block is executed only once. Therewith, matlab2tikz runs successfully the tikz-outcome looks as expected.

egeerardyn:
This fixes #882.


miscco:
Seems reasonable for me, however i am wondering what happens if we have multiple plotyys in a subplot configuration.

So far for that :+1:

egeerardyn:
@miscco That's a very reasonable question. The answer: I don't know.

BTW, I tested my work-around in #874 on top of this PR as well, but I produced other output code (and I didn't investigate any further).

Anyhow, I think the code that handles those YY legends is:
 - a bit too complicated: there are a few m2t fields that have some state in them, while it is not clear what each of them tracks and I suspect their state is set inconsistently: one action is iterating over the references with the label numbers as index. Later, the references are cleared, but the numbers are not.
 - liable for errors: I expect multiply defined labels, things like #874, and your suspicion of troubles with multiple YY plots.

So, I think we should revisit that code and clean it up before a release.

egeerardyn:

egeerardyn:

egeerardyn:
This fixes #882 and #885. Also, this should fix most of #884, but that needs adapted test cases, which I will gladly do after a review.

src/matlab2tikz.m
```diff
@@ -154,32 +154,21 @@ function matlab2tikz(varargin)
...
     checkDeprecatedEnvironment(minimalVersion);
 
-    m2t.cmdOpts = [];
-    m2t.currentHandles = [];
-
-    m2t.transform = []; % For hgtransform groups
+    m2t.args            = []; % For command line arguments
+    m2t.current         = []; % For currently active objects
+    m2t.transform       = []; % For hgtransform groups
     m2t.pgfplotsVersion = [1,3];
```
miscco: Shouldnt we update this to a more recent version?

src/matlab2tikz.m
```diff
@@ -154,32 +154,21 @@ function matlab2tikz(varargin)
...
-    m2t.about.authorEmail = 'nico.schloemer@gmail.com';
-    m2t.about.years = '2008--2015';
-    m2t.about.website = 'http://www.mathworks.com/matlabcentral/fileexchange/22022-matlab2tikz-matlab2tikz';
-    m2t.about.github = 'https://github.com/matlab2tikz/matlab2tikz';
-    m2t.about.wiki = [m2t.about.github '/wiki'];
-    m2t.about.issues = [m2t.about.github '/issues'];
+    m2t.about.name      = 'matlab2tikz';
+    m2t.about.version   = '1.0.0';
+    m2t.about.years     = '2008--2016';
+    m2t.about.website   = 'http://www.mathworks.com/matlabcentral/fileexchange/22022-matlab2tikz-matlab2tikz';
```
miscco: Hm i think we should refer to github too, as matlabcentral is not really up to date all the time or at least needs some effort to keep it updated.

src/matlab2tikz.m
```diff
@@ -349,21 +341,44 @@ function matlab2tikz(varargin)
...
     %% Check for a new matlab2tikz version outside version control
-    if m2t.cmdOpts.Results.checkForUpdates
-        m2tUpdater(m2t.about, m2t.cmdOpts.Results.showInfo);
+    if m2t.args.checkForUpdates
+        m2tUpdater(m2t.about, m2t.args.showInfo);
     end
 
 end
 % ==============================================================================
+function [m2t, counterValue] = getNextValue(m2t, counterName)
```
miscco: I think getNextValue should better be named increaseValue or increaseGlobalValue.

Also how much work would it be to remove counterValue and just refert to m2t.count.* in the code?

src/matlab2tikz.m
```diff
@@ -455,7 +470,7 @@ function matlab2tikz(varargin)
     [m2t, axesBoundingBox] = getRelevantAxes(m2t, axesHandles);
 
     m2t.axesBoundingBox = axesBoundingBox;
-    m2t.axesContainers = {};
+    m2t.ax = {};
```
miscco: Can we make that m2t.axes ? these 2 characters do not hurt that much

src/matlab2tikz.m
```diff
@@ -545,17 +559,19 @@ function addStandalone(m2t, fid, part)
...
+    % make sure to append with '%' to avoid spacing woes
+    FORMAT = ['\\definecolor{%s}{rgb}{' ff ',' ff ',' ff '}%%\n'];
     if ~isempty(names)
-        m2t.content.colors = sprintf('%%\n%% defining custom colors\n');
         colorDef = cell(1, length(names));
         for k = 1:length(names)
-            % make sure to append with '%' to avoid spacing woes
-            colorDef{k}= sprintf(['\\definecolor{%s}{rgb}{', ff, ',', ff, ',', ff,'}%%\n'], ...
-                                  names{k}, specs{k});
+            colorDef{k}= sprintf(FORMAT, names{k}, specs{k});
```
miscco: I dont really like this. It hides the {%s} assignment

src/matlab2tikz.m
```diff
@@ -732,40 +747,54 @@ case guitypes()
...
-    labelNum = m2t.PlotyyLabelNum;
 
     if hasPlotyyReference(m2t,h)
         % Label the plot to later reference it. Only legend entries on the main
         % plotyy axis will have a label
-        labelNum = labelNum + 1;
-        label = sprintf('\\label{plotyyref:leg%d};\n\n', labelNum);
-        m2t.PlotyyLabelNum = labelNum;
+        [m2t, labelNum] = getNextValue(m2t, 'plotyylabel');
+        label = sprintf('\\label{%s};\n\n', plotyyLabelName(labelNum));
```
miscco: Couldnt we just pass m2t to plotyyLabelName? therefore we could omit the assignment of labelNum

src/matlab2tikz.m
miscco: Scrap the last one, i think returning it is better

miscco:
So in total I really like it a lot.

Great Work :+1:

src/matlab2tikz.m

src/matlab2tikz.m
egeerardyn: We could, but I like it better this way. That makes it clear we only need the label number, not a whole lot of other stuff.

miscco: Fine with me, it doesnt really has any benefit if we do assign labelNum anyway

src/matlab2tikz.m
egeerardyn: That line only specifies the minimal version that is needed. Also, there is nothing fancy going on with the version number: we dynamically determine which version is the minimal supported one depending on the feature we use and just show it to the user. In standalone, we always use the `newest` version.

src/matlab2tikz.m
egeerardyn: See 1 line lower.

src/matlab2tikz.m
egeerardyn: I don't really see what you mean. Could you elaborate?

miscco: FORMAT hides the assignment of the other arguments. Which kind of reduces readability

src/matlab2tikz.m
miscco: Yes, but normally the website is what people go to first.

src/matlab2tikz.m
egeerardyn: I don't think it reduces readability: the float format is just a detail (and actually constants). Personally, I am not that fond of constructing the format of those statements in-line when there are some concatenations going on: it typically becomes a hot mess of commas.

However, what could offer the best of both worlds, is that I define `FORMAT` right before it is used instead of before the loop. What do you think?

miscco: :+1: for that. I think those places are not really performance critical

src/matlab2tikz.m
egeerardyn: I will re-write the message we show.

src/matlab2tikz.m
egeerardyn: I agree (it can't be slower than what we had before, so I don't care). As I remember, the performance-critical stuff is rather in `makeTable` (when working with large datasets at least).

egeerardyn:
I've updated the message we show to our users as you asked.

In MATLAB this becomes:

> This is matlab2tikz v1.0.0 (commit 7b24d087dd21834741e6a2be80610f1c1ce6e143).
>  
> The latest developments can be retrieved from [our development branch](https://github.com/matlab2tikz/matlab2tikz/tree/develop).
> You can find more documentation on our [GitHub page](https://github.com/matlab2tikz/matlab2tikz) and [wiki](https://github.com/matlab2tikz/matlab2tikz/wiki).
> If you encounter bugs or want a new feature, go to [our issue tracker](https://github.com/matlab2tikz/matlab2tikz/issues).
> Please visit [FileExchange](http://www.mathworks.com/matlabcentral/fileexchange/22022-matlab2tikz-matlab2tikz) to rate matlab2tikz or download the stable release.

In Octave it looks like:
```
 *** This is matlab2tikz v1.0.0 (commit 7b24d087dd21834741e6a2be80610f1c1ce6e143).
 ***
 *** The latest developments can be retrieved from https://github.com/matlab2tikz/matlab2tikz/tree/develop.
 *** You can find more documentation on
 ***  https://github.com/matlab2tikz/matlab2tikz
 ***   and
 ***  https://github.com/matlab2tikz/matlab2tikz/wiki.
 *** If you encounter bugs or want a new feature, go to
 ***  https://github.com/matlab2tikz/matlab2tikz/issues.
 *** Please visit http://www.mathworks.com/matlabcentral/fileexchange/22022-matlab2tikz-matlab2tikz
 ***   to rate matlab2tikz or download the stable release.
```

In MATLAB everything is flush left, in Octave the spacing is a tiny bit awkward, but good enough if you ask me.

miscco:
Very nice. That should help to keep the users updated.

miscco:
So from my point of view :+1:  

Awesome work!

egeerardyn:
Thank you for your review, you brought up some really valuable points.

src/matlab2tikz.m
```diff
@@ -518,19 +531,19 @@ function matlab2tikz(varargin)
...
-    addCustomCode(fid, '\n', m2t.cmdOpts.Results.extraCodeAtEnd, '');
+    addCustomCode(fid, '\n', m2t.args.extraCodeAtEnd, '');
 
     addStandalone(m2t, fid, 'end');
 end
 % ==============================================================================
 function addStandalone(m2t, fid, part)
     % writes a part of a standalone LaTeX file definition
-    if m2t.cmdOpts.Results.standalone
+    if m2t.args.standalone
```
miscco: So this is not related to the PR, but what do you think about allowing custom preambles/standalones?

Ipersonally have 3 matlab2tikz versions. One i use for my Thesis/Papers which has some modifications for the preamble that are necessary for my workflow, one for development and one for normal stuff/testing.

It would be awesome if we would add "CustomeStandalone" or whatever.

Thoughts?

egeerardyn: I really like it, but please open a feature request for that (maybe with the current changes you have). I suppose it would be enough to add a parameter `preamble` to our input arguments, change a few references here or there and done?

PR #895, opened 2016-05-10
Color map custum label

RPeschke:
added support for custom Tick labels on color bar 

code:
 ```matlab
f= figure();
h=pcolor(peaks);

hcb=colorbar;
set(hcb,'YTick',[-6 0 6 ],'YTickLabel',{'Good','Normal','Bad'})
matlab2tikz('figurehandle',f,...
    'filename','example_colorMap.tex' ,...
    'showInfo', false,...
    'standalone', true);
close(f)

```

without the Changes:
example_colormap_without

With the changes: 
example_colormap_with

### CODE REVIEW

**Comment by miscco** [comment]
@matlab2tikzBot please test this

File: src/matlab2tikz.m [hunk 1]
```diff
@@ -4500,6 +4500,34 @@ function warnFacetedInterp(m2t)
     end
 end
 % ==============================================================================
+function   cbarStyleOptions = getColotBarTicks(handle, cbarStyleOptions)
```
miscco: There is a typo in the function name.

src/matlab2tikz.m
```diff
@@ -4500,6 +4500,34 @@ function warnFacetedInterp(m2t)
...
+if strcmp( get(handle,'TickLabelsMode') , 'manual')
+
```
miscco: Please generate an opts structure as we do in the rest of the code. That makes things much easier

src/matlab2tikz.m
```diff
@@ -4500,6 +4500,34 @@ function warnFacetedInterp(m2t)
...
+    %ColorBar TickLabels 
+    TickLabels= get(handle,'TickLabels');
+    TickLabels_option ='{';
+    for e=1:length(TickLabels)
+        TickLabels_option = [ TickLabels_option ' {' TickLabels{e} '},'];
+    end
+    TickLabels_option(end)='}';
+
+    cbarStyleOptions=cat(1,cbarStyleOptions, {'yticklabels',TickLabels_option});
```
miscco: This seems really cumbersome and I would guess it fails for empty ticks?

Please have a look at
```matlab
function [options] = getAxisTicks(m2t, handle, axis, options)
```

Basically that is exactly what you would want here.

miscco:
Thank you very much for that PR. That is a good usecase and will definitely improve m2t. 

However, I think the code is not there yet. Especially considering that we already have an infrastructure for TickLabels.

From a glance at 
```matlab
function [options] = getAxisTicks(m2t, handle, axis, options)
```
I would guess, that you could achieve al that you want by passing an empty "axis" argument. Maybe we will need some checks against that but I think that might be a suitable approach.

RPeschke:
Thanks for your succession I found that all the functionality was already in the project I just needed to apply it to this axis label type.

src/matlab2tikz.m
```diff
@@ -4562,9 +4562,17 @@ function warnFacetedInterp(m2t)
         end
     end
 
-    % Merge them together in axisOptions.
+   % Merge them together in axisOptions.
     axisOptions = opts_add(axisOptions, strtrim(['colorbar ', cbarTemplate]));
 
+   
+    %color map ticks
+    [m2t, options] = getAxisOptions(m2t, handle, 'y');
```
miscco: Here you can just pass cbarStyleOptions directly.

Then there is no need to merge below.

miscco:
I am unsure about the attached pdf/jpg. They generally do not make thatmuch sense if you are not a developer.

@egeerardyn thoughts?

src/matlab2tikz.m
miscco: Additionally getAxisOptions assumes that options is a valid container, so that would break most likely

egeerardyn:
@miscco I agree. Attaching the produced output makes little sense for a normal user, I would say. So it's probably better to remove them from the repository.

okomarov:
We should add the feature to `getColorbarOptions()` mostly as is, and eventually clean up the code in another PR. 

For this, we need:
- drop the .pdfs and .png
- drop `opts_append_unique()`, not sure what purpose irt serves, since it's not used
- adapt one of the existing ACID example to use custom colorbar labels, akin to the provided example. We do not want proliferation of test cases...

Issue #896, opened 2016-05-16
'LineStyle','none'  does not work with "fill" function


pfbuxton:
Hello,

If you run this code the pdf will have lines along the filled area. Note, in version-0.6.0 the fill command works as expected without the lines.

```matlab
clear;
figure;

theta = linspace( 0.0, 2.0*pi, 20 );
fill( cos(theta), sin(theta), 'Red', 'LineStyle','none' );

matlab2tikz( 'test.tikz', 'standalone',true );
system( 'pdflatex test.tikz' );
```

Thanks,
Peter

egeerardyn:
I can reproduce this in R2014a and R2014b. Thank you for reporting this!

miscco:
It seems that we have to add a check for linewidth in [m2t, str] = drawPatch(m2t, handle)

I would suggest the following
```matlab
        if isLineVisible(handle)
        [m2t, drawOptions] = setColor(m2t, handle, drawOptions, 'draw', ...
                                         s.edgeColor, 'none');
       end
        [m2t, drawOptions] = setColor(m2t, handle, drawOptions, 'fill', ...
                                         s.faceColor);
```

miscco:
Meh doesnt work, one has to set the edge color to none ...

miscco:
I opened a PR that should fix that #898

egeerardyn:
I can confirm (both R2014a and R2014b) that this is fixed by #898.

PR #898, opened 2016-05-16
Check for edge visibility of patches

miscco:
As discussed in #896 the code currently always draws the edge. This was currently shadowed if the edge color was the same as that of the patch and/or if it was set explicitely.

Therefore, add a check for the visibility of a line of patch plots so the edge is only drawn when visible.


miscco:
@egeerardyn Given that this does not lead to any breakage. Would you mind to merge this PR?

egeerardyn:
:+1: Thank you for tackling this issue faster than your shadow!

miscco:
:smile: Just lucky, that i was sitting alone at home with m2t open ;)

Issue #899, opened 2016-05-16
scatter plot: false colors when using MarkerEdgeColor


twippenbeck:
When using MarkerEdgeColor false colors are created after matlab2tikz-conversion with recent dev-version (see MWE + pictures below).

Whilst maybe related to #859 #861 #860 I do not see if these will solve the problem.

Is there a quick fix / workarround available? I have ~300 figures with custom colors to do, so modifying the tikz files by hand is quite ...

Thx!

MWE:

    figure;
    
    colors = [0 1 0; 0 0 1; 1 0 0];

    for iColor = 1:size(colors,1)
        X = rand(10,2);
        scatter(X(:,1),X(:,2),'o','MarkerEdgeColor',colors(iColor,:));
        hold all;
    end
    
    saveas(gcf,'MWE_scatter_EdgeColor_MatlabOrig.png')
    
    matlab2tikz(...
                'filename','MWE_scatter_EdgeColor.tex',...
                'encoding','UTF-8',...
                'standalone',true...
                );

MATLAB figure:
![mwe_scatter_edgecolor_matlaborig](https://cloud.githubusercontent.com/assets/14804753/15301120/6342de62-1bac-11e6-8301-3045c7dbc09f.png)

Output by M2T:
![mwe_scatter_edgecolor_matlab2tikz](https://cloud.githubusercontent.com/assets/14804753/15301128/6cd4b09a-1bac-11e6-9c06-367d79f1bbda.png)

miscco:
Hm, as i understand it, there is a conflict between the cData and Markersize.

Have you tried to pass the color as the fourth input according to 
http://de.mathworks.com/help/matlab/ref/scatter.html

miscco:
So regarding your problem i expect the culprit to be getScatterOptsOneColor

There we have the following code
```matlab
        if hasFaceColor && hasEdgeColor
            drawOptions = opts_add(drawOptions, 'draw', ecolor);
            drawOptions = opts_add(drawOptions, 'fill', xcolor);
        else
            drawOptions = opts_add(drawOptions, 'color', xcolor);
        end
```

So we default to the fill color if only one of the two is set. I dont really know if that is reasonable. Maybe we should set fill and draw seperately?

twippenbeck:
1. fantastic! workaround  (color as the fourth input ) works perfectly for predefined and user defined colors (tested). you made my day :-)
[...]

PR #900, opened 2016-05-16
Always respect EdgeColor in Scatterplots

miscco:
This is related to #899 

Similarly to the variable size case, respect the edge color of markers. I dont really know if we couldnt simplify this further.

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -3619,11 +3619,13 @@ function [m2t, drawOptions] = getScatterOptsOneColor(m2t, h, drawOptions, ...
                                    markerInfo.options);
         drawOptions = opts_add(drawOptions, 'mark size', ...
             sprintf('%.4fpt', sData)); % FIXME: investigate whether to use `m2t.ff`
-        if hasFaceColor && hasEdgeColor
+        if hasEdgeColor
             drawOptions = opts_add(drawOptions, 'draw', ecolor);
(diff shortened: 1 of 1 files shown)


egeerardyn:
Good job, I just added a TODO since I think one of the branches is not that clear (but that is older code).

miscco:
Yeah, I think some of the code is even broken as it checks for length (cData)==3 and size(cData,2)==3. The latter is most likely never executed, as cData most often only has lenth 3.

On the other hand i believe, that your refactoring of Scatter plots in #861 should help there a lot.

egeerardyn:
I agree with you: quite a lot of the code is not properly covered by tests and might be broken badly. I think I will push #861 after the 1.1 release, it's more important we release a new version soon.

twippenbeck:
@miscco @egeerardyn 
please undo that merge of #900 from the private repo of miscco
it totaly destroys other functionalities like e.g. stacked bar plots
I assume some changes of the forked repo have not been compatible
As far as I can see, the merge merged many more lines than the central ones of #900

egeerardyn:
@twippenbeck Were you using the `master` (or version from FileExchange) before that?
If so, please open a new issue for what errors you get right now.
Otherwise, could you give an example (i.e. MATLAB code) that produces poor figures right now?

miscco:
@twippenbeck Please first check, that you have the current develop version of m2t. If there is still a problem please open a new issue. 
However, given that our test suite includes stacked bar plots and others this seems strange.

PR #903, opened 2016-05-17
Fix incorrect bar shift auto assignment for stacked barseries

### LINKED ISSUE #901: Stacked bar plots: errors after #851
merge #900 destroys stacked bar plots in current development version

MWE from matlab doc:

    figure;
    
    y = [2 2 3; 2 5 6; 2 8 9; 2 11 12];
    bar(y,'stacked')
    
    matlab2tikz(...
                'filename','MWE_merge900.tex',...
                'encoding','UTF-8',...
                'standalone',true...
                );

Output with current dev version:
mwe_merge900

miscco:
Move the addition of bar shift auto to the setBarLayoutOfBarSeries function and only apply it to grouped plots. This fixes #901.

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -3810,9 +3810,6 @@ function [m2t, str] = drawBarseries(m2t, h)
     else
         barType = 'ybar';
     end
-    % Tthe bar shift auto feature was introduced in pgfplots 1.13
-    m2t = needsPgfplotsVersion(m2t, [1,13]);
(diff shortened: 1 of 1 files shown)


miscco:
So i have checked, that both the stacked and the grouped case create the correct behavior.

Wasnt there a Test in the suites with stacked barplots?

miscco:
It gives me serious headaches, that the test suite doesnt complain about this, see #904.

miscco:
@egeerardyn Can you please double check this so we can restore the functionality

egeerardyn:
👍

Issue #905, opened 2016-05-20
yyaxis confuses LaTeX


Pratched:
hi, im trying to implement some figures from Matlab via matlab2tikz into LaTeX. But while runnig
```
\begin{figure}
  \input{....tikz}
\end{figure}
```
I do get a few errors containing: 'dimension too large', 'cant work with sizes bigger than 19feet', etc.
I do realize that this is probably a problem made up by LaTeX but when creating the pdf file, the actual figure appears. Well I could ignore the errors for now, but the figure should have a left and a right y axis. Which I createt via `yyaxis` in Matlab. 
It appears to me, that the decision which y axis will be plotted in LaTeX depends on whether I executed `yyaxis left` or `yyaxis right` at last. Beacause I always do get the same error code, but on different lines, depending the last yyaxis in Matlab. And so is the plotted y axis changing.
Is there any way (Matlab, but maybe you even do have the solution for LaTeX) I can (easily) get both y axis plotted?

I do have some basic skills with Matlab and LaTeX but that's it. So I i apologize in advanced for some foolish mistakes.

miscco:
Hi can you please give us a minimal working example?

I have never used the yyaxis function myself, but only plotyy. Maybe that would be a workaround for you?

miscco:
As a side not i would like to mention, that this function has been introduced in Matlab2016a

I dont know if we already support those new function. In any case i only have 2015a so i cannot test this.

Pratched:
Thanks for the fast repley.
Well I actually started with plotyy, but it was way more easier to 
```
hold on; 
plot( x, y1, x, y2);
yyaxis rigth;
plot(x, y3);
```
than try do plot 3 graphs with `plotyy`.

Well, I am using the student version(2016a since a few weeks) and Mathworks.com (or so) suggested to use yyaxis instead of plotyy. No I can see, why I cant find very much information about yyaxis.
I am not quit sure what do you mena by working example. Should I post the Matlab code itself?

Pratched:
Maybe it is importent that x is date. I did implement `datenum` above `hold on`.
```
x=datenum(x);
hold on; 
plot( x, y1, x, y2);
yyaxis rigth;
plot(x, y3);
datetick;
```
x was actualy a date with the format: dd.mm.yyyy HH:SS. Since the Date for all 177 lines of x is the same, after `datetick` only the tim (HH:SS) appears at the x axis.
Maybe this is important as well.

miscco:
[...]

PR #907, opened 2016-05-20
Normalize axis data

miscco:
this is a very hacky try to make datetick compilable for m2t inspired by #905.

The idea is simple. Tacke the rather stupid matlab format and project it into a reasonable range tixz can work with. There is obviously a big drawback, as it is done for all the the axes.

If the xaxis is in datetick one does
```matlab
cleanfigure('transformDates', 'x')
```

### DIFF
--- a/src/cleanfigure.m
+++ b/src/cleanfigure.m
@@ -23,7 +23,7 @@ function cleanfigure(varargin)
 %   or negative values disable this feature.
 %   (default: 1)
 %   CLEANFIGURE('normalizeAxis','xyz',...)
-%   EXPERIMENTAL: Normalize the data of the dimensions spezified by
+%   EXPERIMENTAL: Normalize the data of the dimensions specified by
 %   'normalizeAxis' to the interval [0, 1]. This might have side effects
 %   with hgtransform and friends. One can directly pass the axis handle to
 %   cleanfigure to ensure that only one axis gets normalized.
--- a/src/cleanfigure.m
+++ b/src/cleanfigure.m
@@ -22,6 +22,13 @@ function cleanfigure(varargin)
 %   Scale the precision the data is represented with. Setting it to 0
 %   or negative values disable this feature.
 %   (default: 1)
+%   CLEANFIGURE('normalizeAxis','xyz',...)
+%   EXPERIMENTAL: Normalize the data of the dimensions spezified by
+%   'normalizeAxis' to the interval [0, 1]. This might have side effects
+%   with hgtransform and friends. One can directly pass the axis handle to
+%   cleanfigure to ensure that only one axis gets normalized.
+%   Usage: Input 'xz' normalizes only x- and zData but not yData
+%   (default: '')
 %
 %   Example
 %      x = -pi:pi/1000:pi;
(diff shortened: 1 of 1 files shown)


egeerardyn:
This just needs a bit of documentation (in the header of `cleanfigure`) to show how it can be used.

Note also that you can get it to work on a single axes object: `cleanfigure(hAx, ...)` should work without a problem. That is something you could also state in the docs.

miscco:
Yeah, thats true.

However, i thought about epanding it. In general, this has nothing to do with dateticks, but with a normalization of the axis data. So i am inklined to rename it to normalize_axis.

Also i believe i have to change the setup, so that it affects anything that has [upper(axis), 'Data'] in it. You know the figure setup better than me. Would it be sufficient to just go through all the direct children of the axis, or do i have to consider the grandchilds too?

egeerardyn:
@miscco I suppose that you need to handle all ancestors (but I imagine there are quite a lot of edge cases that this cannot handle: e.g. imagine some `hgtransform` somewhere, then you can no longer determine which dimension is affected).

Anyhow, since this is a work-around (and quite experimental), I'm fine with such limitations. Perhaps we should, however, add a warning to the documentation and/or at runtime to warn the users that this is experimental functionality.

miscco:
So i have reworked this to better express, what it is actually doing and added both the comments as well as the warning in the function itself. 

@egeerardyn I have no idea why the build failed. Can you enlighten me?

egeerardyn:
@miscco Jenkins configuration error. My bad (performed an update recently, apparently not the whole config was migrated). Will check out what is happening. Code itself ran fine, BTW.

egeerardyn:

src/cleanfigure.m
```diff
@@ -22,6 +22,13 @@ function cleanfigure(varargin)
 %   Scale the precision the data is represented with. Setting it to 0
 %   or negative values disable this feature.
 %   (default: 1)
+%   CLEANFIGURE('normalizeAxis','xyz',...)
+%   EXPERIMENTAL: Normalize the data of the dimensions spezified by
```
okomarov: typo: specified

egeerardyn:

egeerardyn:
@miscco I updated Jenkins, this fixed the internal errors which caused the build to fail. Since I think Jenkins is a beast, I will look for alternatives that integrate better with GitHub when I find some time. A possible candidate would be TeamCity.

miscco:
So I found a  bug regarding the TickLabels. Not setting the TickLabelMode to manual changed the TickLabels to their new position. 

I tried a variety of ACID test that seemed to work fine. At least I did not find any corruption in MATLAB.

miscco:
Given this is experimental and we warn about it i will merge this tomorrow if noone objects

src/cleanfigure.m
```diff
@@ -1251,3 +1264,51 @@ function removeNaNs(meta, handle)
...
+                strcmp(val(i), 'y') || ...
+                strcmp(val(i), 'z'));
+    end
+end
+% ========================================================================
+function normalizeAxis(handle, cmdOpts)
+    % Normalizes data from a given axis into the interval [0, 1]
+
+    % Warn about normalizeAxis being experimental
+    warning('cleanfigure:normalizeAxis', ...
```
egeerardyn: 👍

src/cleanfigure.m
```diff
@@ -1251,3 +1264,51 @@ function removeNaNs(meta, handle)
     bool = isnumeric(val) && ~any(isnan(val)) && (isscalar(val) || numel(val) == 2);
 end
 % =========================================================================
+function bool = isValidAxis(val)
+    bool = length(val) <= 3;
+    for i=1:length(val)
+        bool = bool && ...
+               (strcmp(val(i), 'x') || ...
```
egeerardyn: `strcmp` -> `strcmpi` (I will take care of this)

PR #908, opened 2016-05-21
Pr880

egeerardyn:
This fixes #880.


egeerardyn:
@matlab2tikzBot: test this PR

src/matlab2tikz.m
```diff
@@ -1911,6 +1919,11 @@ function assertRegularAxes(handle)
             || ~abs(lineWidth-matlabDefaultLineWidth) <= m2t.tol
         lineOpts = opts_add(lineOpts, 'line width', sprintf('%.1fpt', lineWidth));
     end
+
+    % print no lines
+    if isNone(lineStyle) || lineWidth==0
```
miscco: Dont we have this isVisibleLine function`?

src/matlab2tikz.m
```diff
@@ -4082,16 +4105,27 @@ function warnFacetedInterp(m2t)
...
+        [m2t, table, tableOptions] = makeTable(m2t, '', xData, '', yData, '', zData);
+        % Print out
+        tabOpts  = opts_print(tableOptions);
+        str = sprintf('\\addplot3 [%s]\n table[%s] {%s};\n ', ...
+                         drawOpts, tabOpts, table);
+    else
+        [m2t, table, tableOptions] = makeTable(m2t, '', xData, '', yData);
+        % Print out
+        tabOpts  = opts_print(tableOptions);
+        str = sprintf('\\addplot[%s] plot table[%s] {%s};\n', ...
```
miscco: While you are at it didnt we want to use table here?

egeerardyn: I will tackle those in a separate PR: then it's just a question of search-and-replace-and-update-hashes.

src/matlab2tikz.m
```diff
@@ -1473,14 +1474,21 @@ case guitypes()
             % axis styles (e.g., colors).
             options = opts_add(options, 'separate axis lines');
         end
```
miscco: I am unsure, if the chunck below is needed or correct. 

a) Is there a possibility to change the color in matlab
b) if there is such a possibility, why always use the same color
c) As far as i know setting YColor and friends worked fine.

src/matlab2tikz.m
miscco: Fine with me :+1:

src/matlab2tikz.m
egeerardyn: Indeed we do have such a function.
However, that is for `line` objects (with markers, deviations, ...), while this relates to the lines of an `axes` object. Probably we could use that function (it should be robust enough to handle it), I'm just not sure what the best is:
 - slightly abuse that function for more readable code, or
 - do explicit checks for slightly more complex code

miscco: Nah keep it as it is. Otherwise, there is always the possibiilty that a check fails etc.

egeerardyn: 👍

src/matlab2tikz.m
egeerardyn: In #880 there are some figures to illustrate the behavior (which is where most of this code stems from).
I'm also not 100% convinced about this, but it matches MATLAB behavior better, which is what is expected from `strict` mode.

However, now I think about it: perhaps it is better to only add this fragment when the strict parameter is passed.

egeerardyn:
@matlab2tikzBot: test this PR

egeerardyn:
@matlab2tikzBot: test this PR

src/matlab2tikz.m
```diff
@@ -4491,14 +4525,19 @@ function warnFacetedInterp(m2t)
     pgfplotsColormap = sprintf('colormap={%s}{[1%s] %s}',name, unit, join(m2t, colSpecs, '; '));
 end
 % ==============================================================================
-function fontStyle = getFontStyle(m2t, handle)
+function [m2t, fontStyle] = getFontStyle(m2t, handle)
```
miscco: What i do not like here is the way we handle fontStyle. Would it be possible to use the opts array we have here too? That would make the code less special.

egeerardyn: Sorry for the late response. I think I see what you mean: this function basically translates the MATLAB font spec to LaTeX by building a string to represent it. However, that is unrelated to option arrays.
So, no, we cannot use opts arrays here.

miscco:
Damn, i just spend some hours on Thursday to find out how to display the label of an invisible axis. 
:+1: from me i think it is definitely better to add that now.

egeerardyn:
Okay, I'm merging this as-is. It is not perfect and I have noticed that e.g. on R2014b, `ACID(42)` does not compile (legends in bode plots) well, but this is probably not the fault of this PR anyway. Everything else in the graphical test (only R2014b executed, though), looks fine; and everything passes in the headless test on all environments.

In the future, I think we should clean up the handling a bit:
 - it seems the grids are explicitly colored in TikZ while the defaults should be used (this is necessary to globally override grid styles in your LaTeX document).

miscco:
The problem with bodeplots if the fucked up way it saves the legends. Therefore we have the problem that all axes get the forget plot option and pgfplots bails. See #855 and #799

egeerardyn:
I knew we had those documented somewhere! Thank you for looking those up.

Issue #913, opened 2016-06-08
missing title/legend in bode plot


Pratched:
hello again ... I noticed some issues with m2t and `bode`. When bode-plotting a transfer function in matlab and using m2t for a .tikz-file, latex does not "print" the title of the bode-plot and there are only the colored lines of the graphs visible without the description of the legend.

I checked for this issue for couple of minutes, but didnt get any general results. So if this issue is already reported and there is a solution I apologize in advance ;)
``` 
A=tf(1, [55 8]);
B=tf([3 7], 1);

bode(A, B)
grid on
legend('A', 'B');
title('example')
``` 

edit: xlabel is also missing

miscco:
So i cannot reproduce your error. There is the long standing issue with bode plots, that the "forget plot" option breaks tikz, so the original file doesnt compile. However,  if you remove all occurrences of "forget plot" then it compiles fine with title and all.

Please dont forget to check that you use the development version of m2t.

egeerardyn:
@Pratched what version of MATLAB and matlab2tikz are you using?

Pratched:
@miscco Hm alright, I actually developed this example to ensure, this is not a problem created only by my actual bode plot.
Regarding the development version: I thought the same, but changing to the original version didnt change anything.

@egeerardyn I actually downloaded tha latest one 2 weeks ago or so. I will download the version again right now and have a look. I am using Matlab R2016a (student use)

miscco:
@Pratched Please make sure, that you download the develop branch not master

Pratched:
so i should use develop branch and not master. Alright, thought the other way.

miscco:
Yes develop it is. Master is the classic 1.0 version you can download from other places as well. develop is the shiny new version that gets polished.

Pratched:
Mh. I did used the latest develop branch but no TexMaker says:
"! Missing } inserted.<inserted text>} \end{axis}" in Line 444 which would be somewhere in the tikz-file.
I try again and report...

Pratched:
Second try does not change anything. I am recieving the error given above from latex and it does not creat a pdf at all.

The develop branch I use was updated 19 days ago...

miscco:
Did you read my message stating that bode plots are currently corrupted and you need to remove the "forget plot" manually?

Pratched:
yeah, but I could figure at what "forgot plot" means. clf? If so i restarted matlab many times

PeterPablo:
[...]

PR #914, opened 2016-06-08
Fix default minor grid LineStyle

miscco:
Currently we only check whether the minorGridLS is the matlab default.

However, the one from matlab differs from that of pgfplots. Therefore it would be impossible to get dotted minor grids without manual editing.

Add a check for the matlab default minorGridLS so that we can produce dotted minor grids. While we are at it change the fallback style to that of pgfplots (solid).

This was more or less a side effect from #913

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -954,7 +954,7 @@ function m2t = drawGridOfAxes(m2t, handle)
         gridOpts = opts_new();
         % Get the line style and translate it to pgfplots
         [gridLS, isDefault] = getAndCheckDefault(...
-            'axes', handle, 'GridLineStyle', ':');
+            'axes', handle, 'GridLineStyle', '-');
         if ~isDefault || m2t.args.strict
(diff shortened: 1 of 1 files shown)


egeerardyn:
You can still get dotted grids by passing the `strict` parameter.
Personally, I'm rather a fond of having MATLAB standards being translated to pgfplots standards. Especially in old MATLAB versions, the standards are poor (dotted grid lines are just one example), while the pgfplots standards are far better.

miscco:
Hm thats true for most cases. How about we introduce a generic warning function, that hints at the use of `strict` whenever we override something?

Although that would mean, that we would have to find all those cases.

egeerardyn:
Finding all the cases where we use `strict` should not be too hard: just searching through the code for `strict` will indicate all places.

egeerardyn:
I'm closing this in favor for #918 on the one hand and the fact that the same outcome can be obtained using `strict`.

Issue #766, opened 2015-09-09
Improved warnings


egeerardyn:
This is just a minor improvement, but not that hard to implement.

Sometimes, a single run of `matlab2tikz` produces the same warning over and over again, e.g. 
```
warning: MATLAB's marker 'hexagram' not available in TikZ. Replacing by 'star'.
warning: MATLAB's marker 'hexagram' not available in TikZ. Replacing by 'star'.
warning: MATLAB's marker 'hexagram' not available in TikZ. Replacing by 'star'.
warning: MATLAB's marker 'hexagram' not available in TikZ. Replacing by 'star'.
warning: MATLAB's marker 'hexagram' not available in TikZ. Replacing by 'star'.
warning: MATLAB's marker 'hexagram' not available in TikZ. Replacing by 'star'.
warning: MATLAB's marker 'hexagram' not available in TikZ. Replacing by 'star'.
warning: MATLAB's marker 'hexagram' not available in TikZ. Replacing by 'star'.
warning: MATLAB's marker 'hexagram' not available in TikZ. Replacing by 'star'.
warning: MATLAB's marker 'hexagram' not available in TikZ. Replacing by 'star'.
```

Instead, for some warnings we should keep track if the current run has produces this warning already and don't issue them more than once.

While we're at it, we can just as well use the opportunity to introduce real `msgid` identifiers for those warnings (e.g. `matlab2tikz:hexagramReplaced`) and those `msgid`s would be a nice way to keep track of warnings that are already issued. That way, a user can switch off some warnings only.

## TODO
- [ ] identify which warnings should be one-time warnings (i.e. where the message is not parametrized)
- [ ] add a meaningful `msgid` to all warnings (one-time and others)
- [ ] make sure that one-time warnings only get produced once during execution of `matlab2tikz`

Strictly speaking this is a `+1.0.0` feature (since warning `msgid`s are public API and we cannot recover the old (non-specific) `msgid` be handled in a backwards-compatible way).

Issue #492, opened 2014-12-17
Transform LineWidth to TikZ equivalent


PeterPablo:
Currently `getLineOptions()` ignores the width of a line as long as `strict mode` is not enabled. Wouldn't it make sense to match the linewidths to the ones known by PGFplots / TikZ, e.g `ultrathin`, `thick`, etc.?
**Literature:** 
* pgfplots.pdf -> 4.7.4 Font Size and Line Width
* pgfmanual.pdf -> 15.3.1 Graphic Parameters: Line Width, Line Cap, and Line Join

egeerardyn:
@matlab2tikz/developers Could somebody weigh in on translating line widths (e.g. `3pt`) to their TikZ semantical counterparts (e.g. `thick`, `ultrathin`, ...)?
Personally, I'd leave this implementation as-is since I think we can get more wrong with such a mapping than we can add in value. 
If there is no demand for this feature, maybe we should close this issue?

okomarov:
I once tried to make sense of the marker 'business' after nidrosian's initial big effort. Two days of work later, I found to my disappointment that they do not grow linearly and that my adjustments would work only for the test size. 

This feature request has all the potential to be a second marker size issue and hence I am personally for closing it :sweat_smile:

PeterPablo:
Let's close this -- we can revive this, as soon as we run out of work :sweat_smile:

Issue #262, opened 2013-11-12
Making a list of used pgfplots styles


egeerardyn:
Instead of having each trace in a plot starting with a list of options as so

```
\addplot[color=red,mark size=1.5pt,only marks,mark=square*,mark options={solid}] { ... };
```

I'd like to have an option to output them into a style definition such as 

```
% all style definitions
\pgfplotsset{m2tstyle1/.style={color=red,mark size=1.5pt,only marks,mark=square*,mark options={solid}}}

% all plots
\addplot[m2tstyle1]{ ... };
```

which is then used in the plot. This enforces the separation between content and mark-up better (i.e. one of the basic principles of TeX). Even more so once the external data files are implemented.

A possible use-case for this would be to customize the styles by hand, depending on the output format. E.g. the same data files for a plot, but different style definitions for e.g. a black-and-white print version, plots on a poster, plots in a presentation, ...
This would be only a first step towards such an end: placing all style definitions grouped together makes it easier for the user to edit them by hand or extract them from the generated TikZ file.

As always, I'm willing to have a go at the implementation.

PR #924, opened 2016-06-28
Semantic LineWidth translations

JTSvejda:
Added variables for semantical export of lineWidth. The mapping can be specified by the user

Have fun

Signed-off-by: Jan Taro Svejda <jan.svejda@uni-due.de>


egeerardyn:
@matlab2tikzBot: test this PR

miscco:
Thanks a lot for this nice PR. As @egeerardyn commented, the code quality is really good.

What I currently do not understand is the customization of the semantics. Shouldnt they be pgfplots specific, i.e. not customizable?

src/matlab2tikz.m
```diff
@@ -1911,7 +1915,23 @@ function assertRegularAxes(handle)
     matlabDefaultLineWidth = 0.5;
     if m2t.args.strict ...
             || ~abs(lineWidth-matlabDefaultLineWidth) <= m2t.tol
-        lineOpts = opts_add(lineOpts, 'line width', sprintf('%.1fpt', lineWidth));
```
miscco: I am unsure, whether this should be part of the strict clause. Normally we make sure this 100% matches matlab so i am unsure how this fits into those custom semantics.

src/matlab2tikz.m
```diff
@@ -1911,7 +1915,23 @@ function assertRegularAxes(handle)
...
+        if m2t.args.semanticalLineWidth
+            semStrID = find(lineWidth == m2t.args.semanticalValueList);
```
miscco: You could use ismember() here

src/matlab2tikz.m
```diff
@@ -1911,7 +1915,23 @@ function assertRegularAxes(handle)
...
+            if isempty(semStrID)
+                lineOpts = opts_add(lineOpts, 'line width', sprintf('%.1fpt', lineWidth));
+            else
+                lineOpts = opts_add(lineOpts, m2t.args.semanticalStringList{semStrID}, []);
```
miscco: I think the empty array at the end is superfluous as we default to empty values anyway

miscco:
@JTSvejda What i was getting at, is that this is a really high level of sophistication, that is most likely not really suitable for most of the users. 

I see a clear benefit for using the default semantics. However, once you change them the effort is actually higher if you use those semantics, as you have to edit the tikz file and make sure the widths are correct.

That said, if someone is willing to go the extra length we should not stop them. So my suggestion would be the following.

1. Set the pgfplots default values.
2. Keep the option to change them, adding the following hint about changing default settings to the explanation of the option
```
\tikzset{
ultra thin/.style= {line width=0.10pt},
}
```

3. I would prefer to leave strict as it is and just use the else if

src/matlab2tikz.m
```diff
@@ -204,6 +204,10 @@ function matlab2tikz(varargin)
     ipp = ipp.addParamValue(ipp, 'showWarnings', true, @islogical);
     ipp = ipp.addParamValue(ipp, 'checkForUpdates', isempty(VCID), @islogical);
 
+    ipp = ipp.addParamValue(ipp, 'semanticalLineWidth', false, @islogical);
+    ipp = ipp.addParamValue(ipp, 'semanticalValueList', [.1,.2,.4,.6,.8,1.2,1.6], @isnumeric);
```
egeerardyn: I like that you have these as input argument, however, I think the input format could be improved a bit.

Right now, validity of `semanticalValueList` and `semanticalStringList` is coupled. That is not ideal. What I propose instead is to use a `semanticalLineWidths` parameter instead similar to what we have for `extraColors`.

E.g.
``` matlab
ipp = ipp.addParamValue(ipp, 'semanticalLineWidths', {'ultra thin',0.1; 'very thin', 0.2; ...}, @isValidLineWidthMapping);
```

for which the `isValidLineWidthMapping` function could be something like:
```matlab
function bool = isValidLineWidthMapping(value)
bool = iscell(value) && size(value, 2) == 2; % Nx2 cell array
bool = bool && all(cellfun(@ischar   , value(:,1)); % first column: names
bool = bool && all(cellfun(@isnumeric, value(:,2)); % second column: line width in points
```

Because that is something that can be checked easily and you can always extract both lists you had from that data structure.

src/matlab2tikz.m
```diff
@@ -1911,7 +1915,23 @@ function assertRegularAxes(handle)
...
+            semStrID = find(lineWidth == m2t.args.semanticalValueList);
+            if isempty(semStrID)
+                lineOpts = opts_add(lineOpts, 'line width', sprintf('%.1fpt', lineWidth));
+            else
+                lineOpts = opts_add(lineOpts, m2t.args.semanticalStringList{semStrID}, []);
+            end
+        else
+            lineOpts = opts_add(lineOpts, 'line width', sprintf('%.1fpt', lineWidth));
+        end
+    else % lineWidth = matlabDefaultLineWidth
```
miscco: If you drop the part in the above clause, then make this an else if

src/matlab2tikz.m
miscco: :+1: I was just about to look into that. 

While you are at it can you format it that every option has its own line, it is a really long option.

miscco: @egeerardyn I would generalize that checking function to apply for any [Name, Value] list. Basically this would just be a rename to isValidOptionList or whatever and update the comment of the second line

JTSvejda:
That was my intention too. Therefore by default you use the pgfplot-defaults and if you really like to change you are able to.

So maybe the next task is to apply it also to the axes :P

egeerardyn:
In general, I'm really happy with this. There are a few minor things that should be fixed, but it's certainly something I would love to include.

However, one thing that is really crucial to have this included, is to document it: at the top of the m2t code, you should document the new input arguments such that users can find them and know how to use it.

JTSvejda:
What about this for the simple export option:

    %   MATLAB2TIKZ('semanticalLineWidth',BOOL,...) if true it enables the export
    %   of lineWidth as semantic strings. A line with lineWidth of the list below
    %   is then replaced by a corresponding string.
    %   The mappings correspond to the pgfplot defaults:
    %   'ultra thin'=0.1pt, 'very thin'=0.2pt, 'thin'=0.4pt, 'semithick'=0.6pt, 
    %   'thick'=0.8pt, 'very thick'=1.2pt and 'ultra thick'=1.6pt

(actually I don't know how to show this in the matlab code...)

miscco:
@JTSvejda You can add matlab behind the style opening

miscco:
Please also add a similar explanation to the definition of the list. There you can add the default list as an example.

JTSvejda:
```
    %   MATLAB2TIKZ('semanticLineWidths',CELLMATRIX,...) With this parameter you
    %   are able to specify a customized list of semantic lineWidth
    %   representations used for the semanticLineWidth-BOOL option (default
    %   coresponds to pgfplot-defaults (cf. semanticLineWidth-BOOL option).
    %   Each column of the matrix needs to be a set of a string and a value!
    %   For example: {'ultra thin',0.1; 'very thin', 0.2; ...}
```

miscco:
@JTSvejda I would replace the  "parameter" with "option" and add a remark, that this only works if semanticLineWidth is set to true. Also this should be semanticLineOptions or semanticWidthListrathern than again semanticLineWidth

JTSvejda:
Maybe...
```
    %   MATLAB2TIKZ('semanticLineWidthDefinition',CELLMATRIX,...) With this option you
    %   are able to specify a customized list of semantic lineWidth representations.
    %   The default coresponds to pgfplot-defaults (cf. semanticLineWidth option).
    %   This feature is only applied if the boolean option semanticLineWidth is set to TRUE.
    %   Each column of the matrix needs to be a set of a string and a value!
    %   For example: {'ultra thin',0.1; 'very thin', 0.2; ...}
```

miscco:
So the tinniest little nitpick, i would call it semanticLineWidthList.

Also I would rather say, that this is something we should definitely considering for setting as default. thin is a better input rather than linewidth=0.2.

Also can you please make sure, that you omit the default pgfplots linewidth in the list? Generally we try to not write the default options explicitly .

JTSvejda:
Damn... That was not my intention to add this to this PR...

egeerardyn:
To fix this:
```shell
git checkout -b add-loop-to-drawAxis
git checkout develop
git reset --hard HEAD~1 
git push --force
```

Note that the last two commands are quite dangerous, don't use them lightly in other cases!

JTSvejda:
Thank you... I hope it worked.

miscco:
Seems good.

From painfull experience i would suggest that you always work on a separate branch rather than develop oranything else. There is no real other way that works

miscco:
@JTSvejda How is it going with this PR. It would be great if we could include it into the 1.1 release.

PeterPablo:
I'd like to suggest adding unit tests covering the new functionality. in my
opinion this is a very nice addition that should be enabled by default.
thanks!

src/matlab2tikz.m
```diff
@@ -116,6 +116,22 @@ function matlab2tikz(varargin)
...
+    %   'ultra thin'=0.1pt, 'very thin'=0.2pt, 'thin'=0.4pt, 'semithick'=0.6pt, 
+    %   'thick'=0.8pt, 'very thick'=1.2pt and 'ultra thick'=1.6pt 
+    % 
+    %   MATLAB2TIKZ('semanticLineWidthDefinition',CELLMATRIX,...) With this option you 
+    %   are able to specify a customized list of semantic lineWidth representations. 
+    %   The default coresponds to pgfplot-defaults (cf. semanticLineWidth option). 
+    %   This feature is only applied if the boolean option of semanticLineWidth is set to TRUE.
+    %   The list needs to be specified as a Nx2 matrix with the following
+    %   scheme:
+    %   Each column needs to be a set of one string and one numerical value!
```
miscco: Can you write pair instead of set.

src/matlab2tikz.m
```diff
@@ -1915,22 +1947,19 @@ function assertRegularAxes(handle)
...
             || ~abs(lineWidth-matlabDefaultLineWidth) <= m2t.tol
-        if m2t.args.semanticalLineWidth
-            semStrID = find(lineWidth == m2t.args.semanticalValueList);
-            if isempty(semStrID)
-                lineOpts = opts_add(lineOpts, 'line width', sprintf('%.1fpt', lineWidth));
-            else
-                lineOpts = opts_add(lineOpts, m2t.args.semanticalStringList{semStrID}, []);
-            end
+        if m2t.args.semanticLineWidth ...
+                && ismember(lineWidth, [m2t.args.semanticLineWidthDefinition{:,2}])
```
miscco: Could you put the second check into a separate if clause and add an else clause with a warning that there is no matching semantic width option

src/matlab2tikz.m
```diff
@@ -1911,7 +1947,20 @@ function assertRegularAxes(handle)
...
-        lineOpts = opts_add(lineOpts, 'line width', sprintf('%.1fpt', lineWidth));
+        if m2t.args.semanticLineWidth ...
+                && ismember(lineWidth, [m2t.args.semanticLineWidthDefinition{:,2}])
+            semStrID = lineWidth == [m2t.args.semanticLineWidthDefinition{:,2}];
+            lineOpts = opts_add(lineOpts, m2t.args.semanticLineWidthDefinition{semStrID,1}, []);
+        else
+            lineOpts = opts_add(lineOpts, 'line width', sprintf('%.1fpt', lineWidth));
+        end
+    else % lineWidth = matlabDefaultLineWidth
+        if m2t.args.semanticLineWidth
```
miscco: Here the check whether there is a matching string is missing. Please add a separate if clause and put a warning into the else (not matching string) clause that we fallback to points

src/matlab2tikz.m
```diff
@@ -1909,7 +1945,15 @@ function assertRegularAxes(handle)
     % Also apply the line width if no actual line is there; the markers make use
     % of this, too.
     matlabDefaultLineWidth = 0.5;
-    if m2t.args.strict ...
+    if m2t.args.semanticLineWidth
+        if ismember(lineWidth, [m2t.args.semanticLineWidthDefinition{:,2}])
+            semStrID = lineWidth == [m2t.args.semanticLineWidthDefinition{:,2}];
+            lineOpts = opts_add(lineOpts, m2t.args.semanticLineWidthDefinition{semStrID,1}, []);
+        else
+            warning('matlab2tikz:semanticLineWidth', 'No semantic correspondance for lineWidth of ''%f'' found. Falling back to explicit export in points.', lineWidth);
```
miscco: Please add a line break here. Maybe after the first sentence?

src/matlab2tikz.m
```diff
@@ -1909,7 +1945,15 @@ function assertRegularAxes(handle)
     % Also apply the line width if no actual line is there; the markers make use
     % of this, too.
     matlabDefaultLineWidth = 0.5;
-    if m2t.args.strict ...
+    if m2t.args.semanticLineWidth
+        if ismember(lineWidth, [m2t.args.semanticLineWidthDefinition{:,2}])
+            semStrID = lineWidth == [m2t.args.semanticLineWidthDefinition{:,2}];
+            lineOpts = opts_add(lineOpts, m2t.args.semanticLineWidthDefinition{semStrID,1}, []);
```
miscco: I think you can omit the ,[] at the end.

src/matlab2tikz.m
```diff
@@ -204,6 +220,17 @@ function matlab2tikz(varargin)
     ipp = ipp.addParamValue(ipp, 'showWarnings', true, @islogical);
     ipp = ipp.addParamValue(ipp, 'checkForUpdates', isempty(VCID), @islogical);
 
+    ipp = ipp.addParamValue(ipp, 'semanticLineWidth', false, @islogical);
+    ipp = ipp.addParamValue(ipp, 'semanticLineWidthDefinition', {...
+        'ultra thin',	0.1;...
+        'very thin',	0.2;...
+        'thin',         0.4;...
```
miscco: It seems "thin" and "thick" are not properly aligned.

miscco:
I have some fairly trivial nitpicks left but then its :+1:  from me. 

Great work!

src/matlab2tikz.m
```diff
@@ -204,6 +220,17 @@ function matlab2tikz(varargin)
...
     ipp = ipp.addParamValue(ipp, 'checkForUpdates', isempty(VCID), @islogical);
 
+    ipp = ipp.addParamValue(ipp, 'semanticLineWidth', false, @islogical);
+    ipp = ipp.addParamValue(ipp, 'semanticLineWidthDefinition', {...
+        'ultra thin',   0.1;...
+        'very thin',    0.2;...
+        'thin',         0.4;...
+        'semithick',    0.6;...
+        'thick',        0.8;...
+        'very thick',	1.2;...
```
miscco: Now these ones are off :smiling_imp:

src/matlab2tikz.m
```diff
@@ -1909,7 +1945,15 @@ function assertRegularAxes(handle)
     % Also apply the line width if no actual line is there; the markers make use
     % of this, too.
     matlabDefaultLineWidth = 0.5;
-    if m2t.args.strict ...
+    if m2t.args.semanticLineWidth
+        if ismember(lineWidth, [m2t.args.semanticLineWidthDefinition{:,2}])
+            semStrID = lineWidth == [m2t.args.semanticLineWidthDefinition{:,2}];
+            lineOpts = opts_add(lineOpts, m2t.args.semanticLineWidthDefinition{semStrID,1});
+        else
+            warning('matlab2tikz:semanticLineWidth', 'No semantic correspondance for lineWidth of ''%f'' found.\nFalling back to explicit export in points.', lineWidth);
```
miscco: I meant a line break in the code.
```matlab
warning('matlab2tikz:semanticLineWidth', ['No semantic correspondance for lineWidth of ''%f'' found. , ...
                                                                    Falling back to explicit export in points.'], lineWidth);
```

src/matlab2tikz.m
```diff
@@ -1909,7 +1945,17 @@ function assertRegularAxes(handle)
...
     % of this, too.
     matlabDefaultLineWidth = 0.5;
-    if m2t.args.strict ...
+    if m2t.args.semanticLineWidth
+        if ismember(lineWidth, [m2t.args.semanticLineWidthDefinition{:,2}])
+            semStrID = lineWidth == [m2t.args.semanticLineWidthDefinition{:,2}];
+            lineOpts = opts_add(lineOpts, m2t.args.semanticLineWidthDefinition{semStrID,1});
+        else
+            warning('matlab2tikz:semanticLineWidth',...
+                ['No semantic correspondance for lineWidth of ''%f'' found'...
```
miscco: Then point at the end of the sentence is missing. 
```
found.' ...
```

src/matlab2tikz.m
```diff
@@ -204,6 +220,17 @@ function matlab2tikz(varargin)
...
 
+    ipp = ipp.addParamValue(ipp, 'semanticLineWidth', false, @islogical);
+    ipp = ipp.addParamValue(ipp, 'semanticLineWidthDefinition', {...
+        'ultra thin',   0.1;...
+        'very thin',    0.2;...
+        'thin',         0.4;...
+        'semithick',    0.6;...
+        'thick',        0.8;...
+        'very thick',   1.2;...
+        'ultra thick',	1.6...
```
miscco: Last remaining tab

miscco:
Great work!

You will hate me, but there is one last thing I would like to ask @egeerardyn about, before we push this PR. 

The customized linewidths only work with the corresponding tikzset option, that would currently require manual editing of the tikz file. How about we add that automatically at the beginning of the tikz file?

JTSvejda:
I've created the attached example script for Matlab where one figure is exported twice, each with different options. I enabled the standalone option two have 'executable' tex-files and there everything works fine.
In the exported files is no tikzset option... so are you sure you the need this option somewhere?

[example.zip](https://github.com/matlab2tikz/matlab2tikz/files/356937/example.zip)

miscco:
@JTSvejda What i meant, is that if the user defines a custom semanticLineWidthDefinition this definition will not be automatically put into the tikz picture but has to be added manually (as the code stands right now).

If he only defined new options, this would compile but lead to wrong linewidths, if he defined custom options like "super thick" then it wouldnt even compile out of the box.

We have options to add custom code to various places in the tikzfigure, so I would suggest, that if semanticLineWidthDefinition is defined we add a
```tex
\tikzset{
semanticLineWidthDefinition{1,1}/.style= {line width=num2str(semanticLineWidthDefinition{1,2})pt},
}
```
at the beginning of the file automatically.

JTSvejda:
Ah okay... now I understand you.

But actually if this is added to the code by default, then it's a need for the user to edit each picture to tune his manually defined line width (like "super thick"). There should be a at least a comment in the options documentation and maybe a message while the export of matlab2tkiz.

miscco:
@JTSvejda Thats exactly why i would prefer, if we would automatically add the above tikzset to any custom definition of semanticLineWidth

JTSvejda:
@miscco Maybe I misunderstand you, but with this option automatically added each single tikzpicture-file contains a definition of the line widths. So the same line width (e.g. "super thick") is defined in all tikzpicture-files. 
In this case, if you like to change to change "super thick", you have to change it in all tikzpictures by hand or by re-exporting with matlab2tikz (where you have to make the changes in Matlab before export).

In my opinion the best way is one definition for the whole latex document (maybe in the preamble). Then you only have to change one definition to change "super thick" to another value.

Therefore I thought of a warning massage during the matlab2tikz export, if the user defined his own (not pgfplot default) linewidths. Such that he should not forget to make his linewidth definitions.

miscco:
@JTSvejda I think this depends on the prefered workflow. As far as I know a large number of the users directly includes those figures into their latex documents without the standalone option. Therefore, it is better to define a non-standard definition inside the respective tikz picture environement it is used in.

This would always work and automatically updated with every run of m2t. On the other hand if you edit it manually there is no drawback either way.

So to sum up what I would suggest is the following,

1. Make it the default and remove the option to turn it off. I cannot imagine any user who would prefer the old way (put the code in the else clause of the strict check). Also remove the flag option. 

2. Add a function, that parses semanticLineWidthDefinition and sets a flag if the provided one contains non default options.

3. If that flag is set, add the corresponding tikzset to the 'extraTikzpictureOptions'

I think @PeterPablo was already in favor of setting this to defalt, so would like to know @egeerardyn opinion before you put too much work in it.

egeerardyn:
@miscco 

> The customized linewidths only work with the corresponding `tikzset` option, that would currently require manual editing of the tikz file. How about we add that automatically at the beginning of the tikz file?
>
> edit:
>
> I am looking at 'extraTikzpictureOptions' and 'extraCode' options

I am actually in favor of **not** including the `tikzset` command within the produced figures. As far as I'm concerned, overriding those semantic line widths is a functionality for advanced users who should take care of their own dependencies (in whatever way they prefer). In the same way: things in `extraCode`, `width`, `height`, the extra colotd, ... are also a responsibility of the user to properly set in their document. Without proper care, `standalone` also fails for those settings.

I completely follow the reasoning of @JTSvejda in that regard, and I even think that it is *harmful* to include the `\tikzset` commands within each exported figure: that disallows one to override those settings for a whole document. At that point, semantic translation becomes a pointless level of redirection/abstraction since one cannot overload the behavior (which, IMHO, is the whole point of such abstractions in the first place).

> @JTSvejda I think this depends on the prefered workflow. As far as I know a large number of the users directly includes those figures into their latex documents without the standalone option. Therefore, it is better to define a non-standard definition inside the respective tikz picture environement it is used in.
>
> This would always work and automatically updated with every run of m2t. On the other hand if you edit it manually there is no drawback either way.

You are right that it depends on the workflow.
Personally, I think most people use m2t without a lot of special options as many of our users are not familiar with the underlying `pgfplots` library. As such, they have nothing to gain from these semantic line widths.
IMHO, this is a feature for those users who are familiar with `pgfplots`. For that kind of users, I think it doesn't help to provide a lot of "magic" in the feature, as those advanced users may have very peculiar workflows that we may not envision.

Honesty compels me to say that my personal workflow with m2t relies heavily on abstractions and hence I might be severely biased in favor of such a workflow. I typically use m2t to automatically convert all the options of a figure, but I then extract all the option arrays to an external file (often `styles.inc.tex`) such that I have global control of the appearance of a document. Those `\tikzset` statements would do no good for me. See also #262 for some more context.

### TL;DR
I wouldn't add all the complexity of checking defaults and so on and putting the relevant `\tikzset`s in the output. 
Instead: write in the documentation that it's the responsibility of the user to define those styles if needed.

egeerardyn:
After reading the discussion in #492 again (which covers exactly this feature), I think we should not make those semantic translations the default behavior. It's really fine as an option, but probably not as default behavior.

PeterPablo:
Just my brief input. I didn't actually run any code. Thank you @egeerardyn for digging up my old issue -- honestly, I had totally forgotten about it. I reread the whole conversation of one and a half years ago. Today, I still follow your argumentation, though "naively" still tend towards option 2, anticipating few irritated users.
My vote is for enabling this setting by default. `Strict` mode should fall back to specific line width definitions. No tikzset commands should be generated, as they obfuscate the intention of abstraction.

(For my thesis I took an approach that allowed mixing `standalone` mode with the general inclusion in my main file (using `externalize`) through an `if` clause that checked for the existence of a tex variable. The global preamble takes care for the general compilation and the local preamble takes over for standalone compilation. This isn't strictly related to this discussion - I am just saying that I did experiment with my workflow quite a bit as well. I don't see strong reasons against going forward as suggested.)

src/matlab2tikz.m
```diff
@@ -1909,7 +1945,17 @@ function assertRegularAxes(handle)
     % Also apply the line width if no actual line is there; the markers make use
     % of this, too.
     matlabDefaultLineWidth = 0.5;
-    if m2t.args.strict ...
+    if m2t.args.semanticLineWidth
+        if ismember(lineWidth, [m2t.args.semanticLineWidthDefinition{:,2}])
+            semStrID = lineWidth == [m2t.args.semanticLineWidthDefinition{:,2}];
+            lineOpts = opts_add(lineOpts, m2t.args.semanticLineWidthDefinition{semStrID,1});
+        else
+            warning('matlab2tikz:semanticLineWidth',...
```
egeerardyn: I don't think this warning is necessary.

JTSvejda:
@egeerardyn Actually the warning is not necessary. But if a user expected to replace some widths and wonders that matlab2tikz isn't doing it, then this warning is a good hint.

egeerardyn:
@PeterPablo You are probably very right: the semantic translation can happen by default and should not interfere with normal usage.

My *personal* preference is having raw line widths, but I have no rational arguments in favor for that (probably I'm just used to having them specified that way and since I usually have my own abstractions, I don't often use those `thick` styles). And I don't think that matlab2tikz should cater to my preferences, but to usability for many people and their different workflows.

### Summary
I would suggest to do only step 1 of what @miscco suggested, with some minor modifications:

> Make it the default and remove the option to turn it off. I cannot imagine any user who would prefer the old way (put the code in the else clause of the strict check). Also remove the flag option.

My own remark on this: if the `semanticLineWidths` option is empty, I think that is signal enough that no semantic translation can happen. I.e. that would allow access to the old way of concrete line widths. I think there is also no need to make this feature behave specially for the `strict` mode: if people want concrete widths, they should keep the spec empty and that's that.

egeerardyn:
@JTSvejda I agree with you on the warning and its usefulness. 

I'm just a bit worried about spamming the terminal of our users with such warnings (because if I have a plot with 20 lines each of a custom width, I get 20 of those messages and a user that feels overwhelmed by the output).

I know it is possible to turn off such warnings, but I think that many of our users are not aware of that (hence why we also have the `userWarn` thingie next to the classical `warning off <MSGID>` way).

Can you place a `%TODO: make sure this only warns once #766` next to it? That way, we can remember to treat it as well when we do #766 and leave the warning in place for the time being.

miscco:
@egeerardyn How about a compromise. I actually find it strange to expect the user to provide the default values.

I would propose, that we add a function that modifies the default list when appropriate, so i does not have to be included manually. Obviously this would enable it by default, however i think this is definitely an improvement.
```matlab
function list = createSemanticWidthList(m2t)
  % Defines the default semantic options of pgfplots and updates it when applicable

  % pgfplots defaults
  list = {'ultra thin',  0.1;...
          'very thin',   0.2;...
          'thin',        0.4;...
          'semithick',   0.6;...
          'thick',       0.8;...
          'very thick',  1.2;...
          'ultra thick', 1.6...
          }

  % Update defaults of append the user provided setting
  for i=1:numel(m2t.args.semanticLineWidthDefinition{:,1})
    % Check for redefinitions of defaults
    if ismember(m2t.args.semanticLineWidthDefinition{i, 1}, list{:,1})
      list{i, 2} = m2t.args.semanticLineWidthDefinition{i, 2};
    else
      list{end+1} = m2t.args.semanticLineWidthDefinition{i, :};
    end
  end
end
```

src/matlab2tikz.m
```diff
@@ -428,6 +455,15 @@ function matlab2tikz(varargin)
     bool = iscell(colors) && all(cellfun(isValidEntry, colors));
 end
 % ==============================================================================
+function bool = isValidSemanticLineWidthDefinition(defMat)
+    % Returns true when the input is a cell array of schape Nx2 and
```
miscco: typo schape->shape

miscco:
Regarding the warning, I would say, that with the function proposed above we can trigger the warning only when an option was provided without triggering it.

egeerardyn:
@miscco Oh, you're talking about having to pass in the default values as an input argument. I agree with you that it is inconvenient.

I think that we can make another compromise that allows for both use cases (easily adapting the mapping ánd removing the mapping). If we allow the value `NaN` as input argument to signal to clean the semantic map, then both use cases can be satisfied with minimal annoyance.

I think one of the reasons why I am quite fond of the explicit line widths, is that eventually this may help us write unit tests (e.g. such that we  can say: if you are converting a `line('LineWidth',5)`to TikZ, the output should contain `line width=5pt`). That mapping makes testing a bit harder (but is convenient for users).

miscco:
@egeerardyn I wouldnt say that it makes testing harder. In the end you have the range of positive real numbers except 7 defined values. That still leaves quite some room for testing :smile: 

I have adopted the function above to allow switching it off (empty list) one would also have to add a check for isnan in the isValidSemantic check

miscco:
@egeerardyn Given that this PR is already a nice improvement and contained behind two guards should we merge this PR now and modify the code afterwards. 

In the end i dont know how much time @JTSvejda can muster for our nits.

egeerardyn:
Okay, I will have a go at merging this tonight with your changes included. We can clean it up further if we take the time later on.

PR #928, opened 2016-07-08
Routine to switch axis ruler visibility

JTSvejda:
Added a loop to the drawAxes() function to add the axis property 'axis <xyz> line=none' if the visibility of the <xyz>-axis ruler is set to 'off'.

Best Regards
JTSvejda

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -922,6 +922,14 @@ function m2t = drawAxes(m2t, handle)
     m2t = drawBoxAndLineLocationsOfAxes(m2t, handle);
     m2t = drawGridOfAxes(m2t, handle);
     m2t = drawLegendOptionsOfAxes(m2t);
+    
+    % If the axis ruler is non-visible draw no axes line
+    for axis = 'xyz'
+        if strcmp(get(get(handle, [axis 'Axis']), 'Visible'), 'off')
+        m2t.axes{end}.options = opts_append_userdefined(m2t.axes{end}.options, ...
+            ['axis ' axis ' line=none']);
+        end
+    end
 
     m2t.axes{end}.options = opts_append_userdefined(m2t.axes{end}.options, ...
                                                   m2t.args.extraAxisOptions);


okomarov:
@JTSvejda Thanks for this PR. I took the liberty to edit your post to add the example, as it wasn't completely clear that we wanted to toggle single axis. 

As a side note, this might be a good opportunity to encapsulate axis visibility into a function, with the overall visibility, checked a bit earlier in the code, also part of this function. Moreover, we would need to test for Octave vs Matlab, as I do not think that the former has this option.

miscco:
@okomarov I would second that.

src/matlab2tikz.m
```diff
@@ -922,6 +922,14 @@ case guitypes()
     m2t = drawBoxAndLineLocationsOfAxes(m2t, handle);
     m2t = drawGridOfAxes(m2t, handle);
     m2t = drawLegendOptionsOfAxes(m2t);
+
```
miscco: I think this should be moved above to the following clause

```matlab
if ~isVisible(handle)
```
and an elseif. I would prefer if hide axis is not accompanied with 'axis xyz line = 0'

JTSvejda:
I hope this is the solution to your comment...

src/matlab2tikz.m
```diff
@@ -914,15 +914,24 @@ case guitypes()
...
+    
     m2t.axes{end}.name = 'axis';
 
     m2t = drawBackgroundOfAxes(m2t, handle);
     m2t = drawTitleOfAxes(m2t, handle);
     m2t = drawBoxAndLineLocationsOfAxes(m2t, handle);
     m2t = drawGridOfAxes(m2t, handle);
     m2t = drawLegendOptionsOfAxes(m2t);
-
+
```
miscco: That spurios newline should go away

miscco:
So does this also account for axis ticks and ticklabels?

JTSvejda:
I've tested it with the attached example and the ticks and ticklabels are non-visible if the axis ruler is non-visible.

[example_axisRulerVisibility.zip](https://github.com/matlab2tikz/matlab2tikz/files/357002/example_axisRulerVisibility.zip)

okomarov:
This should go at the beginning of function `getAxisOptions()` and if the axis is not drawn, return early.

PR #929, opened 2016-07-08
Only mkdir when directory name is specified

egeerardyn:
MATLAB warns that `mkdir('')` may stop working in the future.
As such, only do the `mkdir` part when an actual directory is provided.

This PR makes sure that this warning no longer occurs.


miscco:
This is an obvious improvement. :+1:

Issue #931, opened 2016-07-14
legend issue using `quiver` for phasordiagramm


Pratched:
Hello again ;)

Since I am using m2t a lot by now, I need to draw phasor diagramms this time. Anything Is working quite fine, untill I set legend in Matlab.

```

**Comment by miscco** [comment]
Actually your plot is not rendering correctly. What is happening, is that the scaling algorithm used for the size of the Arrows is failing for the legend (due to division by zero). Therefore your legend only contains the line without the arrow.

I am currently unsure how to resolve this.

**Comment by Pratched** [comment]
I am sorry,  actually it should be `x= 1+1i;
y= -1-1i;` but this doesnt really matter....

You are right, the arrows arent as long as they should be. I did notice this before, but didnt get any errors, because I didnt used `legend`. For now its alright,...But I dont get where there is a division by zero....

**Comment by miscco** [comment]
I think #932 should be able to handle this. I am not sure if it is easily applicable to your setup but i changed the plot options to:

```latex
\addplot[-{Straight Barb[angle'=18.263,scale={10/1000*\pgfplotspointmetatransformed}]},color=mycolor1,point meta={sqrt((\thisrow{u})^2+(\thisrow{v})^2)},point meta min=0,quiver={u=\thisrow{u},v=\thisrow{v}},legend image post style=-Straight Barb]
```

PR #932, opened 2016-07-14
Rework the handling of arrows in quiver plots 

miscco:
Currently there is a problem with the scaling of the arrow heads in quiverplots with legend entries. See #931 

To workaround this issue use an unscaled arrowhead in the legend entries. Furthermore rework the way arrowheads are handled to simplify the code.

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -4300,7 +4300,14 @@ function [m2t, str] = drawQuiverGroup(m2t, h)
 
         userInfo(m2t, ['Please change the "arrowHeadSize" option', ...
             ' if the size of the arrows is incorrect.']);
-        arrowHeadSize = sprintf(m2t.ff, abs(m2t.args.arrowHeadSize));
+        arrowHeadSize = abs(m2t.args.arrowHeadSize);
+
+        % Rescale the head size if MaxHeadSize is changed
+        [headSize, isDefault] = getAndCheckDefault('quiver', h, 'MaxHeadSize', 0.2);
+        if ~isDefault
+             arrowHeadSize = arrowHeadSize * headSize / 0.2;
+        end
+        arrowHeadSize = sprintf(m2t.ff, arrowHeadSize);
 
         % Write out the actual scaling for TikZ.
         % `\pgfplotspointsmetatransformed` is in the range [0, 1000], so
@@ -4327,6 +4334,7 @@ function [m2t, str] = drawQuiverGroup(m2t, h)
     quiverOptions = opts_new();
     quiverOptions = opts_add(quiverOptions, 'u', '\thisrow{u}');
     quiverOptions = opts_add(quiverOptions, 'v', '\thisrow{v}');
+
     if is3D
         quiverOptions = opts_add(quiverOptions, 'w', '\thisrow{w}');
         arrowLength = '{sqrt((\thisrow{u})^2+(\thisrow{v})^2+(\thisrow{w})^2)}';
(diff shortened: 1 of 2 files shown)


egeerardyn:

miscco:
@egeerardyn I think we will have to look at the graphical /text output, as this changes the way quiverplots are generated.

src/matlab2tikz.m
```diff
@@ -4235,11 +4208,39 @@ function warnFacetedInterp(m2t)
...
+    else
+        arrowLength = '{sqrt((\thisrow{u})^2+(\thisrow{v})^2)}';
     end
+    plotOptions = opts_add(plotOptions, 'point meta', arrowLength);
+    plotOptions = opts_add(plotOptions, 'point meta min', '0');
+
     plotOptions = opts_addSubOpts(plotOptions, 'quiver', quiverOptions);
 
+    % Check whether there is a legend and add a legend image option without scaling
+    if ~isempty(m2t.legendHandles)
```
egeerardyn: Is there a reason why you check for `m2t.legendHandles` and do not use the function `hasLegendEntry` instead?

egeerardyn:
Oh, then I think I see why this is off. There was a tiny bug in the `guess***` function such that on Octave (Travis and 1 env on Jenkins) this still passes the tests.

Code looks fine by me, except for that one tiny remark (either we should try it with the proper function or comment in the code why that rough check is used).

Once graphical testing has passed, I think this is good to go :+1:

src/matlab2tikz.m
miscco: I changed it to 
```
m2t.currentHandleHasLegend
```

which is the option that hasLegendEntry actually sets

miscco:
Hm Smoething is strange, now the arrow size seems the same for ACID22

miscco:
At a second glance it isnt really strange at all. The option we are changing isn't even considered (MaxHeadSize)

miscco:
I have added a check, that scaled the head size with the maxHeadSize. This should lead to the desired outcome at least for ACID22

src/matlab2tikz.m
```diff
@@ -4327,7 +4300,14 @@ function warnFacetedInterp(m2t)
 
         userInfo(m2t, ['Please change the "arrowHeadSize" option', ...
             ' if the size of the arrows is incorrect.']);
-        arrowHeadSize = sprintf(m2t.ff, abs(m2t.args.arrowHeadSize));
+        arrowHeadSize = abs(m2t.args.arrowHeadSize);
+
+        % Rescale the head size if MaxHeadSize is changed
+        [headSize, isDefault] = getAndCheckDefault('quiver', h, 'MaxHeadSize', 0.2);
```
egeerardyn: Maybe comment where the `0.2` comes from (something like `defaultHeadSize = 0.2;` could be enough).

egeerardyn:
@miscco Have you had any time to have a go at this any further? If not, I would suggest I push back this  PR to a 1.1.1/1.2 release and that I continue preparing the 1.1 release over the weekend (in a private branch)

miscco:
I think we have to close/postpone this PR. The problem is, that the current approach does not scale the arrows correctly. I filed a bug report on pgfplots, but didnt get any feedback yet.

egeerardyn:
Okay, that's fine. If you used the official pgfplots bug tracker, could you link it back to here (or the #931 issue), such that we can keep track of it in the future?

miscco:
Here is the link to the corresponding bugreport for pgfplots:
https://sourceforge.net/p/pgfplots/bugs/126/

PR #936, opened 2016-07-20
Fix small syntax issues in output

egeerardyn:
This PR will address the last minor issues in the output formatting for m2t version 1.1.
Please see the commit messages for details of each run (since this is quite heavy on hash changes).


test/suites/ACID.Octave.3.8.0.md5
```diff
@@ -1,9 +1,9 @@
 alphaImage : a51ff2c6514fb5606ee2fcb732edc7ea
 alphaTest : bd2d78b8d373342d5285132b607a2f46
-annotationAll : 567c721ff42533ad9aa33546b1002e0f
```
egeerardyn: So apparently, these annotation stuff behaves differently on Octave 4.0 and 3.8. If I remember correctly, those are not supported in 3.8 (so it's a bit weird that we cannot use hashes from 3.8 to validate in 4.0, but hashes from 4.0 just validate fine in 3.8). I'm a bit baffled by this (I don't see those tests being marked `unreliable` by 3.8, though).

Issue #933, opened 2016-07-14
Manual positioning of colorbar isn't possible


ulijh:
Hi all,

im trying to use one single colorbar in a figure with multiple plots. Like in this post: [https://de.mathworks.com/matlabcentral/answers/144453-how-to-make-one-colorbar-for-all-subplots](url)

Unfortunately the colorbar doesn't show up in the tikz figure and matlab2tikz prints the following error:

```
Error using matlab2tikz>getColorbarPosOptions (line 4216)
getColorbarOptions: Unknown 'Location' manual.

Error in matlab2tikz>getColorbarOptions (line 4085)
    [cbarTemplate, cbarStyleOptions] = getColorbarPosOptions(handle, ...

Error in matlab2tikz>handleColorbar (line 1186)
            getColorbarOptions(m2t, handle));

Error in matlab2tikz>saveToFile (line 455)
        m2t = handleColorbar(m2t, cbar);

Error in matlab2tikz (line 352)
saveToFile(m2t, fid, fileWasOpen);
```

It would be very helpful for me to have this option to position a colorbar manually.

Thanks,
Uli

ulijh:
I got it to work using quite a dirty hack in matlab2tikz.m. Well, only the case of a vertical colorbar with axis to the right.
In function `getColorbarPosOptions(handle, cbarStyleOptions)` in the `switch lower(loc)` environment I added the case 'manual':

```
        case 'manual'
            cbarTemplate = 'right';
            origUnits = handle.Units;
            handle.Units = 'centimeters';
            cbarStyleOptions = opts_add(cbarStyleOptions, 'at',...
                sprintf('{(%fcm,%fcm)}',handle.Position(1),...
                handle.Position(2)+handle.Position(4)));
            cbarStyleOptions = opts_add(cbarStyleOptions, 'height',...
                sprintf('%fcm',handle.Position(4)));
            handle.Units = origUnits;
```
May be somebody with some more experience in matlab2tikz can take this and make it an actual feature.

Thanks, Uli

miscco:
@ulijh I had a look at it but forgot to anwser here.

The problem is, that is is not possible to determine the placement generically. While we can get the position from it the orientation seems tricky.

ulijh:
@miscco I see the problem, If i find the time, I'll try to come up with something. Could you give me some advice on how convert matlab positions to proper tikz positions in a m2t style?

miscco:
[...]

PR #937, opened 2016-07-21
Manual colorbar location

ulijh:
Hi,
I might have found a solution for manual positioning of colorbars with new Matlab versions, see issue #933.
The (undocumented) `Axes` property of a colobar provides a handle for it's associated axes. Comparing the axes' and the colorbar's posititions along with the `AxisLocation` property, we can decide wheather to use the colorbar template "left" or "right".

The following example works for me:
```
ax = subplot(221);
cb = colorbar('Position',[ax.Position(1)+0.3*ax.Position(3),ax.Position(2),0.02,0.2],...
    'AxisLocation','in');
ax = subplot(222);
cb = colorbar('Position',[ax.Position(1)+0.7*ax.Position(3),ax.Position(2),0.02,0.2],...
    'AxisLocation','in');
ax = subplot(223);
cb = colorbar('Position',[ax.Position(1)+0.3*ax.Position(3),ax.Position(2),0.02,0.2],...
    'AxisLocation','out');
ax = subplot(224);
cb = colorbar('Position',[ax.Position(1)+0.7*ax.Position(3),ax.Position(2),0.02,0.2],...
    'AxisLocation','out');

print -dpng cbLocManual.png
matlab2tikz('cbLocManual.tex','standalone',true)
```

Note that setting colorbar positions manually sets it's `Location` to `manual`. In this case only vertical colorbars are suported in Matlab (Documentation of R2016a).

What do you think?
Cheers, Uli

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -4825,8 +4825,17 @@ function [cbarTemplate, cbarStyleOptions] = getColorbarPosOptions(handle, cbarSt
             cbarStyleOptions = opts_add(cbarStyleOptions,...
                 'xticklabel pos', 'upper');
         case 'southoutside'
-
             cbarTemplate = 'horizontal';
+        case 'manual'
+            cbarTemplate = 'right';
+            origUnits = handle.Units;
+            handle.Units = 'centimeters';
+            cbarStyleOptions = opts_add(cbarStyleOptions, 'at',...
+                sprintf('{(%fcm,%fcm)}',handle.Position(1),...
+                handle.Position(2)+handle.Position(4)));
+            cbarStyleOptions = opts_add(cbarStyleOptions, 'height',...
+                sprintf('%fcm',handle.Position(4)));
+            handle.Units = origUnits;
         otherwise
             error('matlab2tikz:getColorOptions:unknownLocation',...
                 'getColorbarOptions: Unknown ''Location'' %s.', loc)
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -4830,48 +4830,50 @@ function [cbarTemplate, cbarStyleOptions] = getColorbarPosOptions(handle, cbarSt
             origUnits = get(handle,'Units');
             assocAxes = get(handle,'Axes');
             origAxesUnits = get(assocAxes,'Units');
-            set(handle,'Units','centimeters');        % We need absolute
-            set(assocAxes,'Units','centimeters');     % positions for tikz
-            cbarPos = get(handle,'Position');
-            cbarAxesPos = get(assocAxes,'Position');
+            set(handle,'Units','centimeters');        % Make sure we have
+            set(assocAxes,'Units','centimeters');     % same units
+            cbarDim = pos2dims(get(handle,'Position'));
+            cbarAxesDim = pos2dims(get(assocAxes,'Position'));
             set(handle,'Units',origUnits);            % Restore original
             set(assocAxes,'Units',origAxesUnits);     % units
 
-            % Cases of colorbar axis locations (in or out) depending on position
-            % of colorbar relative to it's associated axes.
+            center = @(dims) (dims.left + dims.right)/2;
+            centerCbar = center(cbarDim);
+            centerAxes = center(cbarAxesDim);
+
+            % Cases of colorbar axis locations (in or out) depending on center
+            % of colorbar relative to the center it's associated axes.
             % According to matlab manual (R2016a) colorbars with Location 'manual'
             % can only be vertical.
-            center = @(dims) (dims.left + dims.right)/2;
-            centerCbar = center(pos2dims(cbarPos));
-            centerAxes = center(pos2dims(cbarAxesPos));
             axisLoc = get(handle,'AxisLocation');
             if centerCbar < centerAxes
                 if strcmp(axisLoc,'in')
                     cbarTemplate = 'right';
-                    cbarStyleOptions = opts_add(cbarStyleOptions, 'anchor',...
-                        'south east');
                 else
                     cbarTemplate = 'left';
-                    cbarStyleOptions = opts_add(cbarStyleOptions, 'anchor',...
-                        'south west');
                 end
             else
                 if strcmp(axisLoc,'in')
                     cbarTemplate = 'left';
-                    cbarStyleOptions = opts_add(cbarStyleOptions, 'anchor',...
-                        'south west');
                 else
                     cbarTemplate = 'right';
-                    cbarStyleOptions = opts_add(cbarStyleOptions, 'anchor',...
-                        'south east');
                 end
             end
+
+            % Using positions relative to associated axes
+            calcRelPos = @(pos1,pos2,ext2) (pos1-pos2)/ext2; 
+            cbarRelPosX = calcRelPos(cbarDim.left,cbarAxesDim.left,cbarAxesDim.width);
+            cbarRelPosY = calcRelPos(cbarDim.bottom,cbarAxesDim.bottom,cbarAxesDim.height);
+            cbarRelHeight = cbarDim.height/cbarAxesDim.height;
             
+            cbarStyleOptions = opts_add(cbarStyleOptions, 'anchor',...
+                'south west');
             cbarStyleOptions = opts_add(cbarStyleOptions, 'at',...
-                ['{(' formatDim(cbarPos(1), 'cm') ','...
-                      formatDim(cbarPos(2), 'cm') ')}']);
+                ['{(' formatDim(cbarRelPosX) ','...
+                      formatDim(cbarRelPosY) ')}']);
             cbarStyleOptions = opts_add(cbarStyleOptions, 'height',...
-                ['{(' formatDim(cbarPos(4), 'cm') ')}']);
+                [formatDim(cbarRelHeight),...
+                '*\pgfkeysvalueof{/pgfplots/parent axis height}']);
 
         otherwise
             error('matlab2tikz:getColorOptions:unknownLocation',...
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -4827,21 +4827,26 @@ function [cbarTemplate, cbarStyleOptions] = getColorbarPosOptions(handle, cbarSt
         case 'southoutside'
             cbarTemplate = 'horizontal';
         case 'manual'
-            origUnits = handle.Units;
-            origAxesUnits = handle.Axes.Units;
-            handle.Units = 'centimeters';        % We need absolute
-            handle.Axes.Units = 'centimeters';   % positions for tikz
-            cbarPos = handle.Position;
-            cbarAxesPos = handle.Axes.Position;
-            handle.Units = origUnits;
-            handle.Axes.Units = origAxesUnits;
+            origUnits = get(handle,'Units');
+            assocAxes = get(handle,'Axes');
+            origAxesUnits = get(assocAxes,'Units');
+            set(handle,'Units','centimeters');        % We need absolute
+            set(assocAxes,'Units','centimeters');     % positions for tikz
+            cbarPos = get(handle,'Position');
+            cbarAxesPos = get(assocAxes,'Position');
+            set(handle,'Units',origUnits);            % Restore original
+            set(assocAxes,'Units',origAxesUnits);     % units
 
             % Cases of colorbar axis locations (in or out) depending on position
             % of colorbar relative to it's associated axes.
             % According to matlab manual (R2016a) colorbars with Location 'manual'
             % can only be vertical.
-            if cbarPos(1) + 0.5*cbarPos(3) < cbarAxesPos(1) + 0.5*cbarAxesPos(3) 
-                if strcmp(handle.AxisLocation,'in')
+            center = @(dims) (dims.left + dims.right)/2;
+            centerCbar = center(pos2dims(cbarPos));
+            centerAxes = center(pos2dims(cbarAxesPos));
+            axisLoc = get(handle,'AxisLocation');
+            if centerCbar < centerAxes
+                if strcmp(axisLoc,'in')
                     cbarTemplate = 'right';
                     cbarStyleOptions = opts_add(cbarStyleOptions, 'anchor',...
                         'south east');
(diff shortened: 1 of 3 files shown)


egeerardyn:
@matlab2tikzBot: test this PR

src/matlab2tikz.m
```diff
@@ -4825,8 +4825,48 @@ function warnFacetedInterp(m2t)
             cbarStyleOptions = opts_add(cbarStyleOptions,...
                 'xticklabel pos', 'upper');
         case 'southoutside'
-
             cbarTemplate = 'horizontal';
+        case 'manual'
+            origUnits = handle.Units;
```
egeerardyn: Please use `get(handle,'Units')` syntax (and so on)
I know it looks uglier, but it is backwards compatible to Octave and MATLAB 2014a and older.

src/matlab2tikz.m
```diff
@@ -4825,8 +4825,48 @@ function warnFacetedInterp(m2t)
...
+            cbarPos = handle.Position;
+            cbarAxesPos = handle.Axes.Position;
+            handle.Units = origUnits;
+            handle.Axes.Units = origAxesUnits;
+
+            % Cases of colorbar axis locations (in or out) depending on position
+            % of colorbar relative to it's associated axes.
+            % According to matlab manual (R2016a) colorbars with Location 'manual'
+            % can only be vertical.
+            if cbarPos(1) + 0.5*cbarPos(3) < cbarAxesPos(1) + 0.5*cbarAxesPos(3)
```
egeerardyn: If would write this calculation using intermediate variables:

```matlab
center = @(dims) (dims.left + dims.right)/2;
centerCbar = center(pos2dims(cbarPos));
centerAxes = center(pos2dims(cbarAxesPos));
```

to make it more readable.

src/matlab2tikz.m
```diff
@@ -4825,8 +4825,48 @@ function warnFacetedInterp(m2t)
...
+                    cbarStyleOptions = opts_add(cbarStyleOptions, 'anchor',...
+                        'south west');
+                else
+                    cbarTemplate = 'right';
+                    cbarStyleOptions = opts_add(cbarStyleOptions, 'anchor',...
+                        'south east');
+                end
+            end
+            
+            cbarStyleOptions = opts_add(cbarStyleOptions, 'at',...
```
egeerardyn: Please have a look at the `formatDim` function to print the coordinates:
`formatDim(cbarPos(...), 'cm')` instead of `sprintf('%fcm', cbarPos(...))`.

That way, we have uniform control over how dimensions/coordinates are displayed and what precision is used.

egeerardyn:
Thank you for this pull request.

There are a few small things, I would like to point out:
 - please use the [recommended git workflow](https://github.com/matlab2tikz/matlab2tikz/wiki/Recommended-git-workflow) in the future (i.e. first make a feature branch before making a pull request): that makes your life easier (and hence also our lives)
 - there are a few small things (see above) that could be changed to fit better in our coding style and make use of functionality we have in place for some common tasks.

In general, I really like the quality of your code. I will try to play with it on my own computer to see how robust it is.

ulijh:
@egeerardyn Thanks for your comments! I'll put them in (and use a feature branch next time...)

ulijh:
Hi, I just did the coding style things. There are 2 things I'm not really happy with but my m2t and pgf/tikz knowledge isn't sufficient to properly solve them:

1. Width of colorbar: When adding the width option to tikz colobar only the axis of the colobar scales, not the colors... So currently it simply uses the default width which seems reasonable. I could live with this limitation.
2. The positioning of the colorbar in x-direction is a little bit off. Is there a proper m2t-style way to convert the matlab positions to proper tikz positions? Something like `getAxesPosition` or similar? This example shows the issue: [cbLocManual.pdf](https://github.com/matlab2tikz/matlab2tikz/files/378455/cbLocManual.pdf) (matlab code below).
The colobars south west corner should be at (0.2,0.2) and (0.8,0.2) on the plots to the right.
```
ax = subplot(221);
cb = colorbar('Position',[ax.Position(1)+0.2*ax.Position(3),...
    ax.Position(2)+0.2*ax.Position(4),0.02,0.2],...
    'AxisLocation','in');
grid on
ax = subplot(222);
cb = colorbar('Position',[ax.Position(1)+0.8*ax.Position(3),...
    ax.Position(2)+0.2*ax.Position(4),0.02,0.2],...
    'AxisLocation','in');
grid on
ax = subplot(223);
cb = colorbar('Position',[ax.Position(1)+0.2*ax.Position(3),...
    ax.Position(2)+0.2*ax.Position(4),0.02,0.2],...
    'AxisLocation','out');
grid on
ax = subplot(224);
cb = colorbar('Position',[ax.Position(1)+0.8*ax.Position(3),...
    ax.Position(2)+0.2*ax.Position(4),0.02,0.2],...
    'AxisLocation','out');
grid on

matlab2tikz('cbLocManual.tex','standalone',true)
```
Any ideas? Thanks, Uli

egeerardyn:
Normally, you should be able to set the `width` like in the figure below (this comes from the `pgfplots` manual):

screen shot 2016-07-22 at 17 30 06

The positioning is not perfect, but it is almost right, I would say. By also specifying the `anchor` (probably `south west`) for the color bar, you can get better results.

Something else you might want to consider: in the `test/suites/ACID.m` file, you can add a test case to [automate the testing](https://github.com/matlab2tikz/matlab2tikz/blob/master/test/README.md).
In the big cell array at the top of the file, add `@colorbars` at the end.
At the bottom of the file, you can add e.g. 
```matlab
function [stat] = colorbars()
  stat.description = 'Manual positioning of colorbars';
  stat.issues      = [933 937];

  shift = [0.3 0.3 0.7 0.7];
  axLoc = {'in','in','out','out'};

  for iAx = 1:4
    hAx(iAx) = subplot(2,2,iAx);
    axPos    = get(hAx(iAx), 'Position');
    cbPos    = [axPos(1)+shift(iAx)*axPos(3), axPos(2), 0.2 0.2]; 
    
    hCb(iAx) = colorbar('Position', cbPos, 'AxisLocation', axLoc{iAx});
    title(sprintf('AxisLocation = %s', axLoc{iAx}));
  end
end
``` 
which is equivalent to your first example.

Our graphical testing setup helps to create a PDF file with a comparison of MATLAB EPS files and the generated TikZ files.
Since you are really new to matlab2tikz, don't feel obliged to use the automated framework, as it can be quite daunting, but if you have any questions, I will try my best to guide you through it.

ulijh:
Hi, I've changed the positioning to be relative to the associated axes. It should now match exactly what it is in the matlab figure.
Setting the width in a vertical pgf cbar results in the following for me: 
bildschirmfoto vom 2016-07-25 10-30-09
Pgf sets it's own, reasonable width (if not manually specified), which seems ok to me.

I'll look into the automated testing tonight, I'm very interested in learing about it, but it's quite new to me.

ulijh:
sory, I din't mean to close... just comment

ulijh:
Hi,

Am 22.07.2016 um 17:44 schrieb Egon Geerardyn:
> Something else you might want to consider: in the |test/suites/ACID.m|
> file, you can add a test case to automate the testing
> <https://github.com/matlab2tikz/matlab2tikz/blob/master/test/README.md>.
> In the big cell array at the top of the file, add |@colorbars| at the end.
> At the bottom of the file, you can add e.g.
Test case |@colorbars| added to ACID.m. Do I need to manually ad a hash
to the md5 hash files?

> Our graphical testing setup helps to create a PDF file with a comparison
> of MATLAB EPS files and the generated TikZ files.
The PDF output of the new test looks fine. However I have problems
running some other tests, I guess its my setup with Matlab 9.0? I put an
output of the |runtests.sh| script into <http://pastebin.com/JsKHpNhx>.

Cheers, Uli

egeerardyn:
Don't worry about the failed tests. We normally only test on R2014a and R2014b (see e.g. the [jenkins results](http://pixiebox.synology.me:8090/job/matlab2tikz/528/) of your last commit).
In R2015a and newer, there are a few hashes that have changed, but we have not inspected in detail why those changes did occur (and what we can do about it, probably those are some small things in our code that can be improved).

Regarding the hashes: yes, you have to update them manually. The format is easy to edit by hand if needed, but in general, I recommend using the function we have in place for that.

```matlab
status = testHeadless(101); % 101 is your new test case
makeTravisReport(status);   % just to verify what has happened
saveHashTable(status, 'removedTests', 'keep');
% the "removedTests = keep" argument is to prevent the function from removing all other hashes
```

That will change the relevant hash file, which you can then commit to git. From the information on Jenkins, this should add the hash: `85c2b3530b67fa69b00adc2439bdf58f` (unless there are differences between your version and R2014b). In any case: graphical output is more important than those hashes (so don't fear: I think this PR is almost ready to be included).

src/matlab2tikz.m
```diff
@@ -4825,8 +4825,56 @@ function warnFacetedInterp(m2t)
...
+            end
+
+            % Using positions relative to associated axes
+            calcRelPos = @(pos1,pos2,ext2) (pos1-pos2)/ext2; 
+            cbarRelPosX = calcRelPos(cbarDim.left,cbarAxesDim.left,cbarAxesDim.width);
+            cbarRelPosY = calcRelPos(cbarDim.bottom,cbarAxesDim.bottom,cbarAxesDim.height);
+            cbarRelHeight = cbarDim.height/cbarAxesDim.height;
+            
+            cbarStyleOptions = opts_add(cbarStyleOptions, 'anchor',...
+                'south west');
```
miscco: Out of curriosity. Does this anchor fit every of the 4 available positions. I would guess we would have to adjust that too.

test/suites/ACID.m
```diff
@@ -2742,3 +2743,21 @@
   xlim([-1, 2]);
 end
 % =========================================================================
+function [stat] = colorbars()
```
miscco: I know it is a pita but given that we allow different positions of the color bar i think we should test for all 4 cases.

ulijh: Sure, can do that... But is it actually necessary?

miscco: The problem is that we might have subtle differences between the different cases. On the other hand it is just a pain. Maybe try once that the different hacks all work?

src/matlab2tikz.m
ulijh: As of my understanding, the matlab positions are always [south west X, south west Y, width, hight]. The positions (absolute and relative) in the tikz image and matlab plot are not exactly the same, I assume due to different margins an spaces for titles, labels etc. Thats why I calculate the positions relative to the parent axes.

When checking the plots, the locations seem the same in the m2t image and the matlab image.

ulijh:
Hi, looks like the tests pass now, the graphical looks ok to, so... I'm happy. ;-)
@egeerardyn Thanks for guiding me through all this, much appreciated.

src/matlab2tikz.m
miscco: Sweet, thanks for checking.

ulijh:
@miscco I've added some more realistic test cases. I used absolute positions here. Now we have all 4 position/location combinations with relative and absolute coordinates. The graphical tests look fine, cbars are where they are expected. Is this something like you were thinkin of?

test/suites/ACID.m
```diff
@@ -2742,3 +2747,109 @@
...
+
+  axLoc = 'out'; 
+
+  figure('Units','centimeters','Position',[1,1,11,10]);
+  axPos(1,:) = [1,1,8,3];
+  axPos(2,:) = [1,5,8,3];
+  cbPos      = [9.5,1,0.5,7]; 
+
+  hAx(1) = axes('Units','centimeters','Position',axPos(1,:));
+  imagesc([1,2,3],[4,5,6],rand(3),[0,1]);
```
egeerardyn: It is a bad practice to use `rand` inside testing (don't worry, I will fix it), since the output can change during each run. Instead, functions like `magic` can come in handy to have deterministic data.

test/suites/ACID.m
```diff
@@ -2742,3 +2747,109 @@
...
+
+  hAx(2) = axes('Units','centimeters','Position',axPos(2,:));
+  imagesc([1,2,3],[4,5,6],rand(3),[0,1]);
+
+  hCb = colorbar('Units','centimeters','Position', cbPos,...
+                 'AxisLocation', axLoc);
+  hCb.Label.String = (sprintf('AxisLocation = %s', axLoc));
```
egeerardyn: Please don't use the new-style `hCb.Label.String` syntax in the future, as that is not backwards compatible to MATLAB 2014a and Octave.
I will replace it with
```matlab
hLabel = get(hCb, 'Label');
set(hLabel,'String', ...)
```

test/suites/ACID.m
```diff
@@ -2742,3 +2747,109 @@
...
+end
+% =========================================================================
+function [stat] = colorbarManualLocationRightIn()
```
egeerardyn: Since most of your tests are just a variation on the same theme, that is really a hint that you should extract a helper function where the common functionality happens and you pass in the changing parts as parameters.

test/suites/ACID.m
```diff
@@ -2742,3 +2747,109 @@
...
+  end
+end
+% =========================================================================
+function [stat] = colorbarManualLocationRightOut()
+  stat.description = 'Manual positioning of colorbars';
+  stat.issues      = [933 937];
+
+  axLoc = 'out'; 
+
+  figure('Units','centimeters','Position',[1,1,11,10]);
```
egeerardyn: Please don't use `figure` in your tests: that is something that our test framework takes care of (as that allows us to properly close the figure after testing).

If you need to change the figure, you can always `set(gcf, ....)` to get the same effect.

egeerardyn:
@ulijh Thank you for your work on this. I have left some points of attention, but I will take care of them for you since this has been open for quit a while (and such that I can push this PR out with the imminent release).

test/suites/ACID.m
ulijh: Yes, I agree... , but didn't change it because it doesn't seem affect the hashes. I should have changed it anyways...

ulijh:
@egeerardyn Thanks for taking care of the remaining points and your help and comments. I'll remember them for next time.

PR #948, opened 2016-09-13
Use tex=color rather than font=color for tick labels

### LINKED ISSUE #947: font=\color{black} seems to have unwanted side effects
Hey there,

replacing `['{font=\color{',col,'}}']);` in line 1522 with `['{text={',col,'}}']);` or `['{text=',col,'}']);` will solve a problem that makes appended commands from the pgsplotsset section not to be appended to the axis environment (e.g. the fontsize will be ignored).

See here: http://tex.stackexchange.com/questions/329315/pgfplotsset-append-style-is-not-appended-to-tikzpicture-axis-environment?noredirect=1#comment806848_329315

miscco:
This should fix #947.


egeerardyn:
I will try to fix Jenkins tonight (we have moved over the last weekend, so my server has also been relocated physically with possible hiccups).

phwi:
Thank's. But it's not tex=color but text=color - but I've seen that this is only a mistake in the commit message but implemented correctly :-)

phwi:
Just wanted to link the original topic here, where can be seen that this issue affects more than the `color=text` fix that have not been addressed, yet:

https://github.com/matlab2tikz/matlab2tikz/issues/947

egeerardyn:
This patch is fine by me, if you can still update the hashes for the test suite. :+1: Sorry for the late reply.

okomarov:
This is not actual anymore since #965 takes care of axis font styling, inculding colour.

egeerardyn:
Let's close it then.

PR #956, opened 2016-11-22
Add authors

egeerardyn:
These were some commits that were supposed to go into the previous release, but somehow only ended up on a personal branch of mine.


egeerardyn:
Should be closed: against master instead of develop

PR #958, opened 2016-11-23
Add authors

egeerardyn:
This was los within my personal branches. Originally, those were intended for the previous release, but here they are now.

No code is affected, so CI does not need to be trusted.


egeerardyn:

PR #957, opened 2016-11-22
Github issue numbers -> URL in updater changelog

egeerardyn:
This only works in MATLAB, but brings the github issues
within reach of our users.

In the updater, it should add links to all GitHub issue pages.

Again, this was something I had lying in a personal branch for quite a while.

PR #959, opened 2016-11-23
Github issue numbers -> URL in updater changelog

egeerardyn:
This only works in MATLAB, but brings the github issues
within reach of our users.

Re-PR of #957, against develop this time. CI is worthless in this case, since the updater is not covered anyhow.


egeerardyn:

PR #962, opened 2016-12-13
Fix choice of shader for surface

okomarov:
Line style none gets ignored and `shader=interp faceted` is picked, resulting in **wrong image** and very **heavy .pdfs**.

capture

```
h = ribbon(randn(3,1));
set(h,'CData', get(h,'ZData'),'FaceColor','interp','LineStyle','none')
view(0,90)
```
For this example, the size of the pdf is negligible. On other figures, I get figures > 1MB while with the correct `shader=interp` they are reduced to about 100KB, so an **order of magnitude** better!


egeerardyn:
Looks great! :+1:

PR #964, opened 2016-12-18
Fix #963, i.e. datetime ticks in R2016b 

### LINKED ISSUE #963: datetime or duration (e.g. on x axis) results in error
Based on the [comment](https://github.com/matlab2tikz/matlab2tikz/issues/383#issuecomment-260650219) by [awesomeindustries](https://github.com/awesomeindustries) in [issue 383](https://github.com/matlab2tikz/matlab2tikz/issues/383), the following code results in an error in MALTAB R2016b.

```matlab
data = rand(10, 1);
plotdates = datetime : datetime+numel(data)-1;
figure
plot(plotdates, data)
matlab2tikz file
```

I, too, have this problem on MATLAB R2016b. I do not have R2016a or earlier installed any more, so I cannot test if this is new behaviour or that it has existed before. Issue 383 was closed as solved in November 2015, so this issue could be present in R2016a too, assuming the closer had no problems on R2015b.

The error text:

```matlab
Error using datetime/horzcat (line 1260)
All inputs must be datetimes or date/time character vectors or date/time strings.
Error in matlab2tikz>getXYZDataFromLine (line 1829)
        data = [xData(:), yData(:)];
Error in matlab2tikz>drawLine (line 1735)
    data       = getXYZDataFromLine(m2t, h);
Error in matlab2tikz>handleAllChildren (line 681)
                [m2t, str] = drawLine(m2t, child);
Error in matlab2tikz>drawAxes (line 880)
    [m2t, childrenEnvs] = handleAllChildren(m2t, handle);
Error in matlab2tikz>saveToFile (line 482)
        m2t = drawAxes(m2t, relevantAxesHandle);
Error in matlab2tikz (line 334)
    m2t = saveToFile(m2t, fid, fileWasOpen); 
```

It seems `matlab2tikz>getXYZDataFromLine (line 1829)
        data = [xData(:), yData(:)];` gets `datetime` data and non-`datetime` data and tries to concatenate the two, which isn't allowed for these types of data (`double` and `datetime`). EDIT: it appears to be allowed for `double` and `duration`, e.g. try `[1 seconds]`. However, the error if you issue `[1 datetime]` is the same as the one `matlab2tikz` throws:

```matlab
Error using datetime/horzcat (line 1260)
All inputs must be datetimes or date/time character vectors or date/time strings. 
```

This may be fixable by converting the `data` matrix implementation into a `cell` or `table`. With `table`, a lot more is possible too, so this may be preferable. However, it was introduced in R2013b, which is not too long ago. so a `cell` may be a better type to improve backwards compatibility. It seems not too difficult to adjust the other functions using `data` to work with a `table` or `cell`. However, I can't tell how to properly handle MATLAB's built-in types other than the numeric types, i.e. `datetime` and `duration`, further down the line.

Using a `duration` results in different errors. Example code:

```matlab
data = rand(10,1);
plotdates = seconds(1:10);
figure
plot(plotdates,data)
matlab2tikz file
```
Error output:
```matlab
Error using sprintf
Unable to convert 'duration' value to 'double'.
Error in matlab2tikz>makeTable (line 5391)
        table{iRow} = sprintf(FORMAT, thisData{:});
Error in matlab2tikz>plotLine2d (line 1866)
    [m2t, table, tableOptions] = makeTable(m2t, repmat({''}, size(data,2)), data);
Error in matlab2tikz>writePlotData (line 1806)
            [m2t, Part] = plotLine2d(m2t, opts, dataCell{k});
Error in matlab2tikz>drawLine (line 1744)
    [m2t, dataString]  = writePlotData(m2t, data, drawOptions);
Error in matlab2tikz>handleAllChildren (line 681)
                [m2t, str] = drawLine(m2t, child);
Error in matlab2tikz>drawAxes (line 880)
    [m2t, childrenEnvs] = handleAllChildren(m2t, handle);
Error in matlab2tikz>saveToFile (line 482)
        m2t = drawAxes(m2t, relevantAxesHandle);
Error in matlab2tikz (line 334)
    m2t = saveToFile(m2t, fid, fileWasOpen); 
```
This time the conversion of `duration` to `double` isn't allowed. A `duration` has the difficulty it is dependent on units of time. A `duration` in seconds is different from a `duration` in hours. `matlab2tikz` could just use any unit to determine the coordinates of the TikZ plot, but then the final tick labels should respect the MATLAB plot's tick label format (I assume `matlab2tikz` already does this) or take the `Format` property of the `datetime` or `duration` to create the appropriate labels.

okomarov:
In R2016b datetime ticks are used directly rather than in just rendering the ticks labels. Convert `datetime` ticks to `datenum`. 

The test for versions below R2016b has still to go through the undocumented property, hence the try catch remains. 

Fixes #963

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -1602,21 +1602,22 @@ end
 % ==============================================================================
 function isDatetimeTicks = isAxisTicksDateTime(handle, axis)
     % returns true when the axis has DateTime ticks
-    try
-        % If R2016b or above
-        if isa(get(handle, [upper(axis), 'Tick']),'datetime')
-            isDatetimeTicks = true;
-        else
+
+    % If R2016b or above
+    if isa(get(handle, [upper(axis), 'Tick']),'datetime')
+        isDatetimeTicks = true;
+    else
+        try
             % Get hidden properties of the datetime axes manager
             dtsManager = get(handle, 'DatetimeDurationPlotAxesListenersManager');
             oldState   = warning('off','MATLAB:structOnObject');
             dtsManager = struct(dtsManager);
             warning(oldState);
-            
+
             isDatetimeTicks = dtsManager.([upper(axis) 'DateTicks']) == 1;
+        catch
+            isDatetimeTicks = false;
         end
-    catch
-        isDatetimeTicks = false;
     end
 end
 % ==============================================================================
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -1844,6 +1844,12 @@ function [data] = getXYZDataFromLine(m2t, h)
         xData = get(h, 'X');
         yData = get(h, 'Y');
     end
+    if isa(xData,'datetime')
+        xData = double(xData);
+    end
+    if isa(yData,'datetime')
+        yData = double(yData);
+    end
     is3D  = m2t.axes{end}.is3D;
     if ~is3D
         data = [xData(:), yData(:)];
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -1604,7 +1604,7 @@ function isDatetimeTicks = isAxisTicksDateTime(handle, axis)
     % returns true when the axis has DateTime ticks
 
     % If R2016b or above
-    if isa(get(handle, [upper(axis), 'Tick']),'datetime')
+    if isa(get(handle, [upper(axis), 'Tick']), 'datetime')
         isDatetimeTicks = true;
     else
         try
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -1533,6 +1533,9 @@ function [options] = getAxisTicks(m2t, handle, axis, options)
 
     % hidden properties are not caught by hasProperties
     isDatetimeTicks = isAxisTicksDateTime(handle, axis);
+    if isDatetimeTicks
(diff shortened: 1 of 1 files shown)


src/matlab2tikz.m
```diff
@@ -1599,16 +1602,22 @@ case guitypes()
...
-        dtsManager = get(handle, 'DatetimeDurationPlotAxesListenersManager');
-        oldState   = warning('off','MATLAB:structOnObject');
-        dtsManager = struct(dtsManager);
-        warning(oldState);
 
-        isDatetimeTicks = dtsManager.([upper(axis) 'DateTicks']) == 1;
-    catch
-        isDatetimeTicks = false;
+    % If R2016b or above
+    if isa(get(handle, [upper(axis), 'Tick']),'datetime')
```
egeerardyn: Minor thing: I would add a space after each `,`.

egeerardyn:
Feel free to merge whenever you like :smile:

erikhuizinga:
Thanks for the fix. Will try later. Does this fix also support 3D plots with datetimes on the z axis? I ask, because I see only x and y data are converted to date numbers.

okomarov:
That's a nice pointer... I never considered a use case for datetimes/durations on the z-axis. Can you please raise a separate issue  for that, maybe with an example of use case.

PR #965, opened 2016-12-20
Font style for axis labels

okomarov:
Tick label font style is now rendered, especially the font size, if calling `matlab2tikz(...,'StrictFontSize',true)`.

There is overlap with `getFontStyle()` but it differs in the color treatment, e.g. `XColor` is what we need when dealing with axes. Do not have much time in thinking what's the best integration, but the code is clean and easy to optimize for a future review.


okomarov:
Jesus the nightmare with the previous version defaults...Now I recall that axes had black as default then 0.15, but it wasn't honored in one or two versions, and now the recent versions, since R2016a(?), do honor the 0.15 values...

okomarov:
This should also fix #947

egeerardyn:
@okomarov I agree with your observation that those defaults are a total PITA. I think I have raised that issue with TMW more than a year ago regarding their non-conformance to their own standards.

In any case, I think it is probably the most prudent to move those defaults to the latest versions (as you did) now, since HG1 will at one point or another will no longer be supported (by TMW and also by us, as a consequence).

In this case, we just need to update the hashes. If you are getting reproducible results for newer versions of MATLAB, I think I have a few installs of MATLAB on my server such that we can incorporate them into the CI next to R2014b (or instead of R2014b  eventually, as I will have to buy a license to keep the CI server running after next spring.

PR #971, opened 2016-12-23
Fix #966 and #970

### LINKED ISSUE #966: datetime on z-axis not supported
Although #964 fixes #963 partially, `datetime` data is not yet supported on the z axis in a 3D plot.

(using current develop branch)
```matlab
x = 1:10;
y = seconds(1:10);
z = datetime + seconds(1:10);
figure
plot3(x, y, z)
matlab2tikz xyz
```

This results in [m2t line 1870](https://github.com/matlab2tikz/matlab2tikz/blob/develop/src/matlab2tikz.m#L1870) throwing the same concatenation error, as it tries to concatenate `datetime` data with other non-`datetime` data. The same fix as used in #964 can be applied successfully here.

### LINKED ISSUE #970: pgfplots may have trouble with small datetime ranges on an axis
If the limits of an axis are small, then `pgfplots` may have trouble with the axis limits, adjusts them (bad, bad boy `pgfplots`!) and issues a warning:

> Package pgfplots Warning: Axis range for axis [x | y | z] is approximately empty; enlarging it (it is [[newMin]:[newMax]]) on input line [lineNumber].

For example, try the following and see the LaTeX output below.

```matlab
x = datetime(2017, 1, 1) + seconds(0:10);
y = 0:10;
figure
plot(x, y)
matlab2tikz m2t.tex
```
m2t-5

Note the x axis labels are incorrectly put on top of each other, because `pgfplots` has changes the x axis limits. This is caused by the conversion of `datetime` data to `double` through the `datenum()` function. This results in very large date numbers (more or less the number of days since the year 0), but the axis limits range is relatively small. Compare:
```matlab
data = datenum(datetime(2017, 1, 1) + seconds(0:1));
data(1), range(data)
```
If you run this code, you can see in the output that the difference in order of magnitude between the data and its range is about 10^10, which apparently is too much for `pgfplots` to handle.

This can be solved by subtracting the relatively large offset from the date numbers.

erikhuizinga:
I've fixed #966 and attempted to fix #970. However, I have made substantial changes in [4aba476](https://github.com/matlab2tikz/matlab2tikz/commit/4aba476d71b8892c56fdfb21d0ca19b7a2ef93d0) and I don't know whether or not that will break stuff elsewhere in m2t.

I have no experience with testing m2t. I've tried to run the tests, but both resulted in failure. I'd appreciate it if you thoroughly review my code and test it yourselves. Please elaborate on your methods to run the tests and how you come to the conclusion my PR is ok or not. Thanks!

The code from #970 is repeated here:
```matlab
x = datetime(2017, 1, 1) + seconds(0:10);
y = 0:10;
figure
plot(x, y)
matlab2tikz m2t.tex
```
The result with the changes in this PR:
m2t-6
As you can see, the horizontal axis shows the labels correctly positioned, exactly as in the MATLAB figure.

The actual date label is still missing, for which I'll open another issue.

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -876,11 +876,7 @@ function m2t = drawAxes(m2t, handle)
         m2t = m2t_addAxisOption(m2t, 'point meta max', sprintf(m2t.ff, clim(2)));
     end
     % - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
-    % Recurse into the children of this environment.
-    [m2t, childrenEnvs] = handleAllChildren(m2t, handle);
-    m2t.axes{end} = addChildren(m2t.axes{end}, childrenEnvs);
-    % - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
-    % The rest of this is handling axes options.
+    % Handle axes options.
     % - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
     % Get other axis options (ticks, axis color, label,...).
     % This is set here such that the axis orientation indicator in m2t is set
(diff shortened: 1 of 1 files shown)


egeerardyn:
@matlab2tikzBot: test this PR

okomarov:
To make it easier for us to review, we stick to the one PR per feature or bugfix. So, I kindly ask you to split your PR.

Also, as commented in #970, there is a simple workaround for that.

erikhuizinga:
How do I split a PR? Simply create two new one with the separate features/bugfixes? Or is there a better way?

okomarov:
I am not aware of any smart way to create separate PR. Just issue a new branch for each feature from develop, then cherry-pick the commits that you need.

erikhuizinga:
Just cleaning up my old open PRs.

PR #972, opened 2016-12-26
Fixes invalid syntax in R2016b

okomarov:
Fixes error from invalid legend syntax


okomarov:
Merging since only ACID is changed.

PR #978, opened 2017-02-25
drawArrow()

bene86:
I tried to implement the drawArrow()-function. For me it works as expected.


egeerardyn:
I think something went wrong: you seem to have duplicated the `matlab2tikz.m` file and from the diff of `src/matlab2tikz`, I have the impression that that file does not call either of your new functions anywhere.

The quality of the code itself is good, so certainly pull-able.
Only other small remark: please file a PR against the `develop` branch in the future.

bene86:
hmm, I somehow messed it up, sorry for the confusion. I tried to correct the commit. Is it fine now (apart from the wrong branch, sorry :/ )

egeerardyn:
Thank you for your work on this, don't worry about those small things :smile:

egeerardyn:
This has been merged by 5f77b3d into `develop`. Thank you for your contribution 👍, I will close this PR.

@matlab2tikz/developers I have overruled CI since it appears that some Octave-GraphicsMagick integration stuff is broken (I have noticed something similar on my own computer a while back). Since this covers a completely new case, CI doesn't cover it anyway.

bene86:
there is one thing, I just noticed: I used the stealth-arrowtip of tikz, since it looks the closest to the matlab arrowtip. If you want to use this arrowtip, you need to have load the appropriate library in latex: \usetikzlibrary{arrows.meta}
Is this something which should be addressed somehow?

nschloe:
@bene86 If that's something that's more simular to MATLAB's output, we could enforce it when using `strict=True`.

egeerardyn:
@bene86: I doublechecked it. The `arrows.meta` library is already a requirement for some other parts. I will make sure that in your latest patch, this dependency is also communicated to the end user at runtime.

PR #988, opened 2017-04-08
extended drawArrow() for DoubleEndArrow

bene86:
I extended the drawArrow()-function by the DoubleEndArrows

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -3128,6 +3128,13 @@ function [m2t, str] = drawArrow(m2t, handle)
         return; % there is nothing to plot
     end
     
+    % Arrow Style
+    if isa(handle,'matlab.graphics.shape.DoubleEndArrow')
+        arrowSpec = '{Stealth}-{Stealth}';
+    else
+        arrowSpec = '-{Stealth}';
+    end
+    
     % Color
     color         = get(handle, 'Color');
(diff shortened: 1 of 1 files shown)


src/matlab2tikz.m
```diff
@@ -3142,8 +3142,13 @@ function assertRegularAxes(handle)
     [m2t, pos1, pos2] = getPositionOfArrow(m2t, handle);
 
     styleOpts = opts_print(drawOptions);
-    str       = sprintf('\\draw[-{Stealth}, %s] %s -- %s;\n', ...
-                        styleOpts, pos1, pos2);
+    if isa(handle,'matlab.graphics.shape.DoubleEndArrow')
```
egeerardyn: While correct, I think it would be better to do something like:
```matlab
if isa(handle, 'matlab.graphics.shape.DoubleEndArrow')
    arrowSpec = '{Stealth}-{Stealth}';
else
    arrowSpec = '-{Stealth}';
end
```
and then either insert that `arrowSpec` into `drawOptions` (as the very first thing after `opts_new`), or add it via a `'%s'`  option into the output string.

That would reduce a bit of duplicate code and possibilities for formatting problems in the future.

egeerardyn:
Thank you for your contribution. 👍 
I will look for the dependency you brought up a while back, and ensure it is addressed properly.

PR #990, opened 2017-04-20
Add forget plot to Scatter plot

Job89:
Add forget plot in case not all Scatters are in the legend

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -3930,7 +3930,11 @@ function [m2t, str] = drawScatterPlot(m2t, h, custom)
             'scatter/@pre marker code/.append style', ...
             '{/tikz/mark size=\perpointmarksize}');
     end
-
+    % Add forget plot
+    if ~m2t.currentHandleHasLegend 
+        hiddenDrawOptions = maybeShowInLegend(false, drawOptions);
+        drawOpts = opts_print(hiddenDrawOptions);
+    end
     % The actual printing.
     [m2t, table, tableOptions] = makeTable(m2t, columns, data);
     tableOptions = opts_merge(tableOptions, metaPart);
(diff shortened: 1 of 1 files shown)


src/matlab2tikz.m
```diff
@@ -3930,7 +3930,11 @@ function warnFacetedInterp(m2t)
             'scatter/@pre marker code/.append style', ...
             '{/tikz/mark size=\perpointmarksize}');
     end
-
+    % Add forget plot
+    if ~m2t.currentHandleHasLegend 
+        hiddenDrawOptions = maybeShowInLegend(false, drawOptions);
+        drawOpts = opts_print(hiddenDrawOptions);
```
egeerardyn: I don't think this can work: `drawOpts` is overwritten later on (line 3940/3944), so I am pretty sure your change does not affect the produced output.

What, I think, could work instead of your change:
```matlab
drawOptions = maybeShowInLegend(m2t.currentHandleHasLegend, drawOptions);
```
at the exact location where you put the code.
There is no need for `opts_print` (that is meant for outputting as string), and no need for an `if`, since the function you are calling is:
```matlab
function [drawOptions] = maybeShowInLegend(showInLegend, drawOptions)
    % sets the appropriate options to show/hide the plot in the legend
    if ~showInLegend
        % No legend entry found. Don't include plot in legend.
        drawOptions = opts_add(drawOptions, 'forget plot');
    end
end
```

Thank you for addressing this oversight! 👍  Could you try out with the suggested fix, to see whether that solves the problem you were facing?

Job89:
Your suggestion with drawOptions = maybeShowInLegend(m2t.currentHandleHasLegend, drawOptions); works for me.

egeerardyn:
Thank you 👍

PR #995, opened 2017-06-22
Change the test hashes to expected results from using octave >= 4.2.1

sil2100:
Currently, on Ubuntu, matlab2tikz 'reliable' unit tests are failing when ran against liboctave4 (so the octave library coming from version 4.2). By examining the failures it can be seen that changes in octave caused additional style formatting appears in the axes definitions. The resulting tex files seem to be sane otherwise.

More context and links to test failures can be found in our Ubuntu Launchpad package bug:
https://bugs.launchpad.net/ubuntu/+source/matlab2tikz/+bug/1699176

Those failures are causing delays in the ongoing liboctave3v5 -> liboctave4 transition.


egeerardyn:
Fine by me to merge, but please issue a PR on the `develop` branch instead. I will do the necessary to update master as well afterwards.

It seems that there are some textual differences between Octave 4.0 (Travis) and Octave 4.2. The most likely cause is that our tests are way too strict to be reliable over all these versions. However, the current status of all our test hashes is quite unreliable (limited access to MATLAB, and limited Octave experience from the core devs).

sil2100:
Ok, will do so, thanks! Since I see there are some changes in develop in regards to master, I would need to re-run the tests and re-generate the hashes against the new liboctave4. This might take some moments before I can do that.

Thank you for taking a look at this though! I appreciate upstream approval since I would like to do this hash-change as a cherry pick to the current Ubuntu version to unblock things.

PR #1024, opened 2018-02-12
Add support for categorical variables in bar plots

m-pilia:
Categorical values can be plotted in tikz by defining the values as
symbolic, and passing the data as a set of coordinates. This commit
implements a function to format data as coordinates, and it adds some
checks to handle categorical values in bar plots (previously, trying to
plot a bar plot with categorical values generated an error).

I am not very expert with pgfplots, so I tried to adhere as much as I could to the
style of the pre-existing code. If there is a smarter way to do this, please let me
know!

The test suite generates the following report (with the same fails as in the commit preceeding my PR).

## Test summary
Test results for m2t commit 9330627083 running with MATLAB 9.3 on Linux 4.15.2-2-ARCH.

```matlab
suite = @ACID;
alltests = [1:105];
reliable = [2:5 7:9 11:17 19:23 25:33 35:37 39 41 44:47 50:57 59:93 95:96 98:105];
unreliable = [1 6 10 18 24 34 38 40 42:43 48:49 58 94 97];
failReliable = [2:5 7:9 11:16 20:23 25:27 29:31 36:37 39 44:47 50:57 59:73 75:78 80:93 95:96 98:105];
passUnreliable = [];
skipped = [17];
```

| Unreliable |    0 |   15 |    0 |    15 |
|   Reliable |    8 |   81 |    1 |    90 |
|      Total |    8 |   96 |    1 |   105 |

Build fails with 81 errors. :heavy_exclamation_mark:

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -1519,7 +1519,7 @@ function [m2t, options] = getAxisOptions(m2t, handle, axis)
     end
     % - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
     % get axis limits
-    options = setAxisLimits(m2t, handle, axis, options);
+    [m2t, options] = setAxisLimits(m2t, handle, axis, options);
     % - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
     % get ticks along with the labels
     [options] = getAxisTicks(m2t, handle, axis, options);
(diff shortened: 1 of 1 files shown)


src/matlab2tikz.m
```diff
@@ -5560,6 +5590,66 @@ function warnFacetedInterp(m2t)
...
+    FORMAT = repmat({m2t.ff}, 1, nColumns);
+    FORMAT(cellfun(@isCellOrChar, data)) = {'%s'};
+    FORMAT(cellfun(@iscategorical, data)) = {'%s'};
+    FORMAT = ['(', join(m2t, FORMAT, COLSEP), ')'];
+
+    for iRow = 1:nRows
+        thisData = cell(1,nColumns);
+        for jCol = 1:nColumns
+            if iscategorical(data{jCol}(iRow))
+                % do not make categorical values lowercase
```
egeerardyn: This seems to have a lot of duplication with `makeTable`; while both can be merged into a single function, I guess (if not, we should document why). 

I would guess that this part is one of the few places with differences (and picking the `FORMAT` above), so please just adapt `makeTable` to reflect the changes.

I think we can also reduce the complexity of this function quite a bit. E.g. what I notice here is that we can reduce the number of calls to `iscategorical`, but I guess we could even factor out most calls to `lower`, such that we don't even need to check whether the data is categorical in the loop (and for strings, I think we don't need the `lower` per se). But that last one is a note for myself, it is not something you should address.

src/matlab2tikz.m
```diff
@@ -5693,7 +5622,9 @@ function warnFacetedInterp(m2t)
         end
         table{iRow} = sprintf(FORMAT, thisData{:});
     end
-    table = lower(table); % convert NaN and Inf to lower case for TikZ
+    % convert NaN and Inf to lower case for TikZ
+    table = regexprep(table, '\<NaN\>', 'nan');
+    table = regexprep(table, '\<Inf\>', 'inf');
```
egeerardyn: Nice! I didn't even know about `\<` and `\>` in regular expressions. Thanks! I ❤️  the simplicity of this solution.

egeerardyn:
I will wait for Travis to do basic CI; if it runs without crashing, I will merge this PR. Thank you for tackling this!

egeerardyn:
Okay, silly me; it ran without crashing already in Travis 👍 Thanks again!

m-pilia:
My pleasure!
(Sorry, I was writing an accompanying message to the new commits, but you preceded me!)

PR #1029, opened 2018-03-28
Add test suite ACID.Octave.4.2.2.md5 for Octave 4.2.2

svillemot:
Compared to Octave 4.2.0, the output of several tests changed, hence the need
to update the hashes.

Patch applied to the Debian package.


egeerardyn:
Thank you. At the moment, the condition of our test suite is not really reliable, but if this makes you feel more comfortable working on the code, I'm all in favor 👍

PR #1030, opened 2018-04-05
Fix a bug in isAxisMain()

m-pilia:
Hello!

In MATLAB R2016b the LegendPeerHandle appdata field was renamed to
LayoutPeers (I think it is undocumented, the info comes from the docs
of this third party toolbox: ttp://rosa.unipr.it/FSDA/release_notes.html).

I added a version check to ensure that the correct field name is used on
recent MATLAB versions.


egeerardyn:
Thank you!

m-pilia:
My pleasure!

PR #1046, opened 2019-01-16
Fix #1006: create legend introduced in R20177b+

### LINKED ISSUE #1006: Legend bug - turned on by code - Matlab
Thanks for the awesome code!

**Bug Summary:**
There is no legend in a simple 2D line series plot with 4 lines. The M2T code turns it on, and so it appears in the TeX code.
Using Matlab 2017b (Prerelease) and M2T v1.1.0 on Windows.

**Cause:**
Wen querying axes object for the legend, Line 1064 of M2T has the code:
`legendhandle = legend(axisHandle);`
This actually turns the legend on with default strings ('data1', data2', etc), and so for the remaining M2T code, this default legend appears, along with 
`\addlegendentry{data1}`
in the TeX output.
This seems to be intentional design according to Matlab's documentation on Legend command.

**Suggested solution:**
Replace:
`legendhandle = legend(axisHandle);`
with
`legendhandle =findobj(axisHandle,'tag','legend');`
which does not turn on legend when querying its existence. Tested under my environment.

egeerardyn:
This should fix #1006

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -1061,7 +1061,14 @@ function legendhandle = getAssociatedLegend(m2t, axisHandle)
                 end
             end
         case 'MATLAB'
-            legendhandle = legend(axisHandle);
+	    if isprop(axisHandle, 'Legend') % R2017b behavior
+                legendhandle = axisHandle.Legend;
+	    else
+		legendhandle = legend(axisHandle);
+		% this is the proper way for old versions of MATLAB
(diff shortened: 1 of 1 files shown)


egeerardyn:
@jbjorsell Could you please check if this fixes it for you?

jbjorsell:
In my limited example it works well!

egeerardyn:
Thank you for checking :+1: Will integrate it into develop then :-)

PR #1048, opened 2019-02-11
Fix for GNU/Octave not including legends

corrmaan:
Using matlab2tikz in GNU/Octave 4.4.1 would not produce legends in the tikz code.

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -1048,17 +1048,25 @@ function legendhandle = getAssociatedLegend(m2t, axisHandle)
     % Get legend handle associated with current axis
 
     legendhandle = [];
-    env = getEnvironment();
+    [env, envVersion] = getEnvironment();
     switch env
         case 'Octave'
             % Make sure that m2t.legendHandles is a row vector.
             for lhandle = m2t.legendHandles(:)'
-                ud = get(lhandle, '__appdata__');
+                if isVersionBelow(envVersion, [4,2,2]) % 5865d2fef424
+                  lhandleProp{1}='UserData';
+                  lhandleProp{2}='handle';
+                else
+                  lhandleProp{1}='__appdata__';
+                  lhandleProp{2}='__axes_handle__';
+                end
+                ud = get(lhandle, lhandleProp{1});
                 % Empty if no legend and multiple handles if plotyy
-                if ~isempty(ud) && any(axisHandle == ud.__axes_handle__)
+                if ~isempty(ud) && any(axisHandle == ud.(lhandleProp{2}))
                     legendhandle = lhandle;
                     break
                 end
+                clear lhandleProp
             end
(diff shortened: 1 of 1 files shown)


egeerardyn:
I think this fix may break compatibility for older versions of Octave, while it shouldn't. Preferably, both locations are checked such that it works in different versions of Octave.

corrmaan:
You're right, sorry about that. Looks like this change was part of release 4.2.2 (http://hg.savannah.gnu.org/hgweb/octave/rev/94f8e2b5a88b):
https://hg.savannah.gnu.org/hgweb/octave/rev/5865d2fef424
Another commit that I noticed might be of interest for potential problems with colorbars:
https://hg.savannah.gnu.org/hgweb/octave/rev/5865d2fef424

I'll edit the PR to ignore this change for Octave versions < 4.2.2.

src/matlab2tikz.m
```diff
@@ -1048,17 +1048,25 @@ case guitypes()
...
 
     legendhandle = [];
-    env = getEnvironment();
+    [env, envVersion] = getEnvironment();
     switch env
         case 'Octave'
             % Make sure that m2t.legendHandles is a row vector.
             for lhandle = m2t.legendHandles(:)'
-                ud = get(lhandle, 'UserData');
+                if isVersionBelow(envVersion, [4,2,2]) % 5865d2fef424
```
egeerardyn:
```suggestion
                if isVersionBelow(envVersion, [4,2,2]) % Octave commit 5865d2fef424
```

src/matlab2tikz.m
```diff
@@ -1048,17 +1048,25 @@ case guitypes()
...
+                  lhandleProp{2}='__axes_handle__';
+                end
+                ud = get(lhandle, lhandleProp{1});
                 % Empty if no legend and multiple handles if plotyy
-                if ~isempty(ud) && any(axisHandle == ud.handle)
+                if ~isempty(ud) && any(axisHandle == ud.(lhandleProp{2}))
                     legendhandle = lhandle;
                     break
                 end
+                clear lhandleProp
```
egeerardyn:
```suggestion
```

Clear is not necessary. Yes, we have more variables in the local workspace, but I have also noticed that it can slow down a function considerably. So I would just remove it, unless it clearly interferes with different code (and in that case, it would be better to pick a different name for that variable.

corrmaan:
Looks good to me, thanks!

src/matlab2tikz.m
corrmaan: Looks good to me, thanks!

egeerardyn:
Thank you 👍 

(And just as I press the merge button, of course I notice it's on top of master instead of develop :-( silly me, ah well).

corrmaan:
My bad! I didn't realize there was a develop branch.

PR #1049, opened 2019-02-13
Fix for GNU/Octave not including legends

corrmaan:
Same as https://github.com/matlab2tikz/matlab2tikz/pull/1048 but for the develop branch.


egeerardyn:
Not needed, over the next few days, I will try to prepare a release, do a little sync-up between master and develop to ensure your patches end up in both master and develop :-)

Issue #819, opened 2015-11-23
Unit conversions


egeerardyn:
After the following post on FEX:
> 23 Nov 2015
> Tina Fuhrmann	
> Thanks for the great file! 
> I get the following error 
> `Warning: Can not convert unit 'normalized'. Using conversion factor 1.`
> when resizing a figure: 
> ```matlab
> set(figureHandle,'Units','normalized'); 
> set(figureHandle,'Position',[0 0 0.6 0.75]);
> ```
> Is there something I can do?

it got me thinking whether we shouldn't simplify the relevant part of the code. In principle, MATLAB can do most/all of those conversions for us. 

Otherwise, I also think we can implement the `normalized` case by also passing the handle in and querying its parent as well.

This is by no means high-priority, but it may be a nice part to reduce our codebase a bit.

okomarov:
That's what I suggested some time long ago

egeerardyn:
I'm sorry for not following you advice back then , :-1: for past me.

TimonSchulz:
Hello Everybody,

has this ever been put into action?

I would like to use matlab2tikz v1.1.0. on a figure that I fit to my screen first. However I get the exact same warning as mentioned above:

```
Warning: Can not convert unit 'normalized'. Using conversion factor 1. 
```

This is my code:
```
gosave = false
gomatlab2tikz = true

figure('units','normalized','outerposition',[0 0 1 1])
plot(Zeit1,Temperatur1,'LineWidth',2)
grid on
title('Temperatur-Messprogramm fuer Messreihe 1')
xlabel('Zeit')
ylabel('Temperatur')
xlim([0 150])
ylim([0 400])
xticks(linspace(0,150,16))
yticks(linspace(0,400,21))
set(gca,'FontSize',30)
if gosave == true
    exportgraphics(gcf,'Temperatur_Messprogramm_Messreihe_1.jpg','Resolution',150)
exportgraphics(gcf,'Temperatur_Messprogramm_Messreihe_1.pdf','Resolution',150,'ContentType','vector')
end
if gomatlab2tikz == true
    matlab2tikz('Temperatur_Messprogramm_Messreihe_1.tikz')
end
```

If not has anybody a suggestion of how to change the size of the matlab figure to a requested size, which is compatible with matlab2tikz?

Issue #340, opened 2014-05-22
Figure size conversion error


LucaAmerio:
If the matlab figure hasn't the units set as "Pixels", than matlab2tikz will convert that number to a wrong value of width and height causing occasionally Latex to give the error "Dimension too large".

To reproduce this error:

```
%open a fig file
matlab2tikz('pixels.tex')
set(gcf,'Units','normalized')
matlab2tikz('normalized.tex')
set(gcf,'Units','centimeters')
matlab2tikz('centimeters.tex')
```

The figure isn't changed between the passages, but the three latex files are different.

PR #1062, opened 2019-10-20
add support for normalized units in Matlab figures

jcmonteiro:
This PR should solve #819 and #340.

I ran the tests using runMatlab2TikzTests and got the same output as the one in the develop branch:

```
| Unreliable |    2 |   13 |    0 |    15 |
|   Reliable |   45 |   44 |    1 |    90 |
|      Total |   47 |   57 |    1 |   105 |
```

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -6006,6 +6006,12 @@ function dstValue = convertUnits(srcValue, srcUnit, dstUnit)
         return % conversion to the same unit => factor = 1
     end
 
+    if strcmp(srcUnit, 'normalized')
+        srcUnit = 'px';
+        monitorSize = get(0, 'MonitorPositions');
+        srcValue = srcValue .* monitorSize(1, 3:4);
+    end
+
     units  = {srcUnit, dstUnit};
     factor = ones(1,2);
     for ii = 1:numel(factor) % Same code for srcUnit and dstUnit


egeerardyn:
Thank you for your pul request. 

I don't think this does everything for normalized units, because (if I remember correctly), those units are normalized with respect to the container of the object.
I.e. an axis with normalized width = 0.5 spans half the figure, a figure with normalized width = 0.5, spans half the screen (or even half the desktop possibly on a multi-screen setup, I reckon). 

While this is already an improvement (it handles those situations in some cases, but not everywhere), I think it would be better to either implement the full thing (which means passing in the handle of the object into that function, writing a function to find the parent, doing the conversion) or at least keeping some kind of warning that the conversion might be incorrect.

Because I don't think either issue can be closed as-is.

jcmonteiro:
Oh, I didn't know that the normalized units option worked like this. In the way I implemented it, I guess the conversion is only going to work when converting a figure.

Do you think passing the handle, using this piece of code if it points to a figure, and otherwise issuing a warning is enough to get this PR merged?

egeerardyn:
As far as I'm concerned, that's fine indeed. But you might need to change quite a bit of code to pass in those handles. In any case; as long as it keeps warning the users that the conversion is not perfect, it's fine by me :-)

jcmonteiro:
Closing since it's been years since I looked into this one

PR #1075, opened 2020-05-21
Explain a bit better the options' combinations

doronbehar:
Hi,

Thanks for creating this tool / script. I felt like this small documentation was missing as I've managed to only guess this is how this will work.

Regards.

Also, another question:

Why are you using `MATLAB2TIKZ()` (uppercase)? The function must be called in lower case for Octave users for as far as I can see...

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -1,7 +1,10 @@
 function matlab2tikz(varargin)
     %MATLAB2TIKZ    Save figure in native LaTeX (TikZ/Pgfplots).
     %   MATLAB2TIKZ() saves the current figure as LaTeX file.
-    %   MATLAB2TIKZ comes with several options that can be combined at will.
+    %   MATLAB2TIKZ comes with several options that can be combined at will. For
+    %   example:
+    %
+    %       MATLAB2TIKZ('showInfo', false, 'strict', true, "figure.tex")
     %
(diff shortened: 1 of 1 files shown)


egeerardyn:
The reason for the uppercase is because in MATLAB (at least at some point), help files were (are?) supposed to uppercase their own function name (even though MATLAB itself was increasingly pedantic about the case). 

Thank you for the suggestion, it can indeed help people understand it better. 
However, I think there are two problems with the example you give:
- typo `shofInfo` instead of `showInfo`
- if I remember correctly, the file name must come first, not last.

doronbehar:
> typo shofInfo instead of showInfo

Correct :+1:, forced pushed a fix.

> if I remember correctly, the file name must come first, not last.

Seems to work for me when I use:

```m
matlab2tikz('showInfo', false, 'width', '0.8\textwidth', "plot.tikz")
```

As for the CI error https://travis-ci.org/github/matlab2tikz/matlab2tikz/builds/689631120#L724 :

> The command "./runtests.sh /usr/bin/octave" exited with 1.

This has nothing to do with this PR right?

egeerardyn:
> Seems to work for me when I use: matlab2tikz('showInfo', false, 'width', '0.8\textwidth', "plot.tikz")

Okay, that surprises me and means that out input parsing does something I didn't anticipate.

Travis failing has nothing to do with this PR indeed. In fact, those tests are broken for way too long already.

PR #1084, opened 2020-08-17
Enable to configure mark size explicitly

doronbehar:
This enables me to make these plots:

Screenshot from 2020-08-17 20-50-02

Look like this:

Screenshot from 2020-08-17 20-51-37

Without doing this hacky thing after I run `matlab2tikz`:

```matlab
	fid = fopen(plot_fname,'r');
	f=fread(fid,'*char')';
	fclose(fid);
	fid = fopen(plot_fname,'w');
	fprintf(fid,'%s',regexprep(f,'mark options={solid, ','mark options={solid,  mark size=0.4pt, '));
	fclose(fid);

```

-----

My dearest thank you, for creating matlab2tikz, it's impressively built and it has made me proud of my lab reports.

Regards.

### DIFF
--- a/src/matlab2tikz.m
+++ b/src/matlab2tikz.m
@@ -76,6 +76,9 @@ function matlab2tikz(varargin)
     %   MATLAB2TIKZ('extraTikzpictureOptions',CHAR or CELLCHAR,...)
     %   explicitly adds extra options to the tikzpicture environment. (default: [])
     %
+    %   MATLAB2TIKZ('markerSize', CHAR, ...)
+    %   explicitly specify a speicific 'mark size' to 'mark options={}'. (default: '')
+    %
     %   MATLAB2TIKZ('encoding',CHAR,...) sets the encoding of the output file.
     %
     %   MATLAB2TIKZ('floatFormat',CHAR,...) sets the format used for float values.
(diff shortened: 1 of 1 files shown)


egeerardyn:
I would argue that both are a bit hacky.

You should be able to get the same kind of output, if you adapt the marker size directly on your MATLAB plot objects.
I don't know by heart whether that `0.4pt` is the default or not, but if it is, you can pass the `strict` option to force that it passes to the output.

Another option is to look into `[m2tcustom](https://github.com/matlab2tikz/matlab2tikz/blob/master/src/m2tcustom.m)`, I think you can get the same result as you have by attaching that to the right line object, e.g.

```matlab
h = plot(1:10,'b-o')
m2tcustom(h, 'extraOptions', 'mark options=0.4pt')
matlab2tikz(...)
```
I've seen that it was not yet available on the master branch, but now it should be.

Another function that might help you get hold of the right handles in already existing plots is `findobj`.

doronbehar:
Indeed your method works, only it should be `mark size=0.4pt` - (**size** and not **options**). I wonder though if I'd have ever found out about this.. I added a new section to the FAQ WiKi at https://github.com/matlab2tikz/matlab2tikz/wiki/FAQ#my-plots-markers-dots-size-is-too-big . Feel free to close this and thanks for responding.

PR #1089, opened 2020-10-18
Update cleanfigure.m

saumyadipsss:


egeerardyn:
This is not doing something we want. If you want to clear your variables and clear your terminal, please wrap cleanfigure into your own script instead of rolling it into cleanfigure. Its purpose is to clean a figure, not your variables and terminal.

Issue #1063, opened 2019-10-29
Warnings with MATLAB 2019b


leptoceratops:
I have installed the latest release of MATLAB, i.e. **R2019b** and when I run **matlab2tikz** before hovering over the figure with the mouse pointer everything works fine without any warnings. If I then hover over the figure with the mouse (and the latest axes toolbar appears) then if I run **matlab2tikz** I get the following warnings that were not present in the previous releases.

> Warning: Don't know annotation 'matlab.graphics.shape.internal.Button'. 
> In matlab2tikz>userWarning (line 5934)
  In matlab2tikz>drawAnnotationsHelper (line 3058)
  In matlab2tikz>drawAnnotations (line 3003)
  In matlab2tikz>saveToFile (line 491)
  In matlab2tikz (line 334) 
Warning: Don't know annotation 'matlab.graphics.shape.internal.Button'. 
> In matlab2tikz>userWarning (line 5934)
  In matlab2tikz>drawAnnotationsHelper (line 3058)
  In matlab2tikz>drawAnnotations (line 3003)
  In matlab2tikz>saveToFile (line 491)
  In matlab2tikz (line 334) 
Warning: Don't know annotation 'matlab.graphics.shape.internal.Button'. 
> In matlab2tikz>userWarning (line 5934)
  In matlab2tikz>drawAnnotationsHelper (line 3058)
  In matlab2tikz>drawAnnotations (line 3003)
  In matlab2tikz>saveToFile (line 491)
  In matlab2tikz (line 334) 
Warning: Don't know annotation 'matlab.graphics.shape.internal.Button'. 
> In matlab2tikz>userWarning (line 5934)
  In matlab2tikz>drawAnnotationsHelper (line 3058)
  In matlab2tikz>drawAnnotations (line 3003)
  In matlab2tikz>saveToFile (line 491)
  In matlab2tikz (line 334) 
Warning: Don't know annotation 'matlab.graphics.shape.internal.Button'. 
> In matlab2tikz>userWarning (line 5934)
  In matlab2tikz>drawAnnotationsHelper (line 3058)
  In matlab2tikz>drawAnnotations (line 3003)
  In matlab2tikz>saveToFile (line 491)
  In matlab2tikz (line 334) 
Warning: Don't know annotation 'matlab.graphics.shape.internal.Button'. 
> In matlab2tikz>userWarning (line 5934)
  In matlab2tikz>drawAnnotationsHelper (line 3058)
[...]

PR #1113, opened 2022-11-14
Handle new mouseover GUI in current matlab releases (#1063)

ploebb:
Simple fix for #1063

This toolbar leads to a lot of warnings about unknown annotations:
grafik


egeerardyn:
Thank you!

PR #1114, opened 2022-12-03
Fix arrow and doublearrow annotation not being drawn in final output

federunco:
Added support for "arrowshape" and "doubleendarrowshape" type objects in handleAllChildren fcn


egeerardyn:
Thank you for adding this. :+1:

PR #1121, opened 2023-02-03
Fix setting alpha data in PNGs

raimund-schluessler:
An images `AlphaData` channel might be of type `logical`, but `imwrite` expects alpha to be of type `double, uint8, uint16`. This converts the alpha channel data to double.


egeerardyn:
Thank you

PR #1155, opened 2026-03-04
Add new error message to m2tInputParser.m

hms0411rt:
I was confused why matlab2tikz("out.tex") told me my input was invalid. I'd appreciate it if a message could give a type hint, like an error message telling user to use char instead of string if a string is detected.


hms0411rt:
Closed the PR accidentally while I was trying to change the branch from master to develop

